PR #5, opened 2018-09-27
update Connect mutation doc to show variable passing

blbradley:
Otherwise, had to read library code to find variables argument to mutation function.

Closes aws-amplify/amplify-js#1762.


blbradley:
I used the GitHub editor, not sure why the last line changed.

PR #13, opened 2018-10-09
Codgen docs

yuth:


cli/codegen.md
```diff
@@ -0,0 +1,46 @@
+<a href="https://aws-amplify.github.io/" target="_blank">
+    <img src="https://s3.amazonaws.com/aws-mobile-hub-images/aws-amplify-logo.png" alt="AWS Amplify" width="550" >
+</a>
+
+# AWS Amplify CLI codegen configuration & Workflow
+
+Codegen helps you generate native code for iOS and android, generation of types for Flow and Typescript. It can also generates GraphQL statements(queries, mutations and subscriptions) so that you doesn't have to hand code them.
```
undefobj: Android (cap "A") and TypeScript (cap "S") should be used everywhere.

cli/codegen.md
```diff
@@ -0,0 +1,46 @@
...
+
+Codegen `add` workflow triggers automatically when an AppSync API is pushed to the cloud. You will be prompted if you want to configure codegen when an AppSync API is created and if you opt-in for codegen, subsequent pushes prompts you if they want to update the generated code after changes get pushed to cloud.
+
+When an project is configured to generate code with codegen, it stores all the configuration `.graphqlconfig.yml` file in the root folder of your project. Codegen behind the scene use [`amplify-graphql-docs-generator`](https://www.npmjs.com/package/amplify-graphql-docs-generator) to generate the GraphQL statements and [`aws-appsync-codegen`](https://www.npmjs.com/package/aws-appsync-codegen) to generate the types.
```
undefobj: @yuth can you chat with @mlabieniec on the aws-appsync-codegen reference here? I get that it works like this now but really we want to pull this package in (or port over the code) for maintainability so maybe not point to this otherwise customers will look to that code for issues?

cli/codegen.md
```diff
@@ -0,0 +1,46 @@
...
+
+When generating types, codegen uses GraphQL statements as input. It will generates only the types that are being used in the GraphQL statements.
+
+
+## amplify add codegen
+```
+$ amplify add codgen [--apiId <api-id>]
```
undefobj: ...note that if you use the `--apiId` flag to add an externally created AppSync API, such as one created in the AWS console, you will not be able to manage this API from the Amplify CLI with commands such as `amplify api update` when performing schema updates.

cli/codegen.md
```diff
@@ -0,0 +1,46 @@
...
+```
+$ amplify configure codgen
+```
+The `amplify configure codegen` command allows you to update the codgen configuration after it is added to your project. 
+
+## amplify codegen statements
+The `amplify codegen statements [--nodownload]` command  generates GraphQL statements(queries, mutation and subscription) based on your GraphQL schema. This command downloads introspection schema everytime it is run but it can be forced to use previously downloaded introspection schema by passing `--nodownload` flag
+
+
+## amplify codegen types
```
undefobj: Can we add instructions to how people can put this in their build systems for Xcode, Android Studio, and Webpack (package.json build step)?

cli/codegen.md
undefobj: The thing is, we've already started with a banner there. Let's talk timeline with Mike on the other stuff.

PR #49, opened 2018-10-24
Corrected path/syntax for plugin/modules for use in Vue.use

codylandry:
I had issues using the code from this example and found this to work after logging the values of `AmplifyVue` and `AmplifyModules`.


mlabieniec:
The doc has been updated outside this PR, please let us know if you still feel like it needs to be updated based on this commit @haverchuck FYI

PR #60, opened 2018-10-28
update(GettingStarted-iOS): Proposed flow and ported manual SDK confi…

undefobj:
This has new proposed flow for iOS getting started as well as manually configuring SDK. I also put a navigation on the bottom left for the manual configuration. This is where we'll also give customers some instructions for using the autogenerated SDK classes for other services.


undefobj:
Need to update .spelling file to account for v2.0

PR #76, opened 2018-11-01
Nothing happen after react-native link

TronNatthakorn:
Nothing link in android/settings.gradle. Nothing happen.

Perhaps it should be `react-native link amazon-cognito-identity-js`

### DIFF
--- a/js/start.md
+++ b/js/start.md
@@ -277,7 +277,6 @@ $ react-native init myReactNativeApp
 $ cd myReactNativeApp
 $ npm install --save aws-amplify
 $ npm install --save aws-amplify-react-native
-$ react-native link
 ```
 
 </div>


js/start.md
```diff
@@ -277,7 +277,6 @@ $ react-native init myReactNativeApp
 $ cd myReactNativeApp
 $ npm install --save aws-amplify
 $ npm install --save aws-amplify-react-native
-$ react-native link
```
elorzafe: You are right, it should `react-native link amazon-cognito-identity-js`

mlabieniec: I'm not sure this is correct. `react-native link` command should link any/all native modules that are installed. installing aws-amplify will install the dependency of amazon-cognito-identity-js which exposes a native module and coincidentally should be linked here.

PR #79, opened 2018-11-01
syntax error in example

jamesobd:
There's an extra parenthesis
The e param could be wrapped instead, like (e).

**Note**: Some of the arrow functions on this page are wrapping the single param and some are not. It's just a preferential decision of but it would be nice to keep all the examples consistent.


jamesobd:
The e param could be wrapped instead, like (e).
**Note**: Some of the arrow functions on this page are wrapping the single param and some are not.  It's just a preferential  decision of but it would be nice to keep all the examples consistent.

PR #92, opened 2018-11-04
Clarify source directory

miketheman:
Line ending added by GitHub editor.


js/start.md
```diff
@@ -349,7 +349,7 @@ Create the AWS backend resources and update the aws-exports.js file.
 $ amplify push
 ```
 
-> A configuration file (`aws-exports.js`) will be added to the source directory.
+> A configuration file (`aws-exports.js`) will be added to the `src/` directory.
```
elorzafe: That is not completely true, because you could change the `Source directory path`.
React-native doesn't have `./src` directory for example
Could be something like this? 
```suggestion
> A configuration file (`aws-exports.js`) will be added to the source directory. (`Source Directory Path` question during `amplify init`)
```

miketheman:
Thanks, accepted the change via GitHub's beta accept change feature!

PR #93, opened 2018-11-04
Bug fix - wrong names for props in Vue Auth

lucasmike:
Fixing names for props in Vue auth components. All these props are named ...Config rather than ...Options


haverchuck:
Thanks for submitting this @lucasmike!!! ... we got these fixes in with some responsive tables which should hopefully make it easier to read on mobile.

PR #94, opened 2018-11-04
Add iOS PubSub documentation and fixup Android PubSub

frankmuellr:


muellerfr:
@undefobj, fixed spelling errors. Should be ready to merge.

muellerfr:
@undefobj, I updated the credentialsprovider code, but I am not sure about the IAM policy changes. I was following what JS was doing. If we want to change this we should do it for both Android and iOS, and probably JS also.

undefobj:
> @undefobj, I updated the credentialsprovider code, but I am not sure about the IAM policy changes. I was following what JS was doing. If we want to change this we should do it for both Android and iOS, and probably JS also.

I think staying consistent across platforms is fine here, just linking to IoT is enough.

PR #100, opened 2018-11-06
Adding manual setup for Android SDK and removing old manual setup MDs

frankmuellr:


muellerfr:
@undefobj, @minbi and I discussed this. We felt that the Android SDK does not do anything special for logging like iOS does, so we did not think it was worth mentioning.

PR #103, opened 2018-11-06
Feat/rest awsmobileclient

undefobj:
Updates to the REST section for iOS to use the new AWSMobileClient with API Gateway. Includes instructions for both IAM as well as Cognito User Pools authorization methods.


ios/api.md
```diff
@@ -464,119 +464,77 @@ do {
...
-	So, for an app named `useamplify` with an API resource named `xyz123`, the path of the code file might be `./generated-src/xyz123useamplifyabcdClient.swift`. The API client name would be `xyz123useamplifyabcdClient`.
+So, for an app named `useamplify` with an API resource named `xyz123`, the path of the code file might be **./generated-src/xyz123useamplifyabcdClient.swift**. The API client name would be `xyz123useamplifyabcdClient` and you would use it in your code with `xyz123useamplifyabcdClient.registerClient()` and `xyz123useamplifyabcdClient.client()`.
 
-	- Find the resource name of your API by running `amplify status`.
-	- Copy your API client name to use when invoking the API in the following step.
 
+Find the resource name of your API by running `amplify status`. Copy your API client name to use when invoking the API in the following sections.
 
-8. Invoke a Cloud Logic API.
+#### IAM authorization
```
mlabieniec: Looks like you mostly lower case authorization, but in the above title is “Setup Your Backend”. Maybe standardize on one method for the titles?

undefobj: Is there a proper titling that we should do here? If so let's go through all of them and update.

PR #104, opened 2018-11-06
Turning Android language selector into language indicator for iOS and…

muellerfr:
… Android

Also removing unnecessary Android image asset and log statements.


muellerfr:
@mlab and @undefobj 

After this PR all iOS pages will indicate that the code is Swift, and Android pages will indicate Java. There's no choice to switch for now.

PR #107, opened 2018-11-06
Initial Auth section, with a few placeholders to be cleaned up. Also …

undefobj:
Initial Auth section, with a few placeholders to be cleaned up. Also a small code fix in API section to use AWSMobileClient when using IAM auth in AppSync


ios/authentication.md
```diff
@@ -1,344 +1,417 @@
...
-    ```bash
-    $ amplify push
-    ```
+If you are offline and make a service request, and your tokens are **NOT** valid, the service request will be blocked and notifications for `signedOutUserPoolsTokenInvalid` or `signedOutFederatedTokensInvalid` will be sent to the listener. In the case of the AppSync client this can be ignored and the queries will come from cache or mutations enqueued with credentials automatically refreshing upon reconnection. For all other services, if this happens and you are offline you should not make the service request until you come back online, at which point the `AWSMobileClient` will automatically re-enter the token refresh flow outlined above and then make the service call with the updated credentials.
 
-5. Follow the [Set up Email & Password Login](set-up-email-and-password) steps to connect to your backend from your app.
+## Automated Setup
 
-#### Facebook
+Run the following command in your project's root folder:
```
rohandubal: should we make a statement that we expect you amplify project to be initialized before running `amplify add auth` command?

ios/authentication.md
```diff
@@ -1,344 +1,417 @@
...
-    ```
-    $ amplify status
-    | Category  | Resource name   | Operation | Provider plugin   |
-    | --------- | --------------- | --------- | ----------------- |
-    | Auth      | cognitoa7cbb553 | Create    | awscloudformation |
-    ```
+If you are using both Cognito User Pools and Identity Pools, such as in Federated scenarios, you will need all of the keys mentioned above.
 
-8. To create your backend AWS resources run the following:
+## Initialization
```
rohandubal: I think we have missed the `Podfile` entries step before this step.

ios/authentication.md
```diff
@@ -1,344 +1,417 @@
...
-## Set Up Facebook Authentication
+```swift
+AWSMobileClient.sharedInstance().federatedSignIn(providerName: "graph.facebook.com", token: "FACEBOOK_TOKEN_HERE") { (userState, err)  in
+    if let error = error {
+        print("Federated Sign In failed: \(error.localizedDescription)")
+    }
+}
+```
+
+`federatedSignIn` can be used to obtain federated "Identity ID" using external providers like Google, Facebook or Twitter. If the tokens are expired and new tokens are needed, a notification will be dispatched on the `AWSMobileClient` listener with the user state `signedOutFederationTokensInvalid`. You can give the updated tokens via the same `federatedSignIn` method.
```
rohandubal: Can we add this statement to the end: 

The API calls to get `AWSCredentials` will be asynchronously blocked till you fetch the social provider token and give it to `AWSMobileClient`. Once you give the tokens, the `AWSMobileClient` will fetch `AWSCredentials` using the new tokens and unblock all waiting calls and use the new credentials.

ios/authentication.md
```diff
@@ -395,129 +468,174 @@ when configuring authentication using the AWS Amplify CLI.
...
+1. Add or update your AWS backend configuration file to incorporate your new sign-in. For details, see the last steps in the [Get Started: Set Up Your Backend](./add-aws-mobile-sdk-basic-setup) section.
+
+2. Add the following dependencies in your project's `Podfile`.
+
+	```ruby
+	platform :ios, '9.0'
+	  target 'YOUR-APP-NAME' do
+	    use_frameworks!
+	    pod 'AWSMobileClient', '~> 2.6.33'
+	    pod 'AWSFacebookSignIn', '~> 2.6.33'
```
rohandubal: We can callout in comments that `AWSFacebookSignIn`, `AWSAuthUI` is needed only for using drop in UI, else can be skipped.

ios/authentication.md
muellerfr: federatedSignIn() - be consistent in marking methods as such

muellerfr: same goes for signedOutFederationTokensInvalid()

ios/authentication.md
```diff
@@ -395,129 +468,174 @@ when configuring authentication using the AWS Amplify CLI.
...
-* `auth-google-create-oauth-android-clientid` (required for all Android apps)
+Note that the CLI allows you to select more than one identity provider for your app. You can also run `amplify auth update` to add an identity provider to an existing auth configuration.
 
-* `auth-google-create-oauth-ios-clientid` (required for all iOS apps)
+#### Google with Cognito Identity
 
+To federate Google as a user sign-in provider for AWS services called in your app, you will pass tokens to `AWSMobileClient.sharedInstance().federatedSignIn()`. You must first register your application with Google Sign-In and then configure this with Amazon Cognito Identity Pools.
 
-## Add sign-in
+The following sections detail the Google Sign-In requirements ans steps to integrate Google Sign-In for both iOS and Android apps:
```
muellerfr: ans => and

muellerfr: also should we talk about Android apps  in the iOS docs?

ios/authentication.md
undefobj: Good catch, added

ios/authentication.md
undefobj: done

ios/authentication.md
undefobj: Done

ios/authentication.md
undefobj: Changed for the federatedSignIn() however signedOutFederationTokensInvalid is a dispatched state and not a method.

ios/authentication.md
undefobj: I want to review the Android stuff in the second pass with Roshan's update.

PR #110, opened 2018-11-07
Android storage updates rev1

cbommas:
Review/Scrub of the android Storage section. Also, made some minor changes to the ios Storage section


muellerfr:
:shipit: Only one comment, should we link to the samples?

undefobj:
> :shipit: Only one comment, should we link to the samples?

Let's chat about this tomorrow. I'm ok with it if they're official samples that we're maintaining and tracking issues.

PR #114, opened 2018-11-08
Feat/auth android

undefobj:
Initial Android auth section. Still need to cleanup the Facebook and Google stuff at the end, will do that in next PR


android/api.md
```diff
@@ -1,836 +0,0 @@
-# API
```
muellerfr: Why are you deleting the api doc?

undefobj:
Will reopen

PR #124, opened 2018-11-09
Mlab/updates

mlabieniec:
- Minor theme updates
- Fix nav for cli / platforms
- Move get started page into docs
- notification for aws docs


_config.yml
```diff
@@ -357,7 +359,50 @@ ios_category_list:
...
+
+cli_category_list:
+  - category: Toolchain
+    title: Toolchain
+    excerpt: Better developer experience when building and deploying your apps.
+    base_url: '/cli/'
+    subs:
+      - title: Usage
+        excerpt: Setup and Configure a Project with the CLI
+        url: '/cli/init/?sdk=js'
```
muellerfr: Should this have the URL parameter sdk=js? What if I navigate from Android docs to some CLI category, and then click on another CLI category?

muellerfr: Seems like we should retain the previous URL parameter

mlabieniec: This list isn't actually used right now, i was only using it in experimentation rendering the list in places. But with this implementation this list listing isn't used.

PR #135, opened 2018-11-09
Auth set session manually

powerful23:


js/authentication.md
```diff
@@ -225,6 +225,78 @@ Auth.verifyCurrentUserAttributeSubmit(attr, 'the_verification_code')
...
+    credentialsDomain: 'www.example.com', // Optional, the domain used to get the credentials
+    creentialsToken: 'xxxxxxx' // Optional, the token used to get the credentials
+});
+
+// You can also get session, user, credentials from other methods:
+const session = await Auth.currentSession();
+const user = await Auth.currentAuthenticatedUser();
+const credentials = await Auth.currentCredentials();
+
+if (session.type && session.type === SessionType.Federated_Provider_Session) {
```
elorzafe: could `session` be `null`?

js/authentication.md
```diff
@@ -225,6 +225,78 @@ Auth.verifyCurrentUserAttributeSubmit(attr, 'the_verification_code')
...
+    username: 'Alice',       // Required, user name
+    attributes: {            // Optional, user attributes
+        email: 'alice@gmail.com',
+        phone_number: '123-456-7890'
+    },
+    tokens: {                 // Either id token or access token should be provided
+        idToken: 'xxxxxxx',  // Optional, the id token
+        accessToken: 'xxxxxxx', // Optional, the access token
+        refreshToken: 'xxxxxxx', // Optional, the refresh token
+        expires_at: 1540857110755, // Required, the timestamp when the token expires
```
undefobj: If you provide two tokens, which `expires_at` is this for?

js/authentication.md
```diff
@@ -225,6 +225,78 @@ Auth.verifyCurrentUserAttributeSubmit(attr, 'the_verification_code')
...
+    },
+    provider: AWSCognitoProvider.NAME, // Required, the name of the token provider
```
undefobj: Need to make this provider generic for more than just Cognito

engharb: AWSCognitoProvider Does not exist?! 
my aws-amplify: 1.1.17

svetozar02:
Any updates on this PR? We really need it. Not being able to user social login on react native is a pain

RyanRTS:
Also looking for an update on this, it would really help us out

JohnBRTS:
Any progress/updates on this PR? This would be a significant help in a problem we're currently dealing with in our app

js/authentication.md
powerful23: No, it will either return an object all throw an error.

js/authentication.md
powerful23: The format has been changed.

DavidvanZyl:
Any news on this?

undefobj:
@DavidvanZyl @JohnBRTS @RyanRTS @svetozar02 we reviewed all of the feedback for this PR and related issues, and believe we can actually provide a more comprehensive solution. Could you please look at the following comment: https://github.com/aws-amplify/amplify-js/issues/825#issuecomment-468453597 and related PR & RFC to see if this will meet your needs? If not could you let us know why you still might require this solution? Thanks.

DavidvanZyl:
Looks like #2716 would do the job for me. My current issue is timing. I can get auth with the JS SDK and I _could_ get caching etc with AppSync/Amplify if only I could get the two to work together. Or (more topically) get federated login in my Ionic/Angular app with the latter, rendering the prior unnecessary. @undefobj  For my sanities sake please, as things stand, is it possible for me to make my own UI and have users log in anon and convert to authenticated with a social login and have the caching ability?

undefobj:
@DavidvanZyl sounds like you're looking for something different than this PR, which is offline caching of data from AppSync and using it Authenticated and UnAuthenticated flows. Am I misinterpreting that? If that's the case I would suggest looking at this issue and the linked sample repo: https://github.com/aws-amplify/amplify-js/issues/2363#issuecomment-464523316

DavidvanZyl:
@undefobj It's my understanding that an amplify api will automatically handle caching data from requests? But thats kind of beyond the point of this I guess (I just hoped you might have a solution and save me from this 'mare) The problem as I've seen it from my looking around and trying various solutions is that: While using amplify, federated login does not work. I have therefore, tried to make use of the SDK to provide Auth alongside Amplify but there seems to be no way to integrate the two. Which is what this PR is about so far as Ive understood it. I cant get the data returned by the SDK in terms of IdentityId and AccessId into Amplify Auth to create a session which leaves me with a 'No user' style of error. _Ideally_ Amplify would let me use social logins, or at least let me create a session with data from the SDK so that I can use the api that amplify creates when you use the CLI (which again, I believe does caching). Does that make sense or am have I totally misunderstood something?

undefobj:
@DavidvanZyl 

> @undefobj It's my understanding that an amplify api will automatically handle caching data from requests? 

If by caching in the API category you mean for offline support, etc., then yes this is provided by the AppSync SDK for the Apollo client which is part of the "umbrella" Amplify framework. Please see the documentation here: https://aws-amplify.github.io/docs/js/api#aws-appsync-sdk

> While using amplify, federated login does not work. I have therefore, tried to make use of the SDK to provide Auth alongside Amplify but there seems to be no way to integrate the two. Which is what this PR is about so far as Ive understood it. I cant get the data returned by the SDK in terms of IdentityId and AccessId into Amplify Auth to create a session which leaves me with a 'No user' style of error. _Ideally_ Amplify would let me use social logins, or at least let me create a session with data from the SDK so that I can use the api that amplify creates when you use the CLI (which again, I believe does caching). Does that make sense or am have I totally misunderstood something?

I think there are a couple of things to tease out here. First, "Auth" is a category inside of Amplify which is comprised of libraries, CLI, toolchain, and UI components. What you are referencing above if the `Auth.federatedSignIn()` API of the Auth category in the library. I'm not sure where you're no user error is coming from, but I have a feeling you might be confusing IAM-based authorization on the API category (hence why you're looking at `federatedSignIn` and a Cognito Identity ID). This is all supported and functional today and unrelated to this PR. The current PR is for manually setting a session which many customers were looking to do as there was no Hosted UI federation capabilities with React Native, which per my other link is coming soon. I would suggest opening up a new issue with your use case and details of what you're trying to build so that someone from our oncall could address as you might be chasing something unnecessary here.

PR #151, opened 2018-11-10
Update iOS docs with titles

palpatim:
This PR will cause the iOS docs to have non-empty `<title>` tags. This will aid accessibility and SEO, as well as being good practice anyway.

### DIFF
--- a/ios/add-aws-mobile-user-sign-in-customize.md
+++ b/ios/add-aws-mobile-user-sign-in-customize.md
@@ -1,3 +1,7 @@
+---
+title: Customize the SDK Sign-In UI
+---
+
 # Customize the SDK Sign-In UI
 
 By default, the SDK presents sign-in UI for each sign in provider you enable in your Mobile Hub project (Email and Password, Facebook, Google) with a default look and feel. It knows which provider(s) you chose by reading the `awsconfiguration.json` file you integrated with your app.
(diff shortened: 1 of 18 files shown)


rromanchuk:
@muellerfr not to hijack this PR but just FYI this article https://github.com/aws-amplify/docs/pull/151/files#diff-128fcb47c6b6e4245a0e4b0af5b4cf98 is probably going to do more harm than good and should probably be purged completely.

It's a little too casual to suggest to fix an issue by adding an `NSAppTransportSecurity` exception, especially if the aws mobile ecosystem can't meet this requirement in 2019, although i'm pretty sure all aws certs are updated now. Ironically it's pretty high up in search results, so i can see developers having a valid ATS issue picking up some bad habits

muellerfr:
@rromanchuk, thanks for brining this to my attention. Will discuss with the team and provide an update.

palpatim:
@rromanchuk,

Thanks for pointing that out. That doc was written a while back, when ATS was in early adoption phases. Now that it's a more established pattern amongst developers, and AWS has ensured our services are ATS-compliant, we're removing the out-of-date page. #374

rromanchuk:
@palpatim @muellerfr  yeah makes since, i still had those keys and i was trying to figure out if it was safe now to ditch them, and ran into this first.  I killed them, everything was fine (s3/cognito/analytics) and out to production already. I think the certs were all updated a while back

PR #186, opened 2018-11-12
Moving js/graphql to cli/graphql and enabling redirect

muellerfr:
I tested the redirect locally


_config.yml
```diff
@@ -427,7 +428,7 @@ category_list:
           cta: Read more
           class: aws-icon-ps-60-network-nodes
           icon: '/images/icons/Misc/Tool.svg'
-          url: '/js/graphql'
+          url: '/cli/graphql'
```
apatel72001: Don't you need `?sdk=js` in here now?

muellerfr: Good catch.

PR #213, opened 2018-11-18
Update authentication.md

Gribbs:
Adding a separate example that adds the federated User to the User Pool. The current Auth.federatedSignIn() example in the documentation only adds the federated user to the Cognito Identity pool, not to the User Pool.

### DIFF
--- a/js/authentication.md
+++ b/js/authentication.md
@@ -444,7 +444,40 @@ const federated = {
 ReactDOM.render(<AppWithAuth federated={federated}/>, document.getElementById('root'));
 ```
 
-For React Native as well as web, you can also initiate a federated signin process by calling `Auth.federatedSignIn()` method with a specific identity provider in your code:  
+For React Native as well as web, you can also initiate a federated signin process by calling `Auth.federatedSignIn()` method with a specific identity provider in your code.
+
+Add a federated user to your User Pool:
+
+```javascript
+import { Auth } from 'aws-amplify';
+import awsconfig from '../aws-exports';
+
+
+// Retrieve active Google user session
+ga.signIn().then(googleUser => {
+    const { id_token, expires_at } = googleUser.getAuthResponse();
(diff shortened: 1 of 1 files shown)


powerful23:
@Gribbs I don't think you can run `Auth.federatedSignIn()` in that way. Did you test that? How did you configure your Cognito service to make it work?

Gribbs:
Absolutely I tested it. In fact, I think it’s the source of much confusion in the documentation

Gribbs:
For example, this issue here is having this problem https://github.com/aws-amplify/amplify-js/issues/1096
The solution “you have to use the hosted ui” is wrong. Actually, you have to call Auth.federatedSignIn() using the user pool as the provider. This same issue keeps popping up on the git issues I’ve found but the solution is not documented anywhere.
Please test this yourself by all means

Gribbs:
Several people on this thread (especially the last person) https://github.com/aws-amplify/amplify-js/issues/302 , same issue I think

powerful23:
@Gribbs I tested in my app and got this error message: `{"__type":"NotAuthorizedException","message":"Invalid login token. Issuer doesn't match providerName"}`. 
The method is aimed to create an identity in Cognito Federated Identity Pool and get a corresponding credentials and it has nothing to do with the Cognito User Pool. Can you provide the request and response context you made when calling this method?

Renato66:
just tested that method and it returns  
`Invalid login token. Issuer doesn't match providerName`

Gribbs:
ok I was using this method with a token returned from a google login via the hosted UI . I seem to be getting the same error as you @Renato66 if I just use the google login button. Confused much :(

dinamic:
@Gribbs is there a chance that have used identity pool before?

Gribbs:
I’ve confirmed that I was using a hosted UI configuration and not using any React HOC such as withOAuth() which worked but not quite as documented there in hindsight. My bad

PR #223, opened 2018-11-20
Enhance Cognito Hosted UI doc (WIP)

powerful23:


js/authentication.md
```diff
@@ -428,55 +428,251 @@ this._validAuthStates = ['signedIn'];
...
-```javascript
-const AppWithAuth = withAuthenticator(App);
+#### Federated Sign In
 
-const federated = {
-    google_client_id: '', // Enter your google_client_id here
-    facebook_app_id: '', // Enter your facebook_app_id here
-    amazon_client_id: '' // Enter your amazon_client_id here
-};
+(description of the feature):
```
powerful23: @undefobj More description about the Federation with Cognito Federated Identity Pool

undefobj: Let's use the following wording:

The `Auth.federatedSignIn()` is used to get AWS credentials directly from Cognito Federated Identities, which is different from Cognito User Pools. When an AWS service (such as S3) uses IAM for authorization, the request needs to be signed with AWS credentials and Cognito Federated Identities provides short term AWS credentials for performing this action using mobile or web applications. Amplify automatically refreshes these short term credentials in the background on your behalf, and when using `Auth.signIn()` you **do not** need to call  `Auth.federatedSignIn()` as this process happens automatically in the background for you. `Auth.signIn()` will also provide JWT OIDC tokens from Cognito User Pools which are federated with Cognito Federated Identities on your behalf allowing your application to interact with AWS services, which the other Amplify categories (such as Storage and API) will sign requests automatically.

In general, if you are using Cognito User Pools to manage user Sign-Up and Sign-In you do not need to call `Auth.federatedSignIn()` as this happens automatically behind the scenes when your User Pool is federated with an Identity Pool. You will be able to retrieve User Pool tokens with `Auth.currentSession` and the user object (from User Pols) with `Auth.currentAuthenticatedUser`. The AWS credentials can be found with `Auth.currentCredentials`.

PR #228, opened 2018-11-22
unneeded variable assignments

RaananW:
Once you set public/private in a TypeScript constructor, it is set as this.[variableName] in the constructed class. So there is no need to assign this.amiplifyService, as it is already set.


js/ionic.md
```diff
@@ -108,9 +108,7 @@ import { AmplifyService }  from 'aws-amplify-angular';
 
 export class AppComponent {
   
-  constructor( public amplify:AmplifyService ) {
```
mlabieniec: this would need to be defined as private in order to accessible in the other methods

js/ionic.md
```diff
@@ -140,8 +138,6 @@ import { AmplifyService }  from 'aws-amplify-angular';
   // ...
 constructor( public amplifyService: AmplifyService ) {
 
-    this.amplifyService = amplifyService;
```
mlabieniec: we probably should update this to be a private injection and reference it plainly here

RaananW: I doubt want to change the visibility of the variable, and I actually thought it should start public so not to break different projects unexpectedly. I'll change it to private

jordanranz:
Docs have been updated with this change already.

Thanks for contributing.

PR #247, opened 2018-11-28
Fix variable name in iOS DeltaSync example

palpatim:
Looks like we copy/pasted from Android, where we are indeed using a Map named `allPosts`. For iOS, we're using an array called `postList`.


palpatim:
Suggesting @muellerfr and @undefobj as reviewers.

PR #268, opened 2018-12-04
fixes(#1870)

haverchuck:
Previously updated the angular doc... however, it appears that the ionic doc is actually the one displayed on site.

### DIFF
--- a/js/ionic.md
+++ b/js/ionic.md
@@ -201,7 +201,9 @@ The component will emit two events:
  - `(picked)` - Emitted when an image is selected. The event will contain the `File` object which can be used for upload.
  - `(loaded)` - Emitted when an image preview has been rendered and displayed.
  - `path` - An optional S3 image path (prefix).
- - `storageOptions` - An object which is passed as the 'options' parameter to the .put request.  This can be used to set the 'level' of the objects being uploaded (i.e. 'protected', 'private', or 'public').
+ - `storageOptions` - An object passed within the ‘options’ property in the Storage.put request. This can be used to set the permissions ‘level’ property of the objects being uploaded i.e. ‘private’, ‘protected’, or ‘public’.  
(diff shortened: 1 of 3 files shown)


js/ionic.md
```diff
@@ -200,6 +200,8 @@ The component will emit two events:
 
  - `(picked)` - Emitted when an image is selected. The event will contain the `File` object which can be used for upload.
  - `(loaded)` - Emitted when an image preview has been rendered and displayed.
+ - `path` - An optional S3 image path (prefix).
+ - `storageOptions` - An object which is passed as the 'options' parameter to the .put request.  This can be used to set the 'level' of the objects being uploaded (i.e. 'protected', 'private', or 'public').
```
mlabieniec: Update text:

An object passed within the ‘options’ property in the Storage.put request. This can be used to set the permissions ‘level’ property of the objects being uploaded i.e. ‘private’, ‘protected’, or ‘public’.

js/ionic.md
```diff
@@ -237,7 +239,8 @@ To render the album, use *amplify-s3-album* component in your Angular view:
 </amplify-s3-album>
 ```
 
-`(selected)` event can be used to retrieve the URL of the clicked image on the list:
+- `options` - object which is passed as the 'options' parameter to the .get request.  This can be used to set the 'level' of the objects being requested (i.e. 'protected', 'private', or 'public'
```
mlabieniec: Format same as above

js/ionic.md
mlabieniec: Also add a learn more link to link to storage permissions info/docs here

haverchuck: Done!

js/ionic.md
haverchuck: Done!

PR #270, opened 2018-12-04
Update the aws_exports value to awsmobile

simoncaley:
Currently running through the Amplify - React & React Native documentation.
The exported value in the `aws-exports.js` file is always `awsmobile` and not `aws_exports`.
So if you follow the code examples this the code will break due the code using the `aws_exports`.

Happy to be wrong on this or if there is a better way to approach it :)


haverchuck:
The default exported object in the actual file is labeled awsmobile, but I do not believe it actually needs to be imported that way.  Making this change would require several other documentation changes.  I can see why in some ways it makes sense, but others might think that just using the file name is intuitive as well.  

If you definitely disagree, I'd encourage you to open an issue and request that we change this in the docs across the board.  That way the larger team and community can have a discussion about it.

PR #285, opened 2018-12-06
Add option to currentAuthenticated User

powerful23:
There is an option added to `currentAuthenticatedUser` to bypass the cache.

Now when calling `currentAuthenticatedUser`, it will first check if user data(attributes, mfa methods) is cached. If is, it will read from the cache and return. This avoids sending request to Cognito every time this method is called.

### DIFF
--- a/js/authentication.md
+++ b/js/authentication.md
@@ -242,9 +242,10 @@ You can call `Auth.currentAuthenticatedUser()` to get the current authenticated 
 ```javascript
 import { Auth } from 'aws-amplify';
 
-Auth.currentAuthenticatedUser()
-    .then(user => console.log(user))
-    .catch(err => console.log(err));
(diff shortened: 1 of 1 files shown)


js/authentication.md
```diff
@@ -242,9 +242,10 @@ You can call `Auth.currentAuthenticatedUser()` to get the current authenticated
 ```javascript
 import { Auth } from 'aws-amplify';
 
-Auth.currentAuthenticatedUser()
-    .then(user => console.log(user))
-    .catch(err => console.log(err));
+Auth.currentAuthenticatedUser({
+    bypassCache: false  // By default is false. If set to true, this call will send a request to Cognito to get the latest user data
```
jordanranz: Can you add OPTIONAL here.

PR #292, opened 2018-12-07
Add signIn options

powerful23:


powerful23:
@undefobj it's not published yet.

js/authentication.md
```diff
@@ -114,9 +114,12 @@ Sign in with user credentials:
 ```javascript
 import { Auth } from 'aws-amplify';
 
-Auth.signIn(username, password)
-    .then(user => console.log(user))
-    .catch(err => console.log(err));
+Auth.signIn({
+    username, // Required, the username
```
undefobj: I'd actually like to show the old way as the primary route because it will be what 90% of customers end up using. The new way is good for advanced use cases but is unnecessary complexity for most customers building apps. So let's have something like "Optionally, you can pass an object which has the username, password, and validationData which is sent to a PreAuthentication Lambda trigger..."

powerful23: Done.

PR #307, opened 2018-12-12
Update storage.md

philiiiiiipp:
I do believe that this should be the bucket name. If I give it the ARN it tries to use the ARN as the bucket name ( e.g. as the upload url ). I suppose the idea was to get the provider from the ARN but as we speak this is not done.

Wondering if this is a bug in the documentation or in the code.


haverchuck:
@philiiiiiipp  We're closing this for now since I believe bucket name is correct.  Thanks again for contributing.

PR #332, opened 2018-12-23
Add a note about setting the distribution location (react)

ryanjones:
If you follow the current getting started guide you'll end up with a 404 when you run `amplify publish`. This is because by default amplify init defaults to **build** instead of **dist**. The tutorial sets up **dist** as the folder that should contain the deployment artifacts.

I (blindly) hit enter for everything here when I ran `amplify init`:

image

```
> amplify --version
0.1.38
```

I can't seem to find how to run the docs locally to test (I tried bundle exec jekyll serve, but it blew up). I did run the spell/grammar check.


js/start.md
```diff
@@ -317,6 +317,10 @@ Create new AWS backend resources and pull the AWS services configuration into th
 $ amplify init
 ```
 
+<div id="react" class="tab-content">
+> When asked for the distribution directory (the directory that will be uploaded to S3), answer `dist`. If you miss this step you can run `amplify configure project` to change your dist directory setting.
```
UnleashedMind: react uses 'build' as the distribution folder.

ryanjones: Yes it does, but the getting started setup doesn't call out that it needs to be in the dist folder. If you go with the default build folder the quick start doesn't work and you end up with a 404. Can you call out what you'd like me to change here?

mlabieniec: @ryanjones it shouldn't need to be in the dist folder unless you are using some custom folder structure outside create-react-app? CRA puts the dist build in a `build` folder, which the CLI will use when publishing....is it possible that your build didn't get uploaded for some reason, potentially a cli bug or similar? Was your S3 bucket not populated?

PR #352, opened 2019-01-04
Adding documentation for feature #441

mikeparisstuff:
Adding documentation for a change to the GraphQL transform dealing with API key expiration dates.


cli/graphql.md
```diff
@@ -1169,6 +1169,92 @@ Here is a complete list of searchable operations per GraphQL type supported as o
 | Float | `ne`, `gt`, `lt`, `gte`, `lte`, `eq`, `range`      |
 | Boolean | `eq`, `ne`      |
 
+## AWS CloudFormation Template Parameters
```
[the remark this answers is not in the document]
kaustavghosh06: Sure, lets add it for now and then we can modify it once we have the per request billing feature out.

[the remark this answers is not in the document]
mikeparisstuff: updated.

js/api.md
```diff
@@ -124,20 +127,551 @@ When the update is complete, you can see the changes on your backend by visiting
...
+})
+```
+
+4. Run `ampify push`
+
+Amazon Elasticsearch domains can take a while to deploy. Take this time to read up on Elasticsearch to see what capabilities you are about to unlock.
+
+[Getting Started with Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/getting-started.html)
+
+4. After the update is compiles but before creating any objects, update your Elasticsearch index mapping.
```
nikhil-dabhade: After the update is compiles --> After the update is "complete" ?

js/api.md
```diff
@@ -124,20 +127,551 @@ When the update is complete, you can see the changes on your backend by visiting
...
+
+An index mapping tells Elasticsearch how it should treat the data that you are trying to store. By default if we create an object with field `"location": { "lat": 40, "lon": -40 }`, Elasticsearch will treat that data as an *object* type when in reality we want it to be treated as a *geo_point*. You use the mapping APIs to tell Elasticsearch how to do this.
```
nikhil-dabhade: By default if we create an object with field  --> (add comma) By default, if we create an object with field

cli/graphql.md
```diff
@@ -1169,6 +1187,126 @@ Here is a complete list of searchable operations per GraphQL type supported as o
...
+
+Much of the behavior of the GraphQL Transform logic is configured by passing arguments
+to the directives in the GraphQL SDL definition but certain other things are configured
```
undefobj: Let's update this slightly as we want customers to prefer the SDL. Something like this:

"Much of the behavior of the GraphQL Transform logic is configured by passing arguments to the directives in the GraphQL SDL definition, but certain things can be configured by passing parameters to the CloudFormation template itself as an escape hatch when functionality isn't in the Transformer."

mikeparisstuff: updated.

js/api.md
mikeparisstuff: updated.

mikeparisstuff:
> The docs look pretty great in general, however there is one blocker and that is the `API Category Project Structure` section and related content that is going into `..js/api.md`. This should go in the GraphQL Transformer and/or CLI documentation. The API guide is more targeted at the client interfaces, and we have already received customer feedback that docs are verbose in their current form. We only have Transformer/CLI information in there for getting started and then redirect for more info. Also this is just JS and we would need to duplicate for iOS/Android which is why we have the centralized CLI/Transformer docs.

I pulled the content from /api into /graphql page.

js/api.md
mikeparisstuff: updated.

PR #384, opened 2019-01-17
Update start.md

wmock561:
Changed the AppDelegate.swift code to use UIApplication.LaunchOptionsKey instead of the old UIApplicationLaunchOptionsKey.
Changed the runQuery() function to use cachePolicy: .returnCacheDataAndFetch. This was done because users who put all 3 functions in viewDidLoad() will not be able to fetch data likely due to a race condition.


muellerfr:
Hi @wmock561, thank you for the PR. We'll take a look.

PR #386, opened 2019-01-18
Update api.md

palpatim:
Use `UIApplication.LaunchOptionsKey` instead of `UIApplicationLaunchOptionsKey`


palpatim:
As noted in https://github.com/awslabs/aws-mobile-appsync-sdk-ios/issues/157

PR #389, opened 2019-01-19
Update api.md

palpatim:
Updated references to the deprecated `AWSAppSyncClientInfo` with `AWSAppSyncServiceConfig`.

See https://github.com/awslabs/aws-mobile-appsync-sdk-ios/issues/157


rohandubal:
@palpatim can we also include a change for the code snippets in this section: https://aws-amplify.github.io/docs/ios/api#cognito-user-pools?

PR #404, opened 2019-01-24
fix an error for ios pod dependency

sunchunqiang:


ios/storage.md
```diff
@@ -678,13 +678,15 @@ After you have a resolver for the mutation, to ensure that our S3 Complex Object
   $util.toJson($util.dynamodb.fromS3ObjectJson($context.source.file))
 ```
 The AWS AppSync SDK doesn't take a direct dependency on the AWS SDK for iOS for Amazon S3, but takes in :code:`AWSS3TransferUtility` and :code:`AWSS3PresignedURLClient` clients as part of AWSAppSyncClientConfiguration. The code generator used above for generating the API generates the Amazon S3 wrappers required to use the previous clients in the client code. To generate the wrappers, pass the :code:`--add-s3-wrapper` flag while running the code generator tool. You also need to take a dependency on the AWSS3 SDK. You can do that by updating your Podfile to the following:
-```
-  target 'PostsApp' do
+ 
+  ```ruby
+  target: 'PostsApp' do
     use_frameworks!
     pod 'AWSAppSync' ~> '2.9.0'
```
rohandubal:
```suggestion
    pod 'AWSAppSync', ~> '2.9.0'
```

sunchunqiang: changed

PR #408, opened 2019-01-24
Fixed incorrect method of passing of token

zantism:
.toString() doesn't return the actual token value, thus will cause unauthorized error if you call an API request with it.

.getTokenString() does return the actual token value.


muellerfr:
@zantism - Thank you for submitting this PR!

PR #415, opened 2019-01-29
Add command uses <category> add syntax

dirkgomez:
Fixed invocation of add command so that the example works


jordanranz:
The invocation in the docs:
`amplify add <category-name>`

Should work in addition to:
`amplify <category-name> add`

PR #423, opened 2019-02-04
Update authentication.md

meckhardt:


Petemir:
There's another typo, maybe you could add the fix to your PR so I do not make another one ? It says `retrieveing` instead of `retrieving`, in the code block for "Sign In" in the "Working with the API" section.

PR #430, opened 2019-02-05
Fixed inaccurate client reference

undefobj:
Replaced AWSMobileClient with Auth category

### DIFF
--- a/js/api.md
+++ b/js/api.md
@@ -1009,7 +1009,7 @@ export default withAuthenticator(App);
 
 ##### IAM Auth
 
-When using AWS IAM in a mobile application you should leverage Amazon Cognito Identity Pools. The Amplify CLI can automatically configure this for you when running `amplify add auth`. You can then use the `Auth` category for automatic credentials refresh [as outlined in the authentication section](./authentication) For manual configuration, add the following snippet to your `aws-exports.js` file:
+When using AWS IAM in a mobile application you should leverage Amazon Cognito Identity Pools. The Amplify CLI will automatically configure this for you when running `amplify add auth`. You can then use the `Auth` category for automatic credentials refresh [as outlined in the authentication section](./authentication) For manual configuration, add the following snippet to your `aws-exports.js` file:
 
 ```javascript
 import Amplify, { Auth } from 'aws-amplify';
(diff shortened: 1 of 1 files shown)


js/api.md
```diff
@@ -983,7 +983,7 @@ const client = new AWSAppSyncClient({
...
+Amazon Cognito User Pools is the most common service to use with AppSync when adding user Sign-Up and Sign-In to your application. If your application needs to interact with other AWS services besides AppSync, such as S3, you will need to use IAM credentials with Cognito Identity Pools. The Amplify CLI can automatically configure this for you when running `amplify add auth` and can also automatically federate User Pools with Identity Pools. This allows you to have both User Pool credentials for AppSync and AWS credentials for S3. You can then use the `Auth` category for automatic credentials refresh [as outlined in the authentication section](./authentication). For manual configuration, add the following snippet to your `aws-exports.js` file:
```
mlabieniec: this may cause confusion: "and can also automatically federate User Pools with Identity Pools"

not sure if that means i will get that with `amplify add auth` or if it's something else i need to do

js/api.md
```diff
@@ -1009,7 +1009,7 @@ export default withAuthenticator(App);
 
 ##### IAM Auth
 
-When using AWS IAM in a mobile application you should leverage Amazon Cognito Identity Pools. The Amplify CLI can automatically configure this for you when running `amplify add auth`. You can then use the `AWSMobileClient` for automatic credentials refresh [as outlined in the authentication section](./authentication) For manual configuration, add the following snippet to your `aws-exports.js` file:
+When using AWS IAM in a mobile application you should leverage Amazon Cognito Identity Pools. The Amplify CLI can automatically configure this for you when running `amplify add auth`. You can then use the `Auth` category for automatic credentials refresh [as outlined in the authentication section](./authentication) For manual configuration, add the following snippet to your `aws-exports.js` file:
```
mlabieniec: The Amplify CLI ***will*** automatically configure this for you when running `amplify add auth`

js/api.md
undefobj: Let's leave that as-is since it's not the thing that was updated and hasn't caused confusion yet.

js/api.md
undefobj: Updated

PR #432, opened 2019-02-06
Add section for Cognito Hosted UI on React Native

manueliglesias:


js/authentication.md
```diff
@@ -1366,6 +1369,49 @@ class CustomButton extends React.Component {
...
+      signOut,
+    } = this.props;
+
+    return (
+      <SafeAreaView style={styles.safeArea}>
+        {user && <Button title="Sign Out" onPress={signOut} icon='logout' />}
+        <ScrollView contentContainerStyle={styles.scrollViewContainer}>
+          <Text>{JSON.stringify({ user, error, }, null, 2)}</Text>
+          {!user && <React.Fragment>
+            <Button title="Cognito" onPress={hostedUISignIn} />
```
undefobj: Let's break this into two pieces:
1. Show how to launch the default hosted UI with `hostedUISignIn`
2. Show how you can skip the hosted UI and redirect straight to one of the social providers. Explain that this is for building a custom UI with buttons if you do not want to show the Cognito UI however it will still create a User Pool entry once the OAuth redirect has completed.

manueliglesias: Addressed

PR #433, opened 2019-02-07
Use cacheConfiguration

palpatim:
Use AWSCacheConfiguration instead of the deprecated databaseURL, in advance of the AppSync iOS 2.10.0 release.

@muellerfr Please don't merge this until we release AppSync iOS 2.10.0. I'm just getting this on your radar so you can review in advance of the release.


palpatim:
AppSync 2.10.0 is now released and this change can be safely merged.

PR #458, opened 2019-02-13
"hideAllDefaults" should be "hideDefaults"

wardellc:
"hideAllDefaults" does not hide the default options. It should be "hideDefaults". E.g. like this: https://github.com/aws-amplify/amplify-js/issues/1911#issuecomment-437337399


elorzafe:
@wardellc thanks for your contribution, but this is only for vue and other frameworks use `hideAllDefaults`, we are going to send a PR to make it backwards compatible. But for all frameworks should be the same.

PR #470, opened 2019-02-19
Update documentation on graphQL transform

vparpoil:
@model directive arguments were wrong. Subscriptions were also missing. 
See https://github.com/aws-amplify/amplify-cli/issues/861 for other possible improvements of the documentation on this specific part


vparpoil:
After some days wandering arround, I'm still not really sure how you are supposed to do to disable mutation and susbcription generation : 
does this work ?
```
type Post @model(queries: { get: "post" }, mutations: null, subscriptions: null)
```
I have tested this second approach which do disable unwanted generations but seems quite verbose : 
```
type Post @model(queries: {get: "post"}, mutations: {create:null, update:null, delete: null}, subscriptions: { onCreate:null})
```

PR #487, opened 2019-02-22
Added missing import for buildMutation

ffxsam:
*Issue #, if available:*

*Description of changes:*
Function was used without an import, so it wasn't clear where it was coming from.


js/api.md
```diff
@@ -895,6 +895,7 @@ For example, the below code shows how you would update the `CreateTodoMutation`
 An example of using the `buildMutation` helper to add an item to the cache:
 
 ```javascript
+import { buildMutation } from 'aws-appsync/lib/helpers/offline';
```
powerful23: @ffxsam hi, can you change this to be `import { buildMutation } from 'aws-appsync';` because it is exported by this line: https://github.com/awslabs/aws-mobile-appsync-sdk-js/blob/master/packages/aws-appsync/src/index.ts#L12

ffxsam: Done!

PR #490, opened 2019-02-24
Be explicit about data types for atts and metrics

roborourke:
Right now there’s no direct mention that there’s an expected data type for the attributes and metrics for an event to be recorded and not fail.

*Issue #, if available:*

*Description of changes:*


roborourke:
Thanks :)

PR #544, opened 2019-03-15
Docs for new Hub refactor

undefobj:
Docs for PR: https://github.com/aws-amplify/amplify-js/pull/2850
cc @jordanranz @elorzafe @haverchuck @manueliglesias @powerful23

*Description of changes:*
- Documentation for https://github.com/aws-amplify/amplify-js/pull/2850

### DIFF
--- a/js/hub.md
+++ b/js/hub.md
@@ -2,11 +2,9 @@
 ---
 # Hub
 
-AWS Amplify has a local event bus system called Hub. It is a lightweight implementation of Publisher-Subscriber pattern, and is used to share data between modules and components in your app.
+Amplify has a local eventing systen called Hub. It is a lightweight implementation of Publisher-Subscriber pattern, and is used to share data between modules and components in your app. Amplify uses Hub for different categories to communicate with one another when specific events occur, such as authentication events like a user sign-in or notification of a file download.
 
 ## Installation
-
-Import:
 ```javascript
 import { Hub } from 'aws-amplify';
 
@@ -16,30 +14,20 @@ import { Hub } from '@aws-amplify/core';
 
 ## Working with the API
 
-### dispatch()
+### Listening for messages
 
-You can dispatch an event with `dispatch` function:
+`Hub.listen(channel: string | RegExp, callback)` is used to listen for messages that have been dispatched. You must provide either a named `channel` or a regular expression, along with a callback. In the case of a regular expression only dispatches which contain a `message` in their payload will be matched against your pattern. You can add multiple listeners to your application for different channels or patterns to listen for, or trap generic events and perform your own filtering.
+
 ```javascript
-Hub.dispatch('auth', { event: 'signIn', data: user }, 'Auth');
-```
-
-### listen()
-
-You can subscribe to a channel with `listen` function:
-```javascript
-import { Hub, Logger } from 'aws-amplify';
-
-const logger = new Logger('MyClass');
+import { Hub } from 'aws-amplify';
 
 class MyClass {
     constructor() {
-        Hub.listen('auth', this, 'MyListener');
-    }
-
-    // Default handler for listening events
-    onHubCapsule(capsule) {
-        const { channel, payload } = capsule;
-        if (channel === 'auth') { this.onAuthEvent(payload); }
+        Hub.listen('auth', (data) => {
+            const { channel, payload } = data;
+            if (channel === 'auth') { this.onAuthEvent(payload); }            
+            console.log('A new auth event has happened: ', data.payload.data.username + ' has ' + data.payload.event);
+        })
     }
 
     onAuthEvent(payload) {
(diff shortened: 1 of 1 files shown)


js/hub.md
```diff
@@ -48,24 +36,79 @@ class MyClass {
...
+Amplify's `Auth` category publishes in the `auth` channel when 'signIn', 'signUp', and 'signOut' events happen. You can listen and act upon those event notifications.
 
 ```javascript
 import { Hub, Logger } from 'aws-amplify';
 
-const alex = new Logger('Alexander_the_auth_watcher');
+const logger = new Logger('My-Logger');
 
-alex.onHubCapsule = (capsule) => {
+logger.listener = (capsule) => {
```
manueliglesias:
```suggestion
const listener = (capsule) => {
```

js/hub.md
```diff
@@ -82,9 +125,119 @@ alex.onHubCapsule = (capsule) => {
     }
 }
 
-Hub.listen('auth', alex);
+Hub.listen('auth', logger.listener);
```
manueliglesias:
```suggestion
Hub.listen('auth', listener);
```

js/hub.md
```diff
@@ -48,24 +36,79 @@ class MyClass {
...
+
+### Authentication Events
+
+Amplify's `Auth` category publishes in the `auth` channel when 'signIn', 'signUp', and 'signOut' events happen. You can listen and act upon those event notifications.
 
 ```javascript
 import { Hub, Logger } from 'aws-amplify';
 
-const alex = new Logger('Alexander_the_auth_watcher');
+const logger = new Logger('My-Logger');
```
manueliglesias: I think this needs to be kept as `alex`

```suggestion
const alex = new Logger('My-Logger');
```

js/hub.md
```diff
@@ -16,30 +14,20 @@ import { Hub } from '@aws-amplify/core';
...
-### dispatch()
-
-You can dispatch an event with `dispatch` function:
-```javascript
-Hub.dispatch('auth', { event: 'signIn', data: user }, 'Auth');
-```
+### Listening for messages
 
-### listen()
+`Hub.listen(channel: string | RegExp, callback)` is used to listen for messages that have been dispatched. You must provide either a named `channel` or a regular expression, along with a callback. In the case of a regular expression only dispatches which contain a `message` in their payload will be matched against your pattern. You can add multiple listeners to your application for different channels or patterns to listen for, or trap generic events and perform your own filtering.
```
alebiavati: So if I understand correctly here we allow to filter by `message` with a regexp, but if they do that they cannot filter by channel. I think these two methods of filtering events could be confused as they are passed into `listen` via the same argument, but actually match different event fields.

Listen for channel:
```js
Hub.listen('mychannel') // --> matches the "channel" field
Hub.listen(/mymessage/) // --> matches the "message" field
```

I would expect the regexp to match channel name instead

js/hub.md
```diff
@@ -16,30 +14,20 @@ import { Hub } from '@aws-amplify/core';
...
-        Hub.listen('auth', this, 'MyListener');
-    }
-
-    // Default handler for listening events
-    onHubCapsule(capsule) {
-        const { channel, payload } = capsule;
-        if (channel === 'auth') { this.onAuthEvent(payload); }
+        Hub.listen('auth', (data) => {
+            const { channel, payload } = data;
+            if (channel === 'auth') { this.onAuthEvent(payload); }
```
alebiavati: checking the `channel` in the callback seems redundant since we are listening to a specific channel here.

js/hub.md
```diff
@@ -48,43 +36,208 @@ class MyClass {
...
+
+### Stop Listening
+
+Hub provides a way to stop listening for messages with `Hub.remove(channel: string | RegExp, listener: callback)`. This may be useful if you no longer need to receive messages in your application flow, as well as to avoid any memory leaks on low powered devices when you are sending large amounts of data through Hub on multiple channels.
+
+
+### Channels
+A channel is a logical group name that you use to organize messages and listen on. These are strings and completely up to you as the developer to define for dispatching or listening. However, while you can dispatch to any channel, ***Amplify protects certain channels*** and will flag a warning as sending unexpected payloads could have undesirable side effects (such as impacting authentication flows). The protected channels are currently:
+
+- core
```
alebiavati: I think we should implement a namespace for our channels to make it even harder for our customers to use one of our channels by mistake.

For example, the channel names could be like this: `@aws-amplify/core`, `@aws-amplify/auth`, etc.

alebiavati: then we could simply document that any channel that starts with `@aws-amplify/` is protected

js/hub.md
```diff
@@ -16,30 +14,20 @@ import { Hub } from '@aws-amplify/core';
...
-    }
-
-    // Default handler for listening events
-    onHubCapsule(capsule) {
-        const { channel, payload } = capsule;
-        if (channel === 'auth') { this.onAuthEvent(payload); }
+        Hub.listen('auth', (data) => {
+            const { channel, payload } = data;
+            if (channel === 'auth') { this.onAuthEvent(payload); }            
+            console.log('A new auth event has happened: ', data.payload.data.username + ' has ' + data.payload.event);
```
elorzafe:
```suggestion
            console.log('A new auth event has happened: ', payload.data.username + ' has ' + payload.event);
```

js/hub.md
undefobj: Not exactly, you can still filter by channel just by calling listener separately. It's just in a single API call you can do one or the other. We reviewed the API with customers and our DAs and received good feedback that it was clear.

js/hub.md
undefobj: Good point I can remove, It was in the old docs and just converted.

js/hub.md
undefobj: We talked through this during design as well and it seemed confusing especially for auth flows that are already using the current names. Doing the above would break existing apps too so it wouldn't be backwards compatible. We could look at it again as part of a major release though. Customers can still create independent Hub instances themselves or read/write to channels so that is essentially a namespace.

js/hub.md
undefobj: Fixed

PR #571, opened 2019-03-25
Corrected graphql.md

vsevolodbvd:
*Description of changes:*
Corrected graphql.md


cli/graphql.md
```diff
@@ -2561,7 +2561,7 @@ export class VersionedModelTransformer extends Transformer {
         // @versioned may only be used on types that are also @model
         const modelDirective = def.directives.find((dir) => dir.name.value === 'model')
         if (!modelDirective) {
-            throw new InvalidDirectiveError('Types annotated with @auth must also be annotated with @model.')
+            throw new InvalidDirectiveError('Types annotated with @versioned must also be annotated with @model.')
```
kaustavghosh06: This is meant to be @auth and not @versioned I beleive. cc: @mikeparisstuff

vsevolodbvd: @kaustavghosh06 If I understand correctly this block of code is related to the example of the @versioned transformer implementation. So, why we use @auth in this case?

PR #622, opened 2019-04-15
Adding doc for signing up with email or phone number

powerful23:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-js/pull/2991
*Description of changes:*


js/react.md
```diff
@@ -102,6 +102,111 @@ The signUpFields array in turn consist of an array of objects, each describing a
...
+export default withAuthenticator(App, { signUpConfig });
+```
+
+#### Sign up/in with email/phone number
+If the user pool is set to allow email addresses/phone numbers as the username, you can then change the UI components accordingly by using `usernameAttributes`.
+
+When you are using `email` as the username:
+```js
+import { withAuthenticator, Authenticator } from 'aws-amplfy-react';
+// import { withAuthenticator, Authenticator } from 'aws-amplify-react-native';
```
jordanranz: I think we should split up the react and react-native examples. Doing it all in one example is confusing to the customer

jordanranz: Also we should leave out commented code.

js/react.md
```diff
@@ -102,6 +102,111 @@ The signUpFields array in turn consist of an array of objects, each describing a
...
+```js
+import { Authenticator, withAuthenticator } from 'aws-amplfy-react';
+// import { Authenticator, withAuthenticator } from 'aws-amplify-react-native';
+
+class App {
+  // ...
+
+  render() {
+    return (
+      <Authenticator usernameAttributes='phone_number'/>
```
jordanranz: If we are using only the Authenticator, we should import withAuthenticator

PR #640, opened 2019-04-25
Adding @function directive docs

mikeparisstuff:
*Issue #, if available:*

Docs for https://github.com/aws-amplify/amplify-cli/pull/1346.

*Description of changes:*

Documentation for https://github.com/aws-amplify/amplify-cli/pull/1346.


cli/graphql.md
```diff
@@ -899,6 +899,100 @@ The generated resolvers would be protected like so:
...
+
+```javascript
+exports.handler = function (event, context) {
+  context.done(null, event.arguments.msg);
+};
+```
+
+**If you deployed your function using the 'amplify function' category**
+
+The Amplify CLI provides support for maintaining multiple environments out of the box. When you deploy a function via `amplify add function`, it will automatically add the environment suffix to your Lambda function name. For example if you create a function named **echo** using `amplify add function` in the **dev** environment, the deployed function will be named **echo-dev**. The `@function` directive allows you to use `${env}` to reference the current Amplify CLI environment.
```
plaa: Did you mean `a function named **echofunction**` and `**echofunction-dev**` (matching the `@function` declaration below and the function name in the next paragraph)?

cli/graphql.md
```diff
@@ -899,6 +899,100 @@ The generated resolvers would be protected like so:
...
+
+```
+type Query {
+  echo(msg: String): String @function(name: "echofunction-${dev}")
+}
+```
+
+**If you deployed your function without amplify**
+
+If you deployed your API without amplify then you must provide the full Lambda function name. If we deployed the same function with the name **echofunction** then you would have:
```
undefobj: Is it only name or also function ARN?

cli/graphql.md
```diff
@@ -899,6 +899,100 @@ The generated resolvers would be protected like so:
...
+  //   "typeName": "Query", /* Filled dynamically based on @function usage location */
+  //   "fieldName": "echo", /* Filled dynamically based on @function usage location */
+  //   "arguments": { /* GraphQL field arguments via $ctx.arguments */ },
+  //   "identity": { /* AppSync identity object via $ctx.identity */ },
+  //   "source": { /* The object returned by the parent resolver. E.G. if resolving field 'Post.comments', the source is the Post object. */ },
+  //   "request": { /* AppSync request object. Contains things like headers. */ },
+  //   "prev": { /* If using the built-in pipeline resolver support, this contains the object returned by the previous function. */ },
+  // }
+  context.done(null, event.arguments.msg)
+};
```
undefobj: We should show a full Lambda example where we hardcode the returned data like we do in this `Post` example: https://docs.aws.amazon.com/appsync/latest/devguide/tutorial-lambda-resolvers.html

cli/graphql.md
```diff
@@ -899,6 +899,100 @@ The generated resolvers would be protected like so:
...
+type Query {
+  echo(msg: String): String @function(name: "echofunction", region: "us-east-1")
+}
+```
+
+Calling functions in different AWS accounts is not supported via the @function directive but is supported by AWS AppSync.
+
+**Chaining functions**
+
+The @function directive supports AWS AppSync pipeline resolvers. That means, you can chain together multiple functions such that they are invoked in series when your field's resolver is invoked. To create a pipeline resolver that calls out to multiple AWS Lambda functions in series, use multiple `@function` directives on the field.
```
undefobj: Can you put a note about what happens under the covers? E.g. it's an AppSync Function corresponding to a Lambda function which are part of a Pipeline Resolver and called in left to right sequence based on how you use `@function` inline in your schema.

cli/graphql.md
```diff
@@ -899,6 +899,100 @@ The generated resolvers would be protected like so:
...
+}
+```
+
+In the example above when you run a mutation that calls the `Mutation.doSomeWork` field, the **worker-function** will be invoked first then the **audit-function** will be invoked with an event that contains the results of the **worker-function** under the **event.prev.result** key. The **audit-function** would need to return **event.prev.result** if you want the result of **worker-function** to be returned for the field.
+
+#### Generates
+
+The `@function` directive generates these resources as necessary:
+
+1. An AWS IAM role that has permission to invoke the function as well as a trust policy with AWS AppSync.
```
undefobj: Note that if the customer wishes to interact with other AWS resources using this function they will need to edit the IAM Execution Policy for that Lambda manually in the AWS Console.

BTW we should think about that in the future - it would be nice if we can let them augment in a friendly way within the CLI so that subsequent `amplify push` events don't blow away the updated policy. cc @kaustavghosh06

cli/graphql.md
mikeparisstuff: As of now it is **name** and **region** and the arn will be formatted for you. This does not support cross account access yet but can be added later.

cli/graphql.md
mikeparisstuff: Yes good catch thanks.

cli/graphql.md
mikeparisstuff: Ok, I will update the example.

mikeparisstuff: I updated this with an even more worth while example that shows how to fetch the current user from a cognito user pools which is a question that has come up multiple times in the github issues.

cli/graphql.md
mikeparisstuff: Provided a bit more detail.

cli/graphql.md
```diff
@@ -899,6 +899,242 @@ The generated resolvers would be protected like so:
...
+};
+```
+
+**If you deployed your function using the 'amplify function' category**
+
+The Amplify CLI provides support for maintaining multiple environments out of the box. When you deploy a function via `amplify add function`, it will automatically add the environment suffix to your Lambda function name. For example if you create a function named **echofunction** using `amplify add function` in the **dev** environment, the deployed function will be named **echofunction-dev**. The `@function` directive allows you to use `${env}` to reference the current Amplify CLI environment.
+
+```
+type Query {
+  echo(msg: String): String @function(name: "echofunction-${dev}")
```
yuth: echo(msg: String): String @function(name: "echofunction-${~~dev~~env}")

cli/graphql.md
```diff
@@ -1007,6 +1007,23 @@ exports.handler = async (event) => {
 
 > When deploying the function make sure you supply an environment variable named COGNITO_USERPOOL_ID with the value **UserPoolId** from **amplify-meta.json**
 
+When deploying you function make sure you have provided an execution role with permission to call the Amazon Cognito User Pools admin APIs. Attaching this policy to your function execution role will give you the permissions you need.
```
kvasukib: When deploying **you -> your**

cli/graphql.md
mikeparisstuff: Good catch.

cli/graphql.md
mikeparisstuff: thanks

cli/graphql.md
mwarger: Maybe tangential, but would there be a way to create a pipeline resolver with the default implementation as one of the steps?  Maybe I want to audit with a lambda after the normal VTL template has finished, or maybe do some related work with a lambda before allowing the default template to do it's thing.

cli/graphql.md
kanerogers: @undefobj Can you please explain what you mean by

> if the customer wishes to interact with other AWS resources using this function they will need to edit the IAM Execution Policy for that Lambda manually in the AWS Console.

If I have created the function within Amplify, am I not able to edit `{functionName}-cloudformation-template.json` to attach a PolicyDocument to the function there? Or does the `@function` directive somehow override this policy?

cli/graphql.md
davekiss: I would also make use of a "do your default thing, then run this lambda" or "run this lambda, then do your default thing" behavior

cli/graphql.md
```diff
@@ -899,6 +899,242 @@ The generated resolvers would be protected like so:
...
+  CONFIRMED
+  ARCHIVED
+  COMPROMISED
+  UNKNOWN
+  RESET_REQUIRED
+  FORCE_CHANGE_PASSWORD
+}
+```
+
+Next run `amplify push` and wait as your project finishes deploying. To test that everything is working as expected run `amplify api console` to open the GraphiQL editor for your API. You are going to need to open the Amazon Cognito User Pools console to create a user if you do not yet have any. Once you have created a user go back to the AppSync console's query page and click "Login with User Pools". You can find the **ClientId** in **amplify-meta.json** under the key **AppClientIDWeb**. Paste that value into the modal and login using your username and password. You can know run this query:
```
phishy: A small spelling error. It should be: "You can now run this query:"

cli/graphql.md
mwarger: @mikeparisstuff @undefobj Curious if there were any thoughts on this?

mikeparisstuff: @mwarger @davekiss I think this is a good idea and is one we have discussed before. This issue discusses handling pipeline support more generally and it would be great to get your feedback on how this can work: https://github.com/aws-amplify/amplify-cli/issues/1055.

.spelling
```diff
@@ -361,8 +361,7 @@ PAY_PER_REQUEST
 DynamoDBModelTableWriteIOPS
 DynamoDBModelTableReadIOPS
 ssn
-
-# cli/graphql.md
```
kaustavghosh06: Why do we not need this?

mikeparisstuff:
I added a simpler example for @function that does not require a change to the lambda execution policy.

mikeparisstuff:
Rebased to fix mergability.

PR #647, opened 2019-04-29
Fix Vue Auth navigation link

heitorlessa:
*Issue #, if available:*

*Description of changes:* This PR is a minor fix to the navigation so one doesn't miss `signUpField` which is the attribute they're looking for in the first place.

*Additional context:* When trying to customize Sign Up component with custom fields and attributes, the documentation links do not point to the correct anchor. That can mislead people from going straight into `signUpFields` where the wrong configuration will be silently accepted by the Sign Up component (another issue altogether).

### DIFF
--- a/js/vue.md
+++ b/js/vue.md
@@ -107,8 +107,8 @@ Config:
 | [confirmSignInConfig](#confirmsignin)   | object |
 | [confirmSignUpConfig](#confirmsignup)   | object |
 | [forgotPasswordConfig](#forgotpassword) | object |
-| [signInConfig](#signinconfig)           | object |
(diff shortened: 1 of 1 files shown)


heitorlessa:
Something indeed went wrong.. this is the commit that matters only: https://github.com/heitorlessa/docs/commit/fc8c7933332f32d9a3e944276f35d1730bcd8eac

Let me see if I can fix that or else I'll open up a new PR

heitorlessa:
done. The problem was when I switched to `dev` as the target branch for this PR. It brought up commits it didn't need to. 

@undefobj

PR #655, opened 2019-05-02
Multi-client support in aws-appsync sdk

manueliglesias:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/js/api.md
+++ b/js/api.md
@@ -804,6 +804,7 @@ When using the AWS AppSync SDK offline capabilities (e.g. `disableOffline: false
 
 - Error handling: (`callback`)
 - Custom storage engine (`storage`)
+- A key prefix for the underlying store (`keyPrefix`)
 
 **Error handling**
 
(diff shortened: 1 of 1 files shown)


js/api.md
```diff
@@ -880,6 +881,14 @@ const client = new AWSAppSyncClient({
 });
 ```
 
+
+**Key prefix**
+
+The `AWSAppSyncClient` persists its store data (e.g. cache) using `redux-persist`. Keys in the persisted store will be prefixed by the provided `keyPrefix`.
```
undefobj: Why do we need to mention redux persist? That's an implementation detail. Also call it a cache not a store please.

js/api.md
```diff
@@ -1519,6 +1528,78 @@ client.sync({
     });
 ```
 
+#### AWS AppSync Multi-Auth
+
+AWS AppSync can support multiple authentication modes on a single API. [LINK TO APPSYNC DOCS](http://docs). In order to use this feature with the `aws-appsync` SDK, you can create multiple instances of the client where each instance uses a different auth type.
```
undefobj: Same as the other PR - this should be _authorization_

js/api.md
```diff
@@ -1519,6 +1528,78 @@ client.sync({
...
+    type: AUTH_TYPE.AMAZON_COGNITO_USER_POOLS,
+    jwtToken: async () => (await Auth.currentSession()).getIdToken().getJwtToken(),
+    disableOffline: true,
+  }
+});
+```
+
+**Offline capabilities enabled**
+
+Multiple clients cannot share the same `keyPrefix` since it is used to separate each client's persisted data (e.g. cache). When using multiple clients, make sure that you provide a different `keyPrefix` in the `offlineConfig` object.
```
undefobj: Bold **cannot**

js/api.md
kvasukib: auth type -> authorization type. Is type or mode appropriate?

undefobj:
Looks good

PR #663, opened 2019-05-03
Adding example code for customized storage object of Auth

### LINKED ISSUE #35: Example class for Managing Security Tokens is incomplete or more information need to be added.
**Describe the bug**
In the section 
https://aws-amplify.github.io/docs/js/authentication#managing-security-tokens
that describes `class MyStorage` it is stated 

> you need to sync those items into the memory in this method

```
    // If the storage operations are async(i.e AsyncStorage)
    // Then you need to sync those items into the memory in this method
    static sync(): Promise<void>;
```
How is this supposed to be done?
You can't use _async storage_ unless you know how to do this. 
You can look for other solutions to bridge the gap between sync/async which means that you might not use the provided `sync()` method.
The general description of `class MyStorage` is straight forward.

I have found this
https://github.com/aws-amplify/amplify-js/blob/4644b432/packages/core/src/StorageHelper/reactnative.ts#L66
Would that serve as an example implementation? 
If so, then the documentation needs updating 

**Expected behavior**
A clear and concise example and description of what you are expected to do.

powerful23:
*Issue #, if available:*
Fixes #35 
*Description of changes:*
To add a detailed example showing how to make a customized storage object of Auth.


js/authentication.md
```diff
@@ -451,6 +451,79 @@ Auth.configure({
 });
 ```
 
+Here is the example of how to use AsyncStorage as your storage object which will show you how to sync items from AsyncStorage into Memory:
+```js
```
sammartinez: Non-blocking / NIT: can we change this to `javascript`

js/authentication.md
```diff
@@ -845,7 +918,7 @@ const oauth = {
   scope: ['phone', 'email', 'profile', 'openid', 'aws.cognito.signin.user.admin'],
   redirectSignIn: 'http://localhost:3000/',
   redirectSignOut: 'http://localhost:3000/',
-  responseType: 'code' // or token
+  responseType: 'code' // or 'token', note that REFRESH token will only generated when the responseType is code
```
sammartinez: Missing `be` in  `note that REFRESH token will only generated when the responseType is code` should be place before the `generated`.

Should read: note that REFRESH token will only **be** generated when the responseType is code

js/authentication.md
```diff
@@ -1060,6 +1133,7 @@ const oauth = {
 
     // 'code' for Authorization code grant, 
     // 'token' for Implicit grant
+    // Note that REFRESH token will only generated when the responseType is code
```
sammartinez: Same here as well from above comment

PR #689, opened 2019-05-24
Adding @key directive docs

mikeparisstuff:
*Description of changes:*

Documentation for the @key directive.


cli/graphql.md
```diff
@@ -412,6 +412,147 @@ type Subscription {
...
+```
+directive @key(fields: [String!]!, name: String, queryField: String) on OBJECT
+```
+
+**Argument**
+
+| Argument  | Description  |
+|---|---|
+| fields  | A list of fields in the @model type that should comprise the key. The first field in the list will always be the **HASH** key. If two fields are provided the second field will be the **SORT** key. If more than two fields are provided, a single composite **SORT** key will be created from `fields[1...n]`. All generated GraphQL queries & mutations will be updated to work with custom `@key` directives. |
+| name  | When omitted, specifies that the @key is defining the primary index. When provided, specifies the name of the secondary index. You may one have one @key without a name per @model type.  |
```
davekiss: `You may one have one @key without a name per @model type.` needs clarification

cli/graphql.md
```diff
@@ -412,6 +412,147 @@ type Subscription {
...
+
+```
+# Get customers by email.
+type Customer @model @key(fields: ["email"]) {
+    email: String!
+    username: String
+}
+```
+
+A @key without a *name* specifies the key for the DynamoDB table's primary index. You may only provide 1 @key without a *name* per @model type. The example above shows the simplest case where we are specifying that the table's primary primary index should have a simple key where the hash key is *email*. This allows us to get unique customers by their *email*.
```
davekiss: `we are specifying that the table's primary primary index` remove a `primary`

cli/graphql.md
```diff
@@ -412,6 +412,147 @@ type Subscription {
...
+```
+
+#### Evolving APIs with @key
+
+There are a few important things to think about when making changes to APIs using `@key`.
+
+When you face a situation where you need to enable a new access pattern or change an existing access pattern you should:
+
+1. Create a new index that enables the new or updated access pattern.
+2. If adding a @key with 3 or more fields, you will need to back-fill the new composite sort key for existing data. With a `@key(fields: ["email", "status", "date"])`, you would need to backfill the `status#date` field with composite key values made up of each object's *status* and *date* fields joined by a `#`. You do not need to backfill data for @key directives with 1 or 2 fields.
```
davekiss: Is there any written document to find out more about how to perform a backfill on existing data?

cli/graphql.md
```diff
@@ -412,6 +412,147 @@ type Subscription {
...
+3. Deploy your additive changes and update your application to use the new access pattern.
+4. Once you are certain that you do not need the old index, remove it's @key and deploy the API again.
```
davekiss: `it's` -> `its`

cli/graphql.md
```diff
@@ -412,6 +412,147 @@ type Subscription {
 }
 ```
 
+### @key
+
+The `@key` directive makes it simple to configure custom index structures for `@model` types. 
+
+DynamoDB is a distributed hash table that can execute efficient range queries on extremely large data sets but doing so effectively takes a bit of forethought. DynamoDB query operations may use at most two attributes to efficiently query data. The first query argument passed to a query (the partition key) must use strict equality and the second attribute (the sort key) may use gt, ge, lt, le, eq, beginsWith, and between. Despite these restrictions, DynamoDB can effectively implement a wide variety of access patterns that are powerful enough for the majority of applications.
```
undefobj: Please reword this first sentence to use language appropriate with the Dynamo messaging: https://aws.amazon.com/dynamodb/

undefobj: I don't believe "partition" key is used any longer by Dynamo they use "primary"

undefobj: Remove "Despite these restrictions"

cli/graphql.md
mikeparisstuff: Thanksx

cli/graphql.md
mikeparisstuff: Added clarification

cli/graphql.md
mikeparisstuff: Done

PR #707, opened 2019-05-31
Add try catch block around onCreate.

wlee221:
Try catch block needs to be added in the app (and not in the source code) because onCreate function is passed as a prop from the client app.

*Issue #, if available:* 3369

*Description of changes:*


mlabieniec:
@wlee221 the issue number referenced here doesn't seem correct? 3369

wlee221:
@mlabieniec 

I apologize, I meant to reference 3369 from amplify-js: [link](https://github.com/aws-amplify/amplify-js/issues/3369).

PR #710, opened 2019-06-03
Add default verification code length

janhesters:
Was relevant for me, found out by try and error. This might help future users.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/js/authentication.md
+++ b/js/authentication.md
@@ -387,6 +387,8 @@ Auth.verifyCurrentUserAttributeSubmit(attr, 'the_verification_code')
 });
 ```
 
+By default the verification code has consists of 6 digits.
+
 #### Retrieve Current Authenticated User
 
 You can call `Auth.currentAuthenticatedUser()` to get the current authenticated user object.


js/authentication.md
```diff
@@ -387,6 +387,8 @@ Auth.verifyCurrentUserAttributeSubmit(attr, 'the_verification_code')
 });
 ```
 
+By default the verification code has consists of 6 digits.
```
mlabieniec: @janhesters While i think this would be helpful, it may not be a good idea to add this since we can not guarantee it will always be this way. Perhaps it might make sense to link to service docs for further details?
https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-email-phone-verification.html

What do you think?

janhesters: Ahh dammit, I somehow missed your message. Well, I don't know, took me quite some time to figure the default code length out and I haven't seen it mentioned anywhere in the Cognito docs either.

PR #740, opened 2019-06-14
Add more description about onRegister

powerful23:
*Issue #, if available:*
Related to https://github.com/aws-amplify/amplify-js/issues/3460
*Description of changes:*


js/push-notifications.md
```diff
@@ -340,6 +341,8 @@ PushNotification.onNotificationOpened((notification) => {
 });
 ```
 
+Note: the `onRegister` handler will only be triggered when the token is generated or updated in the service side which means normally it will only be triggered when the app is installed and opened the first time.
```
dusan-dragon: Can you be more clear about `service side`?
What do you referring to as a `service side`?

PR #742, opened 2019-06-15
Update example code to work with amplify defaults

stefangomez:
*Issue #, if available:*
**edit** just found this: https://github.com/aws-amplify/amplify-cli/issues/1307

*Description of changes:*
New users just trying out a new project running through the example docs will encounter an error on the last step of `amplify publish` if they use all the amplify cli defaults. This change will make the webpack build dir coincide with the amplify-cli frontend-js defaults: https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-frontend-javascript/lib/framework-config-mapping.js


mlabieniec:
The default config (for plain javascript) is actually 'dist':
https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-frontend-javascript/lib/framework-config-mapping.js#L47

'build' is for react.

Issue #631, opened 2019-04-18
No left menu for cli items.


elliott-king:
**Describe the bug**
No left menu for cli items.

**To Reproduce**
1. Visit https://aws-amplify.github.io/docs/js/start
2. Click any left menu option that links to a js article (eg: "React & React Native")
3. Left menu now has submenu for headings on page.
4. Click any left menu option that links to a cli article (eg: "GraphQL Transform")
5. No submenu on left.

This seems to be true for IOS and android as well. This makes navigating through the long CLI docs _incredibly_ difficult. It is also impossible to see how the page is broken down.

**Expected behavior**
Left menu should have a submenu for all headings on a given cli page.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: both Firefox & Chrome

**Additional context**
It seems that the js divs get modified into "in-this-page js-sections" after they are clicked, and the cli stays a "section-head"

Thank you for your time.

PR #746, opened 2019-06-18
Left menu extended for cli options

elliott-king:
*Issue #, if available:*
#631 

*Description of changes:*
I modified the comparison that checks whether or not there should be a side menu.
`If page.url >= item.url` => `if item.url >= page.url`

When on a 'cli' page, the item url contains the sdk parameter. The page url is only 'cli/[page].'

An alternative solution would be to make the page url include the sdk parameter. However, I believe that the comparison was still incorrect regardless.

### DIFF
--- a/theme/_layouts/default.html
+++ b/theme/_layouts/default.html
@@ -53,8 +53,9 @@
 												<p class="gray-section-head hidden-xs">{{ category.title }}</p>
 											{% endif %}
 											<div class="feature-list">
-												{% for item in category.subs %}
-													{% if page.url contains item.url %}
+                                                {% for item in category.subs %}
+                                                    {% assign cleanItemUrl = item.url | split: '?' | first %}
+                                                    {% if page.url contains cleanItemUrl %}
 														<div class='in-this-page js-sections'></div>
 													{% else %}
 														<a href="{% if item.url_external  %}
@@ -84,7 +85,8 @@
 											{% endif %}
 											<div class="feature-list">
 												{% for item in category.subs %}
-													{% if page.url contains item.url %}
+                                                    {% assign cleanItemUrl = item.url | split: '?' | first %}
+                                                    {% if page.url contains cleanItemUrl %}
 														<div class='in-this-page js-sections'></div>
 													{% else %}
 														<a href="{% if item.url_external  %}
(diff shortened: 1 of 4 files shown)


theme/_layouts/default.html
```diff
@@ -54,7 +54,7 @@
 											{% endif %}
 											<div class="feature-list">
 												{% for item in category.subs %}
-													{% if page.url contains item.url %}
```
attilah: To keep the logical order or the contains check I'd suggest to update the changes to at all occurrences:
```
{% assign cleanItemUrl = item.url | split: '?' | first %}
{% if page.url contains cleanItemUrl %}
```

elliott-king:
I was initially hesitant to determine what should or should not be a header (and therefore a link). I wasn't sure that was my place to say. 

It seems that you are not allowed to have h5 headers in the navbar (?). It will not indent more than the previous header , and it seems to affect the following headers..?
See: 
Android -> authentication -> Using hosted ui ... -> using amazon cognito -> setting up oauth with ... 
CLI -> Usage -> Headless -> --yes flag (should be submenu of headless and everything above it)
CLI -> Graphql -> API category project structure -> common patterns -> filtering by type (should be submenu of previous header)

Additionally, an h4 header following an h2 header will be shown as if it were h3. They look fine on the page itself, but are not distinguished in the navbar.

Some thoughts: 
Several headers for GraphQL should be shortened, or downgraded.
- "Common patterns for the API category" is too long (it is cut off on my screen)
  - Every subheader for this is too long
- The same is a true for "AWS CloudFormation Template Parameters"
- Perhaps shorten them to "Common Patterns for APIs" and "Cloudformation Template"?

theme/_layouts/default.html
elliott-king: Done.

theme/_layouts/default.html
```diff
@@ -1,325 +0,0 @@
-<!DOCTYPE html>
```
kaustavghosh06: Why do we want to get rid of this file altogether?

elliott-king: Answered below.

elliott-king:
I messed up the commit history (presumably by not force pushing), and got a bit confused. I rewound it & re-committed everything. default.html has no longer been deleted.

kaustavghosh06:
@elliott-king We merged your changes as  a part of our latest CLI docs res-structure. Thanks for submitting a fix for the left menus.

elliott-king:
Ah. That's too bad, but I understand.

Quick note @attilah in `default.html`, the order of the comparison was switched, instead of adding a new line of logic.

PR #781, opened 2019-07-10
Update Analytics docs to add integration with Auth

royjit:


kvasukib:
@royjit The tests are failing. It could be due to spell check. Can you look at it.

PR #788, opened 2019-07-17
removed save flag

dabit3:
--save flag is no longer necessary since npm 5


undefobj:
@dabit3 looks like there are a couple conflicts. Can you resolve so we can merge?

dabit3:
@undefobj ok this should be gtg now.

undefobj:
@dabit3 thanks!

PR #801, opened 2019-07-23
Updating auto-refresh language

raybooysen:
From what I can see in the source code, the language here is confusing. When is "appropriate"? Does that mean just before it expires? When the user calls Auth.currentSession()?

I think it makes it clearer that the refresh will take place only when a user consuming the API makes calls into the API itself.

This is a recreation of #745 which I had to recreate as I deleted my fork.

### DIFF
--- a/js/authentication.md
+++ b/js/authentication.md
@@ -38,7 +38,7 @@ Below are the 3 most common Auth architectures using the Amplify Framework.
 	
 #### Simple Auth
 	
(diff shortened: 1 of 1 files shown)


raybooysen:
@mlabieniec apologies, had to recreate the PR.

mlabieniec:
Thanks @raybooysen :)

PR #802, opened 2019-07-23
Clarify no need for 'amplify codegen' in Android

palpatim:
*Issue #, if available:*

*Description of changes:*


android/start.md
```diff
@@ -107,6 +107,8 @@ $ amplify push
 
 Since you added an API the `amplify push` process will automatically enter the codegen process and prompt you for configuration. Accept the defaults which generate a `./app/src/main/graphql` folder structure with your statements. Run a **Gradle Sync** and **Build** your app, at which point the generated packages are automatically added to your project.
 
+> Note: It is not necessary to run `amplify codegen` for an Android project. Code generation is done by the Gradle build process.
```
desokroshan: `amplify codegen` is needed to update the schema.json and .graphql files in the app if schema is modified from the console at a later point of time. Should we capture that here?

undefobj: Do we have customers getting confused at either of these two points? The point of a "Getting Started" docs is to get customers going very quick, without overly verbose instructions. Perhaps we can make these two clarifications in the API category docs themselves which go into further detail?

desokroshan: Sounds good. Point is that this not captured anywhere in our docs and needs to be documented somewhere.

palpatim:
Thanks all. I'm going to withdraw this PR and put it on API docs instead per @undefobj suggestion.

PR #803, opened 2019-07-23
Fix spurious broken link report

palpatim:
Fix spurious broken link report by redirecting anchor of "API" menu to the site
root. This should not affect human users at all, since the anchor only exists
as a hover target to display the language-specific API menu options.

*Issue #, if available:*

*Description of changes:*


palpatim:
Testing locally to ensure correct behavior

PR #807, opened 2019-07-26
Snipped of code incorrect in social proof exemple

Lucas-Geitner:
There was one div too many.

*Issue #, if available:*
When copy pasting the code to test, they where an error with the following snippet

*Description of changes:*
Remove  a div so the code work on real application


undefobj:
@Lucas-Geitner thanks for the PR!

PR #823, opened 2019-08-02
Added docs for build script hooks.

visusnet:
*Issue #, if available:* https://github.com/aws-amplify/amplify-cli/pull/1951

*Description of changes:*
Added docs for the function build hooks as described in https://github.com/aws-amplify/amplify-cli/pull/1951.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -565,6 +565,9 @@ cleartext
 NSAllowsLocalNetworking
 quoteOfTheDay
 OneGraph
+transpile
+transpiling
+transpiled
 
 # android/analytics.md
 ActivityLifecycleCallbacks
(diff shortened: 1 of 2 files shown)


visusnet:
I have updated the docs again (last 2 forced pushes) for multiple reasons:
 - If you concatenate multiple things into a script definition like "npm run first; npm run second" and the latter command exits with exit code 0, the whole script will have the exit code 0. You need to use && instead of ;.
- I have updated the babel dependencies in the example to use the namespaced packages, e.g. `@babel/*`
- `--root-mode upward` is not necessary if you are already in the directory with the .babelrc file.
- Both, yarn and npm are supported.

visusnet:
I have also rebased master into this branch.

PR #824, opened 2019-08-02
Add a note explaining SDK versioning

palpatim:
Fixes https://github.com/aws-amplify/aws-sdk-ios/issues/1225

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/ios/start.md
+++ b/ios/start.md
@@ -45,6 +45,11 @@ pod install --repo-update
 
 Close your Xcode project and reopen it using `./YOUR-PROJECT-NAME.xcworkspace` file. Remember to always use `./YOUR-PROJECT-NAME.xcworkspace` to open your Xcode project from now on. **Build your Xcode project**.
 
+### AWS SDK Version vs. Semantic Versioning
+
+The AWS SDK for iOS does not follow semantic versioning. Instead, it increments the patch-level version for both backward-compatible bugfixes *and* non-breaking changes, and increments the minor version for breaking changes. Major version changes are rare, and usually indicate a dramatically different programming model.
+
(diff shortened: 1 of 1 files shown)


palpatim:
Also see https://github.com/aws-amplify/docs/pull/825

undefobj:
@palpatim @mlabieniec is this really needed at the start of the quickstart guide? If we need this info can we put it somewhere else or at the end? For instance we could put it here https://aws-amplify.github.io/docs/ios/manualsetup which we link to at the end of the quickstart.

The quickstart is supposed to be just that, a way to get started quickly. We keep on adding stuff that people have to read up front and it intimidates them and slows the on-boarding process.

palpatim:
> @palpatim @mlabieniec is this really needed at the start of the quickstart guide? If we need this info can we put it somewhere else or at the end? For instance we could put it here https://aws-amplify.github.io/docs/ios/manualsetup which we link to at the end of the quickstart.

@undefobj 

Fair point. The other place I could see this being useful would be the "setup/config" section. That's currently labeled in the nav bar as "Manual SDK Setup", and the page itself is titled "Setup Options for the SDK". What about moving it there as a sibling section to the CocoaPods, Carthage, etc docs? 

While we're at it, should we normalize the nav link and page titles to be consistent?

undefobj:
@palpatim 

> Fair point. The other place I could see this being useful would be the "setup/config" section. That's currently labeled in the nav bar as "Manual SDK Setup", and the page itself is titled "Setup Options for the SDK". What about moving it there as a sibling section to the CocoaPods, Carthage, etc docs?

Agree with this, let's do it.

> While we're at it, should we normalize the nav link and page titles to be consistent?

Yes sounds good, the only normalization that was done so far is the main sections (Auth, Storage, etc.) to align with our strategic direction. Everything else is probably up for grabs.

palpatim:
Closing in favor of #865

PR #836, opened 2019-08-05
add documentation for S3 Complex Object handling on iOS

drochetti:
*Issue #, if available:* [aws-amplify/aws-sdk-ios/issues/1308](https://github.com/aws-amplify/aws-sdk-ios/issues/1308)

*Description of changes:*

- this clarifies the confusion around using S3 Complex Objects with AppSync on iOS
- also fixes the build scripts to preview the doc site locally

### DIFF
--- a/ios/storage.md
+++ b/ios/storage.md
@@ -451,11 +451,11 @@ transferUtility.uploadUsingMultiPart(data:data,
 Note: Please review the documentation for [API](./api) before you proceed with the rest of this section. 
 {: .callout .callout--info}
 
-You can also upload and download Amazon S3 Objects using AWS AppSync, a GraphQL based solution to build data-driven apps with real-time and offline capabilities. Sometimes you might want to create logical objects that have more complex data, such as images or videos, as part of their structure.  _For example, you might create a Person type with a profile picture or a Post type that has an associated image_. You can use AWS AppSync to model these as GraphQL types. If any of your mutations have a variable with bucket, key, region, mimeType, and localUri fields, the SDK uploads the file to Amazon S3 for you.
+You can also upload and download Amazon S3 Objects using AWS AppSync, a GraphQL based solution to build data-driven apps with real-time and offline capabilities. Sometimes you might want to create logical objects that have more complex data, such as images or videos, as part of their structure.  _For example, you might create a Person type with a profile picture or a Post type that has an associated image_. You can use AWS AppSync to model these as GraphQL types. If any of your mutations have a variable with `bucket`, `key`, `region`, `mimeType`, and `localUri` fields, the SDK uploads the file to Amazon S3 for you.
 
 Attach the following policy to your IAM role to grant it programmatic read-write access to your bucket:
 
-```
+```json
 {
   "Version": "2012-10-17",
   "Statement": [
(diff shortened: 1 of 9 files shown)


ios/storage.md
```diff
@@ -728,19 +607,38 @@ After adding the new mutation in our operations file, we run the code generator
...
+
+There is an implementation in the iOS Test Suite that can be used as a reference: [aws-mobile-appsync-sdk-ios/AWSAppSyncTestCommon/S3ObjectWrapper.swift](https://github.com/awslabs/aws-mobile-appsync-sdk-ios/blob/master/AWSAppSyncTestCommon/S3ObjectWrapper.swift). That code can be used as drop-in implementation for most use cases.
+
+The mutation operation doesn't require any specific changes in method signature. It requires only an S3ObjectInput with `bucket`, `key`, `region`, `localUri`, and `mimeType`. Now when you do a mutation, it automatically uploads the specified file to Amazon S3 using the `AWSS3TransferUtility` client internally.
 
 ## Working with Pre-Signed URLS
 
-By default, all Amazon S3 resources are private. If you want your users to have access to Amazon S3 buckets
-or objects, you can assign appropriate permissions with an [IAM policy](http://docs.aws.amazon.com/IAM/latest/UserGuide/PoliciesOverview.html).
+By default, all Amazon S3 resources are private. If you want your users to have access to Amazon S3 buckets or objects, you can assign appropriate permissions with an [IAM policy](http://docs.aws.amazon.com/IAM/latest/UserGuide/PoliciesOverview.html).
```
palpatim: Let's add a reminder of, and a link to the docs about, the fact that the Amplify CLI requires transfers to be scoped to the `public/`, `private/`, or `protected` prefixes.

palpatim:
In addition, let's add something like this to the explanation of access levels in [this section](https://aws-amplify.github.io/docs/ios/storage#storage-access):

---

...which still has scoped permissions to just their objects.

#### Working with access levels in your app code

When you upload objects to the S3 bucket the Amplify CLI creates, you must manually pre-pend the appropriate access-level information to the object. The correct prefix will depend on the access level of the object.

* Public: Accessible by all users of your app. Files are stored under the `public/` path in your S3 bucket.
  * Example keys: `"public/my-file.txt"`, `"public/sub-directory/my-image.jpg"`
* Protected: Readable by all users, but writable only by the creating user. Files are stored under `protected/{user_identity_id}/` where the `user_identity_id` corresponds to the unique Amazon Cognito Identity ID for that user.
  * Example keys: `"protected/\(cognitoIdentityId)/my-file.txt"`, `"protected/\(cognitoIdentityId)/sub-directory/my-image.jpg"`
* Private: Only accessible for the individual user. Files are stored under `private/{user_identity_id}/` where the `user_identity_id` corresponds to the unique Amazon Cognito Identity ID for that user.
  * Example keys: `"private/\(cognitoIdentityId)/my-file.txt"`, `"private/\(cognitoIdentityId)/sub-directory/my-image.jpg"`

These keys must be prepended when you are uploading the object, and the same key must be specified as part of the object's key during download.

---

That's a little bit rough, but I think it starts to get the important ideas across:
- Amplify CLI creates policies with access levels
- The app developer is responsible for manually constructing the correct prefix
- The prefix must be specified on both upload and download

ios/storage.md
drochetti: Done in the latest commit. Let me know what you think.

ios/storage.md
```diff
@@ -728,26 +607,63 @@ After adding the new mutation in our operations file, we run the code generator
...
+```
+
+**Note:** These keys must be prepended when you are uploading the object, and the same key must be specified as part of the object's key during download.
 
-Alternatively, you can use pre-signed URLs to give your users temporary access to Amazon S3 objects.  When you create a pre-signed URL, you must provide your security credentials, specify a bucket name, an object key, an HTTP method, and an expiration date and time. The pre-signed URL is valid only for the specified duration.
+### Temporary Access
+
+However, what if you wanted to provide permissions temporarily, for example: _you want to share a link to file temporarily and have the link expire after a set time_. To do this using an IAM policy would require you to first setup the policy to grant access and then at a later time remember to delete the IAM policy to revoke access. 
 
 ### Building a Pre-Signed URL
```
palpatim: If we rename the whole section as I proposed above, we can remove this header.

ios/storage.md
```diff
@@ -728,26 +607,63 @@ After adding the new mutation in our operations file, we run the code generator
...
+extension AWSS3TransferUtility: AWSS3ObjectManager {
+    // ...
+}
+```
+
+There is an implementation in the iOS Test Suite that can be used as a reference: [aws-mobile-appsync-sdk-ios/AWSAppSyncTestCommon/S3ObjectWrapper.swift](https://github.com/awslabs/aws-mobile-appsync-sdk-ios/blob/master/AWSAppSyncTestCommon/S3ObjectWrapper.swift). That code can be used as drop-in implementation for most use cases.
+
+The mutation operation doesn't require any specific changes in method signature. It requires only an S3ObjectInput with `bucket`, `key`, `region`, `localUri`, and `mimeType`. Now when you do a mutation, it automatically uploads the specified file to Amazon S3 using the `AWSS3TransferUtility` client internally.
 
 ## Working with Pre-Signed URLS
```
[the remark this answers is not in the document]
drochetti: Much clearer separation indeed, I updated as per your suggestion

ios/storage.md
```diff
@@ -728,26 +607,63 @@ After adding the new mutation in our operations file, we run the code generator
...
-or objects, you can assign appropriate permissions with an [IAM policy](http://docs.aws.amazon.com/IAM/latest/UserGuide/PoliciesOverview.html).
+By default, all Amazon S3 resources are private. If you want your users to have access to Amazon S3 buckets or objects, you can assign appropriate permissions with an [IAM policy](http://docs.aws.amazon.com/IAM/latest/UserGuide/PoliciesOverview.html).
 
-However, what if you wanted to provide permissions temporarily, for example: _you want to share a link to file temporarily and have the link expire after a set time_. To do this using an IAM policy would require you to first setup the policy to grant access and then at a later time remember to delete the IAM policy to revoke access. 
+### Working with access levels in your app code
+
+When you upload objects to the S3 bucket the Amplify CLI creates, you must manually prepend the appropriate access-level information to the object. The correct prefix - `public/`, `protected/` or `private/` - will depend on the access level of the object as documented in the [Storage Access section](./storage#storage-access). For example:
```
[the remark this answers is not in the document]
drochetti: good catch, fixed

ios/storage.md
drochetti: done!

ios/storage.md
```diff
@@ -728,26 +607,63 @@ After adding the new mutation in our operations file, we run the code generator
...
+
+```swift
+let s3Object = S3ObjectInput()
```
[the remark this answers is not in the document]
drochetti: Good call, since the example mutates the properties, it needs to be `var` and not `let`

ios/storage.md
```diff
@@ -692,16 +571,16 @@ After you have a resolver for the mutation, to ensure that our S3 Complex Object
 ```
 The AWS AppSync SDK doesn't take a direct dependency on the AWS SDK for iOS for Amazon S3, but takes in `AWSS3TransferUtility` and `AWSS3PresignedURLClient` clients as part of AWSAppSyncClientConfiguration. The code generator used above for generating the API generates the Amazon S3 wrappers required to use the previous clients in the client code. To generate the wrappers, pass the `--add-s3-wrapper` flag while running the code generator tool. You also need to take a dependency on the AWSS3 SDK. You can do that by updating your Podfile to the following:
  
-  ```ruby
+```ruby
   target: 'PostsApp' do
     use_frameworks!
     pod 'AWSAppSync', ~> '2.9.0'
```
palpatim: Update AWSAppSync version to `2.14.1`

ios/storage.md
```diff
@@ -692,16 +571,16 @@ After you have a resolver for the mutation, to ensure that our S3 Complex Object
...
     pod 'AWSS3', ~> '2.9.0'
```
palpatim: Update AWSS3 version to `2.10.2`

ios/storage.md
```diff
@@ -732,25 +611,60 @@ After adding the new mutation in our operations file, we run the code generator
...
+### IAM Policy Based Permissions
+
+When you upload objects to the S3 bucket the Amplify CLI creates, you must manually prepend the appropriate access-level information to the `key`. The correct prefix - `public/`, `protected/` or `private/` - will depend on the access level of the object as documented in the [Storage Access section](./storage#storage-access). For example:
+
+```swift
+var s3Object = S3ObjectInput()
+// Accessible by all users
+s3Object.key = "public/myFile.txt"
+// Readable by all users, but writable only by the creating user
+s3Object.key = "protected/\(cognitoIdentityId)/myFile.txt"
```
undefobj: Could we put a link on how to get the IdentityID from AWSMobileClient? https://aws-amplify.github.io/docs/ios/authentication#utility-properties

drochetti:
> Made one small comment, however to approve need @mlabieniec to review the config.yml since that has been known to call production issues when changed in the past.

Thanks, I'll check with @mlabieniec 
I made the changes so the docsite could be tested locally. The previous configuration would result in errors when the `node_modules/` folder was present.

ios/storage.md
drochetti: good call, done

drochetti:
Changes to the `_config.yml` verified at my own GH Pages: https://drochetti.github.io/docs/

Issue #838, opened 2019-08-06
Wrong page link: Firebase credentials


beaurushton:
**Describe the bug**

This issue duplicates https://github.com/aws-amplify/docs/issues/475 which was auto closed.  The same issue still persists.

https://aws-amplify.github.io/docs/js/push-notifications#setup-for-android

Get your push messaging credentials for Android in Firebase console. Click here for instructions.
The link behind "Click here for instructions" is dead.

**To Reproduce**
Click on link
Observe 404

**Expected behavior**
Page found

**Screenshots**
N/A

**Desktop (please complete the following information):**
N/A

**Smartphone (please complete the following information):**
 N/A

**Additional context**
N/A

beaurushton:
Correct link looks to be https://aws-amplify.github.io/docs/android/push-notifications-setup-fcm

palpatim:
Fixed on #842

PR #842, opened 2019-08-07
Fix push notifications url

wizage:
*Issue #, if available:* N/A

*Description of changes:*
Push notifications URL for android was not correct so I updated to be pointing to the right location.


wizage:
Fixes #838

PR #843, opened 2019-08-07
Update GettingStarted for iOS - Xcode 10.2 and AppSync 2.14.0

lawmicha:
AppSync 2.12.0 with AWSCore 2.10.0 produces the error:
```
[!] CocoaPods could not find compatible versions for pod "AWSCore":
  In snapshot (Podfile.lock):
    AWSCore (= 2.10.2, ~> 2.10.0)

  In Podfile:
    AWSAppSync (~> 2.12.0) was resolved to 2.12.2, which depends on
      AWSCore (~> 2.9.0)

    AWSCore (~> 2.10.0)
```

Update AppSync to 2.14.0
Update Xcode version to 10.2 (reference from https://github.com/awslabs/aws-mobile-appsync-sdk-ios, appsync 2.14.0 requires Xcode 10.2)

I also added some minor changes on the rest of the page, please let me know if those are fine as well.


ios/start.md
```diff
@@ -10,7 +10,7 @@ Build an iOS app using the AWS Amplify CLI and the AWS SDK for iOS. The Amplify
 
 [Install and configure the Amplify CLI](..)
 
-[Install Xcode](https://developer.apple.com/xcode/downloads/) version 9.2 or later.
+[Install Xcode](https://developer.apple.com/xcode/downloads/) version 10.2 or later.
```
undefobj: Why is an Xcode upgrade needed?

lawmicha: i saw from here https://github.com/awslabs/aws-mobile-appsync-sdk-ios
Note: AWS AppSync uses Swift 4.2.1. Use Xcode 10.2 or greater to build.

ios/start.md
```diff
@@ -53,10 +52,9 @@ Create new AWS backend resources and pull the AWS services configuration into th
 ```bash
 $ cd ./YOUR_PROJECT_FOLDER
 $ amplify init        #accept defaults
-$ amplify push        #creates configuration file
```
undefobj: Push is needed to kick off a codegen

lawmicha: seems like on amplify cli version 1.7.1, amplify push has no changes, and amplify init already created the configuration file

undefobj: Config file is created yes but we need a push to do the codegen of GraphQL types from the SDL. It's also the flow we want to ensure a path to success for creation of the backend for someone getting started.

ios/start.md
undefobj: Hmm that might be wrong. @royjit what do you think?

ios/start.md
lawmicha: in Step 4: Add API and Database, yes, ie.
```
amplify add api
amplify push
```
but this one is just to create configuration file, and choose the AWS account, etc. with single `amplify init` command. when I tried to amplify push, there was nothing pending to be pushed

undefobj: Ok it looks like we had a CLI behavior change, this was not the case before.

ios/start.md
royjit: Looks like AWS SDK will work for Xcode >= 9.2 except for AppSync GraphQL which require >=10.2.

ios/start.md
undefobj: We should validate for Android as well

ios/start.md
undefobj: Is there a reason for this? It basically means any customer would need to default to the highest water mark when using multiple services.

ios/start.md
lawmicha: good idea, will report back

ios/start.md
lawmicha: I recall seeing one of the dependencies that AppSync uses required the later Swift version. For developers with Mac's on 10.12.6 (Sierra), they can only install up to XCode 9.2.1, so if they go to step 4 and add Appsync, it will fail. they will have to upgrade their OS and xcode to get step 4 working. I think we should just convey the information in step 4 rather than have them debug the problem, which is that you will need xcode 10.2 for AppSync. If they don't need to use appsync, they can move onto Next Steps (Auth, storage, etc..). what do you guys think?

palpatim: Another way to reframe the discussion is how consistent we want to be in our docs. We can either document the minimum required versions necessary for all supported SDKs to run (in this case, Xcode 10.2.1), or we can document the minimum for *most* of the SDKs, and add exceptions where needed. I'd favor consistency, and the newer version, since it reduces complexity and cognitive burden, but I'm unclear how many legacy customers that would impact, so I could be talked out of that.

lawmicha: thanks for all chiming in, I think we can close on this with a brief meeting from tim's last point. i'll set this up

PR #853, opened 2019-08-12
Updating authentication docs for using the withAuthenticator HOC with…

Ashish5591:
… React Native
*Description of changes:*

Adding documentation to address issues related to linking dependencies when using the withAuthenticator HOC with React Native:

https://github.com/aws-amplify/amplify-js/issues/3817

### DIFF
--- a/js/authentication.md
+++ b/js/authentication.md
@@ -559,6 +559,13 @@ To learn more about tokens, please visit [Amazon Cognito Developer Documentation
 
 For React and React Native apps, the simplest way to add authentication flows into your app is to use the `withAuthenticator` Higher Order Component.
 
+If you are using the `withAuthenticator` HOC with React Native, make sure to link `react-native-vector-icons` and launch your app again:
+
+```bash
+react-native link react-native-vector-icons
+react-native run-ios
+```
+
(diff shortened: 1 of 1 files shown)


js/authentication.md
```diff
@@ -559,6 +559,13 @@ To learn more about tokens, please visit [Amazon Cognito Developer Documentation
 
 For React and React Native apps, the simplest way to add authentication flows into your app is to use the `withAuthenticator` Higher Order Component.
 
+If you are using the `withAuthenticator` HOC with React Native, make sure to link `react-native-vector-icons` and launch your app again:
```
sammartinez: Would it make more sense to state `..make sure to link 'react-native-vector-icons' then launch your app:`, just thinking if a customer is just getting started with this and reading it they may not be launching the app with out this.

bluezebragames: I might nitpick it to `before launching your app:`

Ashish5591: Thanks for the feedback. I think that makes sense. However its very common for the app to be already launched before any changes are made. Usually you run `yarn start` or `react-native run-ios` with the initial project to get it started, and then start making changes. (This is how we have written our own getting started docs as well)

How about I change it to `..make sure to link 'react-native-vector-icons' and then (re)launch your app:` It covers both cases.

PR #868, opened 2019-08-24
Subscription @auth support and custom claims

undefobj:
Adding `@auth` support for subscriptions as well as custom claims for users and groups in JWTs.

### DIFF
--- a/cli-toolchain/graphql.md
+++ b/cli-toolchain/graphql.md
@@ -578,11 +578,12 @@ based on attributes found the parent type.
 directive @auth(rules: [AuthRule!]!) on OBJECT, FIELD_DEFINITION
 input AuthRule {
     allow: AuthStrategy!
-    ownerField: String # defaults to "owner"
-    identityField: String # defaults to "username"
-    groupsField: String
-    groups: [String]
-    operations: [ModelOperation]
+    ownerField: String # defaults to "owner" when using owner auth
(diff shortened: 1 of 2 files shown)


undefobj:
Addressing:
- https://github.com/aws-amplify/amplify-cli/issues/1043
- https://github.com/aws-amplify/amplify-cli/issues/1766

undefobj:
Tagging @SwaySway @yuth for reviews

cli-toolchain/graphql.md
```diff
@@ -914,255 +913,103 @@ mutation CreateDraft {
...
-        }
-	}
-}
-```
+Finally, if you use both owner and group authorization then the username argument becomes optional. This means the following:
 
-Lastly, add the resolver mapping templates to the API project's `resolvers/` directory.
+- If you don’t pass the user in, but are a member of an allowed group, the subscription will notify you of records added.
+- If you don’t pass the user in, but are NOT a member of an allowed group, the subscription will fail to connect.
+- If you pass the user in who IS the owner but is NOT a member of a group, the subscription will will notify you of records added of which you are the owner.
```
onlybakam: _typo (double 'will'):_ If you pass the user in who IS the owner but is NOT a member of a group, the subscription will **will** notify you of records added of which you are the owner

undefobj: Fixed!

PR #878, opened 2019-08-29
Update storage.md

TrekSoft:
Modifies documentation to match CLI

*Description of changes:*
The documentation says to select "read/write" but the CLI displays "create/update" instead so modifies it to match.


TrekSoft:
Hmm, in my CLI I only saw create/update

TrekSoft:
Ahh, I see what you're saying now - I though it was a radio selection list but now see it's a checkbox list. Will update the CR here

PR #893, opened 2019-09-04
Adds warning about account activation

TrekSoft:
Users trying to run Amplify init will get an ugly and obscure error if their AWS account hasn't activated yet which can take up to 24 hours to happen. Adds a callout which informs users of this and what they can do to speed that process up.

This is what it looks like: https://drive.corp.amazon.com/documents/ddaudeli@/activationCallout.png

And this is the page it links to: https://aws.amazon.com/premiumsupport/knowledge-center/create-and-activate-aws-account/

### DIFF
--- a/theme/_layouts/index.html
+++ b/theme/_layouts/index.html
@@ -36,7 +36,7 @@
                                 </a>
                             </div>
                             <img id="aws-logo" class="hidden-xs" src="{% if jekyll.environment == 'production' %}{{ site.amplify.baseurl }}{% endif %}/images/Logos/aws_logo.png" />
-			    <div class="full-width callout callout--action">Note: It may take up to 24 hours for your AWS account to activate. You'll get an error creating an Amplify project if it is not activated. Visit <a href="https://aws.amazon.com/premiumsupport/knowledge-center/create-and-activate-aws-account/">this page</a> for information on getting your account activated sooner.</div>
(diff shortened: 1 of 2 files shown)


theme/_layouts/index.html
```diff
@@ -36,6 +36,7 @@ <h1>Get Started</h1>
                                 </a>
                             </div>
                             <img id="aws-logo" class="hidden-xs" src="{% if jekyll.environment == 'production' %}{{ site.amplify.baseurl }}{% endif %}/images/Logos/aws_logo.png" />
+			    <div class="below-registration-button callout callout--action">Note: It may take up to 24 hours for your AWS account to activate. You'll get an error creating an Amplify project if it is not activated. Visit <a href="https://aws.amazon.com/premiumsupport/knowledge-center/create-and-activate-aws-account/" target="_blank">this page</a> for information on getting your account activated sooner.</div>
```
palpatim: 1. Fix indentation
1. Run this by @undefobj to confirm placement. As a "getting started" page, it doesn't really seem appropriate to be putting troubleshooting information here. That said, if we're getting a lot of customer feedback about the account activation error, it may make sense to hoist it higher. My instinct is that this probably needs to be closer to the step-by-step instructions.

undefobj: I haven't heard any reports of this yet, though it does occasionally happen. I would suggest not merging this change unless the issue starts to become more prevalent.

TrekSoft: I counted 11 reports of it between these two issues:

- https://github.com/aws-amplify/amplify-cli/issues/191
- https://github.com/aws-amplify/amplify-cli/issues/972

undefobj: I think 8 reports over a year+ is a minor thing. This is ok to leave out for now.

TrekSoft: Ok sounds good - closing out

PR #902, opened 2019-09-09
Update prediction.md reference to Pinpoint

jodion:
Changing the reference of Pinpoint in the manual setup section to Cognito Federated Identities.

*Issue #, if available:*

*Description of changes:*
There was a typo referring to Pinpoint in the documentation. I believe it should be about Cognito Federated Identities.

### DIFF
--- a/js/predictions.md
+++ b/js/predictions.md
@@ -65,7 +65,7 @@ Amplify.addPluggable(new AmazonAIPredictionsProvider());
 
 #### Manual Setup
 
-The manual setup enables you to use your existing Amazon Pinpoint resource in your app.
+The manual setup enables you to use your existing Amazon Cognito Identity Pool resource in your app.
 
 ```javascript
 import Amplify from 'aws-amplify';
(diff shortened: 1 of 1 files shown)


js/predictions.md
```diff
@@ -65,7 +65,7 @@ Amplify.addPluggable(new AmazonAIPredictionsProvider());
 
 #### Manual Setup
 
-The manual setup enables you to use your existing Amazon Pinpoint resource in your app.
+The manual setup enables you to use your existing Amazon Cognito Identity Pool resource in your app.
```
elorzafe:
```suggestion
The manual setup enables you to use your existing Amazon AI and ML resources in your app.
```

elorzafe: Thanks for pointing this out, the idea is that if you have already configured permissions (IAM roles and policies on your identity pool to have access to AI services) you can manually configure those on the library

jodion: That makes sense. I was using Cognito as that was the only configurable parameter in the code section below, but if you are using Rekognition, you may want to point to your own face library and Cognito would be used from a different category. So what you wrote is better.

elorzafe: @jodion can you commit the suggestion so I can merge the pr. Thanks!

PR #906, opened 2019-09-10
Update the "GraphQL from Lambda" section to reflect multi-auth changes

kaustavghosh06:
*Issue #, if available:*

*Description of changes:*


cli-toolchain/quickstart.md
```diff
@@ -272,30 +272,20 @@ Behind the scenes, the CLI automates populating of the resource identifiers for
...
+    ]
+) {
   id: ID!
   name: String
   description: String
 }
 ```
 
-Next, run `amplify console` and select the GraphQL API. On the **Settings** page add an **Additional authorization provider** and select **IAM**. Next, on the **Schema** page add in the `@aws_iam` directive where appropriate per the [AppSync documentation](https://docs.aws.amazon.com/appsync/latest/devguide/security.html#using-additional-authorization-modes). For example if you just want your Lambda function to have access to run a single mutation, add the directive onto that mutation (`createTodo` below) as well as the return type:
+In the above example we want your Lambda function to have access to run a single mutation (`createTodo`) and hence we explicitly mention `create` in the `operations` list.
```
undefobj: Can you add to the end of this something like "To grant access for application users to perform other actions, such as read only you would add another auth rule. See XXX link for more examples"

kaustavghosh06: Added.

PR #914, opened 2019-09-13
Added documentation tied to codegen without init

nikhname:
*Issue #, if available:*

*Description of changes:*
Updated general usage and added a fifth workflow.

### DIFF
--- a/cli-toolchain/graphql.md
+++ b/cli-toolchain/graphql.md
@@ -2291,13 +2291,13 @@ You should have newly generated GraphQL statements and Swift code that matches t
 
 **Flow 5: Introspection Schema outside of an initialized project**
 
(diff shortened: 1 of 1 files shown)


cli-toolchain/graphql.md
```diff
@@ -2289,6 +2289,26 @@ $amplify codegen types
 ```
 You should have newly generated GraphQL statements and Swift code that matches the schema updates. If you ran the second command your types will be updated as well. Alternatively, if you run `amplify codegen` alone it will perform both of these actions.
 
+**Flow 5: Introspection Schema outside of an initialized project**
+
+If you would like to generate statements and types without initializing an amplify project, you can do so by providing your introspection schema named `schema.json` in your project directory and adding codegen from the same directory.
```
kaustavghosh06: How do we fetch the schema.json file from the AppSync Console?

nikhname: Adding instructions for this

PR #920, opened 2019-09-16
Add more clarification around @auth and permitted operations behavior

kaustavghosh06:

### DIFF
--- a/cli-toolchain/graphql.md
+++ b/cli-toolchain/graphql.md
@@ -647,8 +647,8 @@ type Todo
 }
 ```
 
-In this schema, only the owner of the object has the authorization to perform read (getTodo and listTodos) operations on the owner created object. But this does not prevent any other owner to update/delete some other owner's object. 
-Here's a truth table for the above-mentioned schema:
+In this schema, only the owner of the object has the authorization to perform read (getTodo and listTodos) operations on the owner created object. But this does not prevent any other owner (any user other than the creator or owner of the object) to update/delete some other owner's object. 
+Here's a truth table for the above-mentioned schema. In the table below `other` refers to any user other than the creator or owner of the object.
 
 |       | getTodo | listTodo | createTodo | updateTodo | deleteTodo |
 |-------|---------|----------|------------|------------|------------|
@@ -668,7 +668,7 @@ type Todo
 }
 ```
 
-Here's a truth table for the above-mentioned schema:
+Here's a truth table for the above-mentioned schema. In the table below `other` refers to any user other than the creator or owner of the object.
 
 |       | getTodo | listTodo | createTodo | updateTodo | deleteTodo |
 |-------|---------|----------|------------|------------|------------|
@@ -1236,7 +1236,7 @@ type Todo
 }
 ```
 
-In this schema, only the owner of the object has the authorization to perform update operations on the `content` field. But this does not prevent any other owner to update some other field in the object owned by another user. If you want to prevent update operations on a field, the user would need to explictly add auth rules to restrict access to that field. One of the ways would be to explictly specify @auth rules on the fields that you would want to protect like the following:
+In this schema, only the owner of the object has the authorization to perform update operations on the `content` field. But this does not prevent any other owner (any user other than the creator or owner of the object) to update some other field in the object owned by another user. If you want to prevent update operations on a field, the user would need to explictly add auth rules to restrict access to that field. One of the ways would be to explictly specify @auth rules on the fields that you would want to protect like the following:
 
 ```graphql
 type Todo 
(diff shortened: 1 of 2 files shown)


cli-toolchain/graphql.md
```diff
@@ -634,6 +634,51 @@ You can use the *operations* argument to specify which operations are augmented
...
+  @model
+  @auth(rules: [{ allow: owner, operations: [read] }])
+{
+  id: ID!
+  updatedAt: AWSDateTime! 
+  content: String!
+}
+```
+
+In this schema, only the owner of the object has the authorization to perform read (getTodo and listTodos) operations on the owner created object. But this does not prevent any other owner to update/delete some other owner's object.
```
undefobj: What is "any other owner" or "other" in this statement/table? Can you add a "For instance, another owner may have access as they're part of...."

cli-toolchain/graphql.md
```diff
@@ -1178,6 +1224,31 @@ type Employee @model {
...
+type Todo
+  @model
+{
+  id: ID! 
+  updatedAt: AWSDateTime! 
+  content: String! @auth(rules: [{ allow: owner, operations: [update] }])
+}
+```
+
+In this schema, only the owner of the object has the authorization to perform update operations on the `content` field. But this does not prevent any other owner to update some other field in the object owned by another user. If you want to prevent update operations on a field, the user would need to explictly add auth rules to restrict access to that field. One of the ways would be to explictly specify @auth rules on the fields that you would want to protect like the following:
```
undefobj: Same comment as above, what is "any other owner"?

cli-toolchain/graphql.md
```diff
@@ -1178,6 +1224,56 @@ type Employee @model {
...
+```
+
+In this schema, only the owner of the object has the authorization to perform update operations on the `content` field. But this does not prevent any other owner (any user other than the creator or owner of the object) to update some other field in the object owned by another user. If you want to prevent update operations on a field, the user would need to explicitly add auth rules to restrict access to that field. One of the ways would be to explicitly specify @auth rules on the fields that you would want to protect like the following:
+
+```graphql
+type Todo 
+  @model
+{
+  id: ID! 
+  updatedAt: AWSDateTime! @auth(rules: [{ allow: owner, operations: [update] }]) // or @auth(rules: [{ allow: groups, groups: ["Admins"] }])
```
yuth: updatedAt: AWSDateTime! @auth(rules: [{ allow: owner, operations: [update] }]) ~~//~~ # or @auth(rules: [{ allow: groups, groups: ["Admins"] }])

cli-toolchain/graphql.md
```diff
@@ -1178,6 +1224,56 @@ type Employee @model {
 **Note** The `delete` operation, when used in @auth directives on field definitions, translates
 to protecting the update mutation such that the field cannot be set to null unless authorized.
 
+**Note**: When specifying operations as a part of the @auth rule on a field, the operations not included in the operations list are not protected by default. For example, let's say you have the following schema:
```
yuth: **Note**: When specifying operations as a part of the `@auth` rule on a field, the operations not included in the operations list are not protected by default. For example, let's say you have the following schema:

cli-toolchain/graphql.md
```diff
@@ -1178,6 +1224,56 @@ type Employee @model {
...
+type Todo
+  @model
+{
+  id: ID! 
+  updatedAt: AWSDateTime! 
+  content: String! @auth(rules: [{ allow: owner, operations: [update] }])
+}
+```
+
+In this schema, only the owner of the object has the authorization to perform update operations on the `content` field. But this does not prevent any other owner (any user other than the creator or owner of the object) to update some other field in the object owned by another user. If you want to prevent update operations on a field, the user would need to explicitly add auth rules to restrict access to that field. One of the ways would be to explicitly specify @auth rules on the fields that you would want to protect like the following:
```
yuth: In this schema, only the owner of the object has the authorization to perform update operations on the `content` field. But this does not prevent any other owner (any user other than the creator or owner of the object) to update some other field in the object owned by another user. If you want to prevent update operations on a field, the user would need to explicitly add auth rules to restrict access to that field. One of the ways would be to explicitly specify `@auth` rules on the fields that you would want to protect like the following:

cli-toolchain/graphql.md
```diff
@@ -634,6 +634,51 @@ You can use the *operations* argument to specify which operations are augmented
...
+type Todo
+  @model
+  @auth(rules: [{ allow: owner, operations: [read] }])
+{
+  id: ID!
+  updatedAt: AWSDateTime! 
+  content: String!
+}
+```
+
```
yuth: In this schema, only the owner of the object has the authorization to perform read (`getTodo` and `listTodos`) operations on the owner created object. But this does not prevent any other owner (any user other than the creator or owner of the object) to update/delete some other owner's object.

PR #943, opened 2019-10-01
update react tutorial docs minor

ashika01:
*Description of changes:*
- Adding couple of links for clarification around auth.
- Code change to make this tutorial run without any errors.

### DIFF
--- a/js/react.md
+++ b/js/react.md
@@ -1,8 +1,8 @@
 ---
 ---
 
-
 # React & React Native
+
 This tutorial walks you through how to use AWS Amplify to build a React application. You can use a similar process with a React Native application (omitting hosting).
 
 ## Installation
@@ -14,25 +14,26 @@ $ amplify configure
 
 If you're using Windows, we recommend the [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/install-win10).
 
-- Ensure you have [Create React App](https://github.com/facebook/create-react-app) installed. 
+- Ensure you have [Create React App](https://github.com/facebook/create-react-app) installed.
 - Create a new project as follows:<br>
   `npx create-react-app myapp`<br>
   `cd myapp`<br>
 
-***Getting Started with the CLI***
+**_Getting Started with the CLI_**
 To get started, initialize your project in the new directory:
 
 ```
 amplify init
 ```
 
-After you answer the provided questions, you can use `amplify help` at any time to see the overall command structure, and `amplify help <category>` to see actions for a specific category. 
+After you answer the provided questions, you can use `amplify help` at any time to see the overall command structure, and `amplify help <category>` to see actions for a specific category.
 
 The Amplify CLI uses AWS CloudFormation, and you can add or modify configurations locally before you push them for execution in your account. To see the status of the deployment at any time, run `amplify status`.
 
-***Publishing Your Web App***
+**_Publishing Your Web App_**
 
 Without making any changes to your React application, add web hosting as follows:
+
 ```
 amplify add hosting
 ```
@@ -54,11 +55,13 @@ Now that your app is in the cloud, you can add some features like enabling users
 Run `amplify push` to provision your auth resources in the cloud. The `./src/aws-exports.js` file that's created has all of the appropriate cloud resources defined for your application.
 
 Next, add the Amplify library to your web application as follows:
+
 ```
 yarn add aws-amplify aws-amplify-react
 ```
 
 If integrating with a React Native app, use:
+
 ```
 yarn add aws-amplify aws-amplify-react-native
 react-native link amazon-cognito-identity-js # DO NOT run this when using Expo or ExpoKit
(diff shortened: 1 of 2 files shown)


js/react.md
```diff
@@ -101,58 +105,34 @@ The signUpFields array in turn consist of an array of objects, each describing a
...
   defaultCountryCode: '1',
   signUpFields: [
     {
-      label: 'Email',
+      label: 'My custom email label',
       key: 'email',
       required: true,
       displayOrder: 1,
       type: 'string'
     },
```
ashika01: I think we should just show a sample for `signUpFields` here rather than huge sample which would require additional configuration to execute without error.

js/react.md
```diff
@@ -238,26 +220,27 @@ const signUpConfig = {
...
 };
 const usernameAttributes = 'My user name';
 
-export default withAuthenticator(App, { 
-  signUpConfig, 
+export default withAuthenticator(App, {
+  signUpConfig,
   usernameAttributes
 });
 ```
```
ashika01: Using custom attributes here needs additional configuration in aws cognito or at initial setup. For some one following the tutorial, it is misleading that this would work and results in error.

js/react.md
```diff
@@ -238,26 +220,27 @@ const signUpConfig = {
...
 
-Next, we'll add some features, like tracking user behavior analytics and uploading/downloading images in the cloud. Start by running `amplify add analytics` in your project. You can enable analytics for authenticated users only, or for users that aren't authenticated. You would be prompted to ask whether you want to allow guests and unauthenticated users to send analytics events, so you can choose `Yes`. You can also try a new project without authentication configured to test this feature.
+Next, we'll add some features, like tracking user behavior analytics and uploading/downloading images in the cloud.
+
+Before starting, to enable analytics in amplify, we need to associate user to an identity pool. Run `amplify add auth` and configure authentication with using manual configuration. Then, Start by running `amplify add analytics` in your project. You can enable analytics for authenticated users only, or for users that aren't authenticated. You would be prompted to ask whether you want to allow guests and unauthenticated users to send analytics events, so you can choose `Yes`. You can also try a new project without authentication configured to test this feature.
```
ashika01: Mention of a gotcha around auth and identity pool so everyone can compile and run this without fails.

ashika01:
Only commented code include actual changes to docs. Reminder are styling changes. I am opening a new PR with some lint configuration and pre hook for uniformity.

js/react.md
```diff
@@ -85,13 +88,14 @@ You can now use `amplify publish` to build and publish your app again. This time
...
 
 #### SignUp Configuration
-The SignUp component provides your users with the ability to sign up.  It is included as part of the ```Authenticator``` component.
 
-Usage: 
-```<Authenticator signUpConfig={signUpConfig}/>```
+The SignUp component provides your users with the ability to sign up. It is included as part of the `Authenticator` component.
+
+Usage:
+`<Authenticator signUpConfig={signUpConfig}/>`
```
sammartinez: NIT: Can we add a space prior to the closing bracket?

js/react.md
sammartinez: I believe that is fine, part of it was to show all of the different options in regards to what can be passed in. Not sure if we want to add another additional callout of what else is available.

js/react.md
```diff
@@ -201,15 +182,16 @@ class App2 {
 export default withAuthenticator(App2, { usernameAttributes: 'phone_number' });
 ```
 
-Note: if you are using custom signUpFields to customize the `username` field, then you need to make sure either the label of that field is the same value you set in `usernameAttributes` or the key of the field is `username`.
+**Note:** if you are using custom signUpFields to customize the `username` field, then you need to make sure either the label of that field is the same value you set in `usernameAttributes` or the key of the field is `username`.
```
sammartinez: NIT: Can we capitalize the `I` in `if`

js/react.md
```diff
@@ -201,15 +182,16 @@ class App2 {
...
 
 For example:
+
 ```js
 import React, { Component } from 'react';
 import { withAuthenticator } from 'aws-amplify-react';
+//make sure to configure aws amplify
```
sammartinez: I believe this can just be `...` WDYT ?

js/react.md
sammartinez: Got it. Thanks for updating this.

js/react.md
sammartinez: I like this, but what about making it bullet points steps?

js/react.md
```diff
@@ -330,10 +314,15 @@ This is the GraphQL schema that you'll deploy to AWS AppSync. If you're familiar
 After the deployment is complete, open your `App.js` again and update the import to include both the `API` category and `graphqlOperation` method as follows:
 
 ```javascript
-import Amplify, { Analytics, Storage, API, graphqlOperation } from 'aws-amplify';
+import Amplify, {
+  Analytics,
+  Storage,
+  API,
+  graphqlOperation
+} from 'aws-amplify';
```
sammartinez: P R E T T I E R

js/react.md
ashika01: Great call out. thanks

js/react.md
ashika01: Keeping it as split paragraphs for consistency with the rest instead of bullet point.

js/react.md
```diff
@@ -182,14 +183,17 @@ class App2 {
...
 For example:
 
 ```js
 import React, { Component } from 'react';
+import Amplify from 'aws-amplify';
+import awsconfig from './aws-exports';
 import { withAuthenticator } from 'aws-amplify-react';
-//make sure to configure aws amplify
+
+Amplify.configure(awsconfig);
```
sammartinez: Sweet

PR #946, opened 2019-10-02
Add kotlin example for Android push-notifications

yongjhih:
According to Kotlin first, we should consider providing kotlin code examples. That's what I have for sharing. We can go through to other pages later.


jpignata:
Thanks so much for this pull! We'll be adding Kotlin examples to our new docs site, and will do it evenly for all pages at once. Appreciate this and thank you.

PR #948, opened 2019-10-03
iOS Auth - add note for redirect signin/signout URI

lawmicha:
*Issue #, if available:*
It was confusing to go through the CLI flow without knowing what the redirect signin and signout URI is for an app. Adding a note about what the value should be

*Description of changes:*
image

### DIFF
--- a/ios/authentication.md
+++ b/ios/authentication.md
@@ -931,6 +931,12 @@ Do you want to use the default authentication and security configuration?
   I want to learn more.
 ```
 
+When prompted to set up the redirect signin and signout URI, enter `myapp://`. You can also update these values later with `amplify update auth`
+```
+Enter your redirect signin URI: myapp://
+Enter your redirect signout URI: myapp://
+```
+
 After going through the CLI flow, run the following command to deploy the configured resources to the cloud:
 ```terminal
 $ amplify push


lawmicha:
> I'm not sure how this clarifies anything for the developer

As a developer, I had to enter a random URL because I didn't know what to enter and there were no defaults provided in the CLI. (perhaps this is the change we need instead of documentation update). Later, there are instructions to add it to Info.plist here: https://aws-amplify.github.io/docs/ios/authentication#setup-amazon-cognito-hosted-ui-in-ios-app
And then it's clear from the "Manual Setup" above that it should be "myapp://"

Another we had about this was do the URLs have to be unique across different apps? Does this have to be a namespace? Is the guidance to use "myapp://" across all apps?

undefobj:
> > I'm not sure how this clarifies anything for the developer
> 
> As a developer, I had to enter a random URL because I didn't know what to enter and there were no defaults provided in the CLI. (perhaps this is the change we need instead of documentation update). Later, there are instructions to add it to Info.plist here: https://aws-amplify.github.io/docs/ios/authentication#setup-amazon-cognito-hosted-ui-in-ios-app
> And then it's clear from the "Manual Setup" above that it should be "myapp://"
> 
> Another we had about this was do the URLs have to be unique across different apps? Does this have to be a namespace? Is the guidance to use "myapp://" across all apps?

I think the iOS and Android docs for this should actually be rewritten to match what the JS docs do which is simpler, and follows the setup flow which is what is leading to your confusion. Take a look here: https://aws-amplify.github.io/docs/js/authentication#social-providers-and-federation

This is a fairly large refactor of the iOS/Android Auth docs though. Lets discuss who can do this as I don't know if this change alone actually addresses the core concern.

lawmicha:
ah yes, the JS docs is a good place to start with to refactor the iOS and Android Auth docs. Closing this out for now then

Issue #949, opened 2019-10-04
Update docs


bluelinecoding:
https://aws-amplify.github.io/docs/js/start?ref=amplify-rn-btn&platform=react-native

Don't need to link anymore, should be pod install

Ashish5591:
Thanks for bringing it to our attention. 
Will update the docs to add the pod install step for RN >=0.60 and keep react-native link for 0.59 and below.

Ashish5591:
Closing the issue now that the docs have been updated.

PR #950, opened 2019-10-05
Updating React Native getting started docs for 0.60 and above

Ashish5591:
*Related Issues*
#949

*Description of changes:*
Updating React Native getting started docs for 0.60 and above

### DIFF
--- a/js/start.md
+++ b/js/start.md
@@ -225,7 +225,7 @@ $ npm install aws-amplify
 $ npm install aws-amplify-react-native
 ```
 
-If you are using React Native 0.60 and above, autolinking is enabled, but for iOS you would need to run the following step:
+If you are using React Native 0.60 and above, [autolinking](https://github.com/react-native-community/cli/blob/master/docs/autolinking.md) is enabled, but for iOS you would need to run the following step:
 
 ```bash
 $ cd ios && pod install && cd ..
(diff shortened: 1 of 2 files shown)


js/start.md
```diff
@@ -223,9 +223,18 @@ $ react-native init myReactNativeApp
 $ cd myReactNativeApp
 $ npm install aws-amplify
 $ npm install aws-amplify-react-native
-$ react-native link
 ```
 
+If you are using React Native 0.60 and above, autolinking is enabled, but for iOS you would need to run the following step:
```
nikhil-dabhade: 1. Let's add a link for autolinking for readers who do not know what autolinking is : https://github.com/react-native-community/cli/blob/master/docs/autolinking.md

2. Not 100% clear on the instructions here: We are saying that for iOS you need to run the following steps even if they have React Native 0.60 ?

Ashish5591: 1 -> I agree. Added a link to the RN docs.
2 -> Yes thats right, on iOS you need to run the following steps if they have React Native 0.60.

nikhil-dabhade: For no. 2 let's phrase it -->
If you are using React Native 0.60 and above, autolinking is enabled. Run the following  commands for iOS:

Ashish5591: Updated!

PR #960, opened 2019-10-11
[feat] Add pricing links to all used services

mtliendo:
*Description of changes:*

I went ahead and added relevant pricing links to each category so that users can better be informed of costs related to adding services and if they are covered under a free tier


mtliendo:
for fields like API and Predictions, I also added a description of the services.

sammartinez:
@mtliendo Thank you for contributing and taking the time to put this pull request together. Unfortunately, after further discussion internally, we will not be merging this pull request and are going to close this. Since we are an Open Source project, this is the reason why we don't want to link pricing specific to AWS as there are other use cases where AWS pricing does not occur.

PR #972, opened 2019-10-17
Added state change to 'authData' property

kev-ho:
*Issue #, if available:*

I was using the authData property to have a component state who the current user was.

On restarting the session, the user remained authenticated but authData would be null which isn't technically true.

*Description of changes:*
Changed code example to update authData property. 


elorzafe:
@kev-ho there was a large change on this repo, are you interested in solve the conflict and make the change. Otherwise I should close this PR.

Thanks

mauerbac:
No response. Closing

PR #977, opened 2019-10-21
Prepping content for new docs

harrysolovay:


sammartinez:
Can you put this on a separate branch that is not on master? Since it is a feature I feel we should have it not go straight to master

PR #986, opened 2019-10-26
Update authentication.md

kneekey23:
federated sign in with facebook for iOS 13 does not work unless you are on latest version of pods which is 2.12.1

*Issue #, if available:*

*Description of changes:*
updating pod files to reflect latest sdk 2.12.1


kneekey23:
> @kneekey23
> 
> The pods are specified to be "compatible with minor versions up to but not including the next major version". Is there something about 2.12.1 that makes it important for us to specify that version? If so, we need to bump all of the dependencies, not just `authentication.md`. Refer to the [iOS SDK release script](https://github.com/aws-amplify/aws-sdk-ios/blob/master/CircleciScripts/bump_iosdocs_version.py) for an example of what we update during minor version bumps in the release pipeline.

ya fb sdk login not working on iOS 13 that we fixed with this release. Does this not qualify as that important? I remember us discussing me updating the docs with this release, sorry if I missed some places, I will check out that script.

PR #992, opened 2019-10-29
feat: add custom transformer docs

attilah:
*Description of changes:*

Adding a paragraph about custom GraphQL transformer registration.

### DIFF
--- a/cli-toolchain/plugins.md
+++ b/cli-toolchain/plugins.md
@@ -318,7 +318,9 @@ function transform(schema: string): Template {
 ### The Transformer Context
 
 The transformer context serves like an accumulator that is manipulated by transformers. See the code to see what methods are available
-to you: [TransformerContext.ts](https://github.com/aws-amplify/amplify-cli/blob/master/packages/graphql-transformer-core/src/TransformerContext.ts)
+to you.
+
+[https://github.com/aws-amplify/amplify-cli/blob/master/packages/graphql-transformer-core/src/TransformerContext.ts](https://github.com/aws-amplify/amplify-cli/blob/master/packages/graphql-transformer-core/src/TransformerContext.ts)
 
 > For now, the transform only support cloudformation and uses a library called `cloudform` to create cloudformation resources in code. In the future we would like to support alternative deployment mechanisms like terraform.
 
--- a/cli-toolchain/plugins.md
+++ b/cli-toolchain/plugins.md
@@ -13,15 +13,15 @@ The Amplify CLI Core maintains a `plugins.json` file to store the plugin managem
 Each Amplify CLI plugin contains a *amplify-plugin.json*  file to manifest itself as a valid plugin.  <br/>
 The Amplify CLI Core provides a set of utility commands under `amplify plugin` for plugin management, and to facilitate the development of plugins.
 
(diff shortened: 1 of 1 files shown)


hisham:
If it helps, I just released a simple @lowercase graphql transformer to convert strings to lower case before saving to dynamodb. Should serve as a good "hello world" type example of a standalone graphql transformer but does something actually useful:

https://github.com/hisham/graphql-lowercase-transformer
https://www.npmjs.com/package/graphql-lowercase-transformer

DynamoDB does not support case insensitive string search, so this transformer forces searchable strings to be in lowercase in the backend (see https://github.com/aws-amplify/amplify-cli/issues/526).

hisham:
I also had the following feedback when building this transformer. I followed your documentation plus the VersionedModelTransformer as my example but ran into the following issues:  

- Custom transformer libs should have a 'default' exported class, see the imported.default code at https://github.com/aws-amplify/amplify-cli/pull/1396#issue-275898816. If the lib did not have a default export then you get an unhelpful error message which is something like "TypeError: Can't call 'call' on undefined'. Had to spend some time going through amplify code to see where this error was coming from. 
- I could not get my transformer to work if it called any class methods. I would get errors such as 'assertModelDirective' is not a method. All functions had to be property arrow functions. See my code https://github.com/hisham/graphql-lowercase-transformer/blob/master/src/LowercaseTransformer.ts. It's probably some kind of ts/js nuance but I wasn't able to resolve. So I converted all my private methods to private property arrow functions. 

cc'ing @ambientlight since he wrote the custom transformer code.

hisham:
Also graphql transformers to be looked up via npm registry need to be installed globally or else you'll get `Cannot find module` error: 

i.e. `npm install -g lowercase-graphql-transformer`

This wasn't immediately clear to me.

ambientlight:
> I could not get my transformer to work if it called any class methods. I would get errors such as 'assertModelDirective' is not a method. All functions had to be property arrow functions. See my code https://github.com/hisham/graphql-lowercase-transformer/blob/master/src/LowercaseTransformer.ts. It's probably some kind of ts/js/es5 nuance but I wasn't able to resolve. So I converted all my private methods to private property arrow functions

@hisham: same thing for me, will be great to have docs mentioning this.
@attilah: I am wondering, since the deployment target is node, are we really required `ts` to emit es5?

hisham:
@ambientlight yes I am using globally installed amplify. If amplify was installed locally I think it would look it it up locally. Looks like @attilah recently did a PR offering option for both globally installed and locally installed custom transformer packages, see https://github.com/aws-amplify/amplify-cli/pull/2814

attilah:
@hisham @ambientlight Yes, I changed the loading a bit, it has issues about location and initialization as you also faced, I even enabled the pojo scenario, a single object with a method on it and a required properties to work. I've more ideas around it that will come in future changes.

lenarmazitov:
> Also graphql transformers to be looked up via npm registry need to be installed globally or else you'll get `Cannot find module` error:
> 
> i.e. `npm install -g lowercase-graphql-transformer`
> 
> This wasn't immediately clear to me.

I've just installed your npm module globally, but when I execute **amplify api gql-compile** with @lowercase directive I've get error

```
Error: Unknown directive "lowercase".
```

mauerbac:
Outdated. Closing

PR #1022, opened 2019-11-21
Updated docs with new @key info

dabit3:
*Issue #, if available:*

*Description of changes:*


cli-toolchain/graphql.md
```diff
@@ -1874,215 +2166,740 @@ type Comment {
 }
 ```
 
-#### Generates
+#### Generates
+
+In order to keep connection queries fast and efficient, the GraphQL transform manages
+global secondary indexes (GSIs) on the generated tables on your behalf. In the future we
+are investigating using adjacency lists along side GSIs for different use cases that are
```
yuth: I think this is not relevant after the `keyName` change

undefobj: @dabit3 lets remove the "In the future..." statement here. We should just document what we have now.

cli-toolchain/graphql.md
```diff
@@ -1694,41 +1686,64 @@ The `@function` directive generates these resources as necessary:
 
 ### @connection
 
-The `@connection` directive enables you to specify relationships between `@model` object types.
-Currently, this supports one-to-one, one-to-many, and many-to-one relationships. You may implement many-to-many relationships
-yourself using two one-to-many connections and joining `@model` type. See the usage section for details.
+The `@connection` directive enables you to specify relationships between `@model` types. Currently, this supports one-to-one, one-to-many, and many-to-one relationships. You may implement many-to-many relationships yourself using two one-to-many connections and a joining `@model` type. See the usage section for details.
+
```
undefobj: Please remove "yourself"

cli-toolchain/graphql.md
```diff
@@ -1694,41 +1686,64 @@ The `@function` directive generates these resources as necessary:
...
+One can find a full schema example for a set of 17 access patterns at [docs/cli-toolchain/graphql#17-data-access-patterns](#17-data-access-patterns).
```
undefobj: Do we also have this in this PR or is that in the other one?

cli-toolchain/graphql.md
```diff
@@ -1739,11 +1754,288 @@ mutation CreateProject {
 }
 ```
 
-> **Note** The **Project.team** resolver is configured to work with the defined connection.
+> **Note** The **Project.team** resolver is configured to work with the defined connection. This is done with a query on the Team table where `teamID` is passed in as the partition key.
```
undefobj: Do we need to say "partition key"? In the context of GraphQL you just need to know it's an argument.

cli-toolchain/graphql.md
```diff
@@ -1739,11 +1754,288 @@ mutation CreateProject {
...
+
+```graphql
+type Post @model {
+  id: ID!
+  title: String!
+  comments: [Comment] @connection(keyName: "byPost", fields: ["id"])
+}
+
+type Comment @model
+  @key(name: "byPost", fields: ["postID", "content"]) {
```
undefobj: Is this `@key` needed for this use case? Or is it just an another example of useage?

cli-toolchain/graphql.md
```diff
@@ -1739,11 +1754,288 @@ mutation CreateProject {
...
+        }
+      }
+    }
+  }
+}
+```
+
+**Many-To-Many Connections**
+
+You can implement many to many yourself using two 1-M @connections and a joining @model. For example:
```
undefobj: You also need an `@key` for this use case. I know it's called out below but that's after this documentation section.

cli-toolchain/graphql.md
```diff
@@ -1739,11 +1754,288 @@ mutation CreateProject {
...
+                }
+            }
+        }
+    }
+}
+```
+
+#### Alternative Definition
+
+The above definition is the recommended way to create relationships between model types in your API. This involves defining index structures using `@key` and connection resolvers using `@connection`. There is an older parameterization of @connection that creates indices and connection resolvers that is still functional for backwards compatibility reasons. It is recommended to use @key and the new @connection via the fields argument because it gives you more control and can do everything that the old @connection can do.
```
undefobj: I would remove "because it gives you more control and can do everything that the old @connection can do"

cli-toolchain/graphql.md
```diff
@@ -1874,215 +2166,740 @@ type Comment {
...
 
-#### Generates
+#### Generates
+
+In order to keep connection queries fast and efficient, the GraphQL transform manages
+global secondary indexes (GSIs) on the generated tables on your behalf. In the future we
+are investigating using adjacency lists along side GSIs for different use cases that are
+connection heavy.
+
+> **Note** The `@connection` directive manages these GSIs under the hood but there are limitations to be aware of. After you have pushed a `@connection` directive you should not try to change it. If you try to change it, the DynamoDB UpdateTable operation will fail due to one of a set of service limitations around changing GSIs. Should you need to change a `@connection`, you should add a new `@connection` that implements the new access pattern, update your application to use the new `@connection`, and then delete the old `@connection` when it's no longer needed.
```
undefobj: We should reword this. First "under the hood" isn't professional. Second this talks about limits in the CloudFormation of DynamoDB in a negative light, but those controls exist for a reason. We can remove those lines and just speak factually about what customers need to do.

cli-toolchain/graphql.md
```diff
@@ -1874,215 +2166,740 @@ type Comment {
...
+
+| GraphQL Type        | Searchable Operation           |
+|-------------:|:-------------|
+| String      | `ne`, `eq`, `match`, `matchPhrase`, `matchPhrasePrefix`, `multiMatch`, `exists`, `wildcard`, `regexp` |
+| Int     | `ne`, `gt`, `lt`, `gte`, `lte`, `eq`, `range`      |
+| Float | `ne`, `gt`, `lt`, `gte`, `lte`, `eq`, `range`      |
+| Boolean | `eq`, `ne`      |
+
+
+## 17 Data Access Patterns
```
undefobj: Are these supposed to go into a separate md file per the link above?

cli-toolchain/graphql.md
dabit3: Yes, this is in this PR.

cli-toolchain/graphql.md
dabit3: Tested this out, pretty sure for the bi-directional to work we need the @key.

cli-toolchain/graphql.md
dabit3: Not unless you think they should go into a separate file, for now they link to this part of the page.

PR #1048, opened 2019-12-05
Updated pod path for Predictions

royjit:
Since Predictions uses iOS 13, it has to be outside the main AmplifyPlugins pod.

### DIFF
--- a/ios/predictions.md
+++ b/ios/predictions.md
@@ -132,9 +132,9 @@ Add the dependencies to the `Podfile`:
 target :'YOUR-APP-NAME' do
 	use_frameworks!
 
-	pod 'Amplify', '~> 1.0.0'
-	pod 'AWSPluginsCore', '~> 1.0.0'  # pod 'AWSPluginsCore', '~> 1.0.0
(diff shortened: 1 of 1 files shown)


ios/predictions.md
```diff
@@ -131,10 +131,7 @@ Add the dependencies to the `Podfile`:
 ```ruby
 target :'YOUR-APP-NAME' do
 	use_frameworks!
-
-	pod 'Amplify', '~> 1.0.0'
-	pod 'AmplifyPlugins/AWSPluginsCore', '~> 1.0.0'  # pod 'AWSPluginsCore', '~> 1.0.0
-	pod 'AmplifyPlugins/AWSPredictionsPlugin', '~> 1.0.0'
+	pod 'AWSPredictionsPlugin'
```
nikhil-dabhade: Do we not need the "Amplify" and core plugin ?

royjit: We dont, I was able to make it work without that.

lawmicha: just confirmed for AWSAPIPlugin

PR #1050, opened 2019-12-05
Fix typo in "Use NPX" section of the amplify push docs

levibuzolic:
Noticed this code snippet didn't match the heading.


nikhil-dabhade:
@nikhname Can you confirm this is looks good ?

levibuzolic:
No problem, will update. Given this is the first mention of this package
probably worth telling people to install it.

I guess that’s where the npx command would make more sense.

nikhname:
> No problem, will update. Given this is the first mention of this package probably worth telling people to install it. I guess that’s where the npx command would make more sense.

Hey @levibuzolic, so amplify-push is not actually a separate package, just a script that gets added to the project directory after the npx amplify-app is run. So there no need for any other installation.

PR #1051, opened 2019-12-06
Fixes Android Start Doc

TrekSoft:
Per our Chime call @nikhil-dabhade 

### DIFF
--- a/android/start.md
+++ b/android/start.md
@@ -23,6 +23,8 @@ This page guides you through setting up a backend and integration into your Andr
 
 ## Prerequisites
 
+* These steps currently only work on Mac. If you have a Windows machine, follow the steps on one of our categories such as [API here](./api).
+
 * [Install Node](https://nodejs.org/en/)
 
 * [Install Android Studio](https://developer.android.com/studio/index.html#downloads) version 3.1 or higher.
@@ -31,13 +33,18 @@ This page guides you through setting up a backend and integration into your Andr
 
 * This guide assumes that you are familiar with Android development and tools. If you are new to Android development, you can follow [these steps](https://developer.android.com/training/basics/firstapp/creating-project) to create your first Android application using Java.
 
+* If you had previously installed Amplify CLI, update to the latest version by running:
+
+```terminal
+$ npm install -g @aws-amplify/cli
+```
 
 ## Step 1: Configure your app
 You can use an existing Android app or create a new Android app in Java as per the steps in prerequisite section.
 
 a. Open your **project** `build.gradle` and add the following:
 * `mavenCentral()` as a repository
-* `classpath 'com.amplifyframework:amplify-tools-gradle-plugin:0.1.0'` as a dependency
+* `classpath 'com.amplifyframework:amplify-tools-gradle-plugin:0.2.0'` as a dependency
 * A plugin `'com.amplifyframework.amplifytools'` as shown in the example below:
 
 ```gradle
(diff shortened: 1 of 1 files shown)


android/start.md
```diff
@@ -70,9 +77,11 @@ dependencies {
 }
 ```
 
-c. Sync the project and ensure it built successfully after the sync.
+**Note**
+If you get the following error message, "ERROR: Process 'command 'npx'' finished with non-zero exit value 1” this may be due to your user not having permissions to the node_modules folder on your machine. Follow the steps [at this link](https://docs.npmjs.com/resolving-eacces-permissions-errors-when-installing-packages-globally) to resolve it.
```
nikhil-dabhade: 1. Any reason we removed the instruction that says "Sync the project".  Adding the dependency and saving it may not build it for everyone as developers may have auto-sync/build turned off.
2. This error can happen only after the user builds a project using "Make Project", right ? If yes, this note should go in step c.

3. Updated wording: If you get the following error message: "ERROR: Process 'command 'npx'' finished with non-zero exit value 1”, this may be due to the logged in user on your machine having insufficient permissions to access the node_modules folder on your machine.

android/start.md
```diff
@@ -100,19 +109,30 @@ b. To generate the Java classes for these models, click the Gradle Task dropdown
 
 ## Step 3: Add API and Database
 
-a. Run `amplify configure` from the root of your application folder to set up Amplify with your AWS account.
+a. Run `amplify configure` in Terminal from the root of your application folder to set up Amplify with your AWS account.
+
+    - Browser will pop up prompting you to sign in / create new account
```
nikhil-dabhade: Your default browser will open a tab prompting you to sign in / create a new AWS account

android/start.md
```diff
@@ -100,19 +109,30 @@ b. To generate the Java classes for these models, click the Gradle Task dropdown
...
+    - Once done, return to terminal and press Enter
+    - Choose a region
+    - Choose a username (can use default)
+    - Browser will popup again to finish creating that user - you can use all default properties.
```
nikhil-dabhade: Your default browser will open a tab prompting you to walkthrough the process of creating an IAM user. At the end of the process. Save the Access ID and Secret key

android/start.md
```diff
@@ -100,19 +109,30 @@ b. To generate the Java classes for these models, click the Gradle Task dropdown
...
+    - Leave the page open on the finish step and return to Terminal
```
nikhil-dabhade: Return to the terminal.

android/start.md
```diff
@@ -100,19 +109,30 @@ b. To generate the Java classes for these models, click the Gradle Task dropdown
...
+    - Press Enter
+    - It will then ask you to enter the access key ID from the finish page of the browser. Make sure to backspace the default <ACCESS KEY ID> and paste in the key from the browser
```
nikhil-dabhade: Make sure to backspace the default <ACCESS KEY ID> and copy-paste the key for the IAM user you just created.

android/start.md
```diff
@@ -31,13 +33,18 @@ This page guides you through setting up a backend and integration into your Andr
 
 * This guide assumes that you are familiar with Android development and tools. If you are new to Android development, you can follow [these steps](https://developer.android.com/training/basics/firstapp/creating-project) to create your first Android application using Java.
 
+* If you had previously installed Amplify CLI, update to the latest version by running:
```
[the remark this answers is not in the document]
TrekSoft: I thought about this but it's literally any previous version so don't think that's worth adding.

android/start.md
```diff
@@ -23,6 +23,8 @@ This page guides you through setting up a backend and integration into your Andr
 
 ## Prerequisites
 
+* These steps currently only work on Mac. If you have a Windows machine, follow the steps on one of our categories such as [API here](./api).
```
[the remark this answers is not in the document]
TrekSoft: That won't work - that sets up a different schema. None of this page will work for Windows. I think this section should be left as is.

android/start.md
TrekSoft: 1. Yes - Nikhil L said it's not necessary and I've tried it and it's not. Make project by itself is enough.

2. You're right this should now go after the new step C so I'll update that.

3. Ok

PR #1077, opened 2019-12-26
Clarify wording of DataStore guide

jamesonwilliams:


android/datastore.md
```diff
@@ -10,23 +10,22 @@ title: DataStore
...
-
+Amplify DataStore provides a persistent on-device storage repository for you to write, read, and observe changes to your application data. DataStore works both online and offline. When online, DataStore seamlessly synchronizes data with the cloud, and across devices. You model your application data by defining types in the GraphQL schema definition language (SDL). This schema is automatically converted to Java-language **Model** classes that you can code into your business logic. You can use DataStore for offline use cases in a "local only" mode. In local-only mode, you do not need an AWS account, nor do you need to provision any AppSync or DynamoDB resources. DataStore enables lightweight synchronization and flexible conflict resolution strategies when paired with AWS AppSync.
```
undefobj: Any reason why you pulled Delta Sync out? This was here to raise awareness to customers up front of the feature rather than later.

android/datastore.md
```diff
@@ -66,238 +65,245 @@ type Post @model {
...
+{
+  "userAgent": "aws-amplify-cli/2.0",
+  "version": "1.0"
+}
 ```
 
 # Save Data
 
-To write any data to the DataStore you can pass an instance of a Model to `DataStore.save()` and it will be persisted in offline storage. At this point you can use it as an item in a normal data store such as querying, updating or deleting. If you choose to later connect to the cloud the item will be synchronized using GraphQL mutations and any other systems connected to the same backend can run queries or mutations on these items as well as observe them with GraphQL subscriptions.
+To write any data to the DataStore, you can pass an instance of a Model to `DataStore.save()` and it will be persisted in offline storage. The offline storage supports standard storage operations like querying, updating and deleting. If you choose to connect to the cloud at a later time, the local items will be synchronized using GraphQL mutations. Any other systems connected to the same backend can then run queries or mutations on your DataStore items, or observe changes to them via GraphQL subscriptions.
```
undefobj: Why delineate between offline storage? The point of DataStore is so that customers don't need to think about online vs offline, it's just operating on data.

android/datastore.md
```diff
@@ -66,238 +65,245 @@ type Post @model {
...
-{: .callout .callout--info}
+You can subscribe to changes on your Models by using the DataStore's `observe` method. This method reacts dynamically to all changes in local storage. These changes could be the result of local modifications, or the result of changes observed on your GraphQL endpoint, if using remote synchronization.
 
-You can subscribe to changes on your Models by using `observe` in the DataStore API. This reacts dynamically to updates of data to the underlying Storage Engine, which could be the result of GraphQL Subscriptions as well as Queries or Mutations that run against the backing AppSync API if you are synchronizing with the cloud.
+The `AWSDataStorePlugin.observe()` method returns an `io.reactivex.Observable<DataStoreItemChange<? extends Model>>`. You can subscribe to this Rx Observable to get notifications whenever any model is created, updated, or deleted. To use a method in this family, you must make Rx available at runtime by adding `implementation 'io.reactivex.rxjava2:rxandroid:2.1.1'` to your module's `build.gradle`.
```
undefobj: Should we hyperlink to the Rx website?

android/datastore.md
jamesonwilliams: Makes sense! I'll add it back in.

android/datastore.md
jamesonwilliams: Sure, sounds like a good addition -- will add.

android/datastore.md
jamesonwilliams: @undefobj Do you mean in this sentence?
> The offline storage supports standard storage operations like [...]

Agree; here, the distinction may not be helping. I'll make this into just `The DataStore supports standard storage operations like [...]`.

jamesonwilliams:
> Minor comments, looks good.

Thanks @undefobj ! _Finally_ got around to updating this with your feedback ... :-)

PR #1092, opened 2020-01-08
Docs reference an Auth.signUp value that is incorrect

iamtheworstdev:
Auth.signUp returns void, which can be seen in its own [code comment](https://github.com/aws-amplify/amplify-js/blob/54fbdf4b1393567735fb7b5f4144db273f1a5f6a/packages/amazon-cognito-identity-js/src/CognitoUserPool.js#L92)

The last line of code that Auth.signUp performs a non-returning Promise call in client.request, which also returns void per its [code comments](https://github.com/aws-amplify/amplify-js/blob/54fbdf4b1393567735fb7b5f4144db273f1a5f6a/packages/amazon-cognito-identity-js/src/Client.js#L20). 

Appreciate feedback on whether the docs are wrong or with the Amplify team has gotten overzealous with killing promise chains.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/js/authentication.md
+++ b/js/authentication.md
@@ -328,16 +328,6 @@ Auth.resendSignUp(username).then(() => {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


iamtheworstdev:
I believe I am incorrect.

PR #1113, opened 2020-01-28
mutations should be an array, not a string

ChristopheBougere:
*Description of changes:*

I was implementing this trick: *Filtering by related (@connection designated) type* and had this issue when pushing: `Schema Creation Status is FAILED with details: Schema has the following errors:  - The mutations argument on the aws_subscribe directive has an invalid value`.
The reason is mutations should be an array, not a string.


ChristopheBougere:
Any chance to get this merged @kneekey23 ?

PR #1134, opened 2020-02-06
the config variable in './aws-exports' had wrong name

matrec4:
The automatically provided file './aws-exports' declared the config variable to "awsmobile". However, react Native javascript examples in the documentation used either "config" or "awsconfig". Created a "No Credentials" error when GraphQL bridges interacted with backend, as the API ans PubSub were not properly configured.


mauerbac:
thanks for helping out @matrec4 . So sorry we didn't get to this sooner. This should now be address by our use existing resources section.

PR #1140, opened 2020-02-10
Update Amplify iOS Storage

lawmicha:
Remove's progress handler for `Storage.list`, `Storage.remove`, and `Storage.getURL` since those return Void and do not report progress

### DIFF
--- a/ios/storage.md
+++ b/ios/storage.md
@@ -121,16 +121,21 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
         }
         print("AWSMobileClient initialized, userstate: \(userState)")
         
-        let storagePlugin = AWSS3StoragePlugin()
-        do {
-            try Amplify.add(plugin: storagePlugin)
-            try Amplify.configure()
-            print("Amplify configured with storage plugin")
-        } catch {
-            print("Failed to initialize Amplify with \(error)")
-        }
+        configureAmplifyWithStorage()
+        
     }
     return true
+}
+
+func configureAmplifyWithStorage() {
+    let storagePlugin = AWSS3StoragePlugin()
+    do {
+        try Amplify.add(plugin: storagePlugin)
+        try Amplify.configure()
+        print("Amplify configured with storage plugin")
+    } catch {
+        print("Failed to initialize Amplify with \(error)")
+    }
 }
 ```
 
--- a/ios/storage.md
+++ b/ios/storage.md
@@ -120,17 +120,16 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
             return
         }
         print("AWSMobileClient initialized, userstate: \(userState)")
+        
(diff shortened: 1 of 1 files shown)


ios/storage.md
```diff
@@ -101,37 +101,39 @@ open <YOURAPP>.xcworkspace
...
-            print("Amplify configured with storage plugin")
-        } catch {
-            print("Failed to initialize Amplify with \(error)")
+    AWSMobileClient.default().initialize { (userState, error) in
+        guard error == nil else {
+            print("Error initializing AWSMobileClient. Error: \(error!.localizedDescription)")
+            return
         }
+        print("AWSMobileClient initialized, userstate: \(userState)")
+    }
```
wooj2: Seems like we should wait for initialize to properly come up prior to creating the storage plugin etc.  So maybe the code should be something like:
```
    func application(_ application:..........) -> Bool {
        AWSMobileClient.default().initialize { (userState, error) in
                guard error == nil else {
                   .........
                 }
                 self.startAmplify()
        }
    }
    func startAmplify() {
            let storagePlugin = AWSS3StoragePlugin()
             do {
                   try Amplify.add(.......
                ......
             }
    }
```

ios/storage.md
```diff
@@ -245,8 +247,6 @@ func getURL() {
           print("Completed: \(data)")
       case .failed(let storageError):
           print("Failed: \(storageError.errorDescription). \(storageError.recoverySuggestion)")
-      case .inProcess(let progress):
-          print("Progress: \(progress)")
       default:
```
wooj2: Curious why we are removing `.inProcess` ?  Maybe it would be better to get rid of the default case instead?

ios/storage.md
lawmicha: good point, updating the code

ios/storage.md
lawmicha: the reason why we are removing it is because the Type that is returned for `.inProcess` in this case is Void, because getURL does not emit progress events. And currently all operations are built on top of the AmplifyOperation class which provides events callback with these different cases.

https://github.com/aws-amplify/amplify-ios/blob/461c3cfd8923fbbb6eb10b60495b0f4594879cf4/Amplify/Core/Support/AmplifyOperation.swift#L18

I can see a way to remove the default case is if we create another class like `AmplifyRequestOperation` and another `AsyncEvent` that doesn't have inProcess (https://github.com/aws-amplify/amplify-ios/blob/461c3cfd8923fbbb6eb10b60495b0f4594879cf4/Amplify/Core/Support/AsyncEvent.swift#L18)

this is definitely a good callout regarding having to "default" and know which cases to handle. there have been some discussions around removing the .unknown case

lawmicha: adding something over in amplify-ios https://github.com/aws-amplify/amplify-ios/issues/331

PR #1149, opened 2020-02-13
Adds note about GUEST user state on first initialize

TrekSoft:
Adds note to Android documentation that currently was only present in iOS documentation.


sdk/android/authentication.md
```diff
@@ -263,6 +263,8 @@ When complete run `amplify push` and your `awsconfiguration.json` will work auto
 
 If you login in your app either using the "Drop-In Auth" or the `AWSMobileClient` APIs then the user session will transition to an authenticated role.
 
+Note: If initialize is giving you the state as ‘SIGNED_OUT’, try calling `AWSMobileClient.getInstance().getCredentials()` and then calling initialize again.
```
royjit: Why do we need to call initialize again?

jpignata: https://github.com/aws-amplify/docs/commit/434315670f1297b2598cd2c1896861acdca50ab2

Would the fact that initialize returns `SIGNED_OUT` by default help clarify?

TrekSoft: Oof, didn't realize this wasn't merged already - yeah this is just the note iOS already had up which is a stop gap measure to this bug.

PR #1179, opened 2020-03-03
Update default landing section and note for Amplify iOS/Android and SDKs

nikhil-dabhade:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/android/analytics.md
+++ b/android/analytics.md
@@ -2,7 +2,7 @@
 
 **Note**
 This guide shows how to build an app using our Amplify Libraries for Android (Preview) and the Amplify CLI toolchain.
-To use the existing AWS Mobile SDK for iOS instead, [click here.](../sdk/android/analytics)
+To use the existing AWS Mobile SDK for Android instead, [click here.](../sdk/android/analytics)
 {: .callout .callout--warning}
 
 # Analytics
--- a/android/analytics.md
+++ b/android/analytics.md
@@ -1,8 +1,8 @@
 <br />
 
 **Note**
-Amplify Android is in preview mode and not intended for production usage at this time. We welcome feedback to improve your experience in using Amplify Android.
-[Click here](../sdk/android/analytics) to access the Analytics category guide for Android SDK 2.0 docs.
+This guide shows how to build an app using our Amplify Libraries for Android (Preview) and the Amplify CLI toolchain.
+To use the existing AWS Mobile SDK for iOS instead, [click here.](../sdk/android/analytics)
 {: .callout .callout--warning}
 
 # Analytics
--- a/android/authentication.md
+++ b/android/authentication.md
@@ -7,6 +7,13 @@ title: Authentication
 {% assign media_base = base_dir | append: page.dir | append: "media" %}
 {% assign common_media = base_dir | append: "/images" %}
 
+<br />
+
+**Note**
+This guide shows how to add Auth using the existing AWS Mobile SDK for Android and the Amplify CLI toolchain.
+Currently, this is our supported way of using Auth with Amplify Libraries for Android (preview).
+{: .callout .callout--warning}
+
 # Authentication
 
 The `AWSMobileClient` provides client APIs and building blocks for developers who want to create user authentication experiences. This includes declarative methods for performing authentication actions, a simple "drop-in auth" UI for performing common tasks, automatic token and credentials management, and state tracking with notifications for performing workflows in your application when users have authenticated.
--- a/android/datastore.md
+++ b/android/datastore.md
@@ -10,7 +10,7 @@ title: DataStore
 <br />
 
 **Note**
-Amplify Android is in preview mode and is not intended for production usage at this time. We welcome feedback to improve your experience in using Amplify Android.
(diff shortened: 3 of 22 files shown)


android/analytics.md
```diff
@@ -1,8 +1,8 @@
 <br />
 
 **Note**
-Amplify Android is in preview mode and not intended for production usage at this time. We welcome feedback to improve your experience in using Amplify Android.
-[Click here](../sdk/android/analytics) to access the Analytics category guide for Android SDK 2.0 docs.
+This guide shows how to build an app using our Amplify Libraries for Android (Preview) and the Amplify CLI toolchain.
+To use the existing AWS Mobile SDK for iOS instead, [click here.](../sdk/android/analytics)
```
TrekSoft: Should be for Android not iOS

kneekey23: do we want to explain the difference here between the SDK and Amplify? I think that's still confusing for customers.

android/datastore.md
```diff
@@ -10,7 +10,7 @@ title: DataStore
 <br />
 
 **Note**
-Amplify Android is in preview mode and is not intended for production usage at this time. We welcome feedback to improve your experience in using Amplify Android.
+This guide shows how to build an app using our Amplify Libraries for Android (Preview) and the Amplify CLI toolchain.
```
TrekSoft: What was the reasoning behind this change (and the others like it) since it seems like it significantly weakens the warning about the nature of the current state of the library?

android/authentication.md
```diff
@@ -7,6 +7,13 @@ title: Authentication
 {% assign media_base = base_dir | append: page.dir | append: "media" %}
 {% assign common_media = base_dir | append: "/images" %}
 
+<br />
+
+**Note**
+This guide shows how to add Auth using the existing AWS Mobile SDK for Android and the Amplify CLI toolchain.
+Currently, this is our supported way of using Auth with Amplify Libraries for Android (preview).
```
kneekey23: -> "Currently, this is our supported way of using Auth with Amplify Libraries for Android until we release the Auth category for Amplify for Android"

android/datastore.md
kneekey23: why not add in this line like the others? 
`To use the existing AWS Mobile SDK for Android instead, [click here.](../sdk/android/start)`

ios/authentication.md
```diff
@@ -8,6 +8,14 @@ title: Authentication
 {% assign image_base = base_dir | append: page.dir | append: "images" %}
 {% assign common_media = base_dir | append: "/images" %}
 
+<br />
+
+**Note**
+This guide shows how to add Auth using the existing AWS Mobile SDK for iOS and the Amplify CLI toolchain.
+Currently, this is our supported way of using Auth with Amplify Libraries for iOS (preview).
```
kneekey23: same comment as above. "until we add Auth category to Amplify"

sdk/android/analytics.md
```diff
@@ -6,6 +6,13 @@ title: Analytics
 {% endif %}
 {% assign media_base = base_dir | append: page.dir | append: "media" %}
 
+<br/>
+
+**Note**
+This guide shows how to build an app using AWS Mobile SDK for Android and the Amplify CLI toolchain.
+To use our new, preview developer experience with new Amplify Libraries for Android, [click here.](../../android/analytics)
```
kneekey23: maybe another place where we can explain the difference to our customers? why do I want to use Amplify over the SDK

sdk/android/api.md
```diff
@@ -1,11 +1,19 @@
...
 {% if jekyll.environment == 'production' %}
   {% assign base_dir = site.amplify.docs_baseurl %}
 {% endif %}
 {% assign media_base = base_dir | append: page.dir | append: "media" %}
 
+<br />
+
+**Note**
+This guide shows how to build an app using AWS Mobile SDK for Android and the Amplify CLI toolchain.
+To use our new, preview developer experience with new Amplify Libraries for Android, [click here.](../../android/api)
```
kneekey23: same comment as above

kneekey23: and same for all below

android/datastore.md
nikhil-dabhade: The current state of library is preview. We are reflecting that state in the old and new message as well. The reasoning for change is first we should be directing customer to docs for products that are GA and not preview by default. This is a widely adopted AWS practice to direct to GA rather than preview. We did the other way around to get more feedback when we launched. Second, existing users should not get confused on state of SDKs as we work on Amplify which seemed to be the case earlier.  Hope that clarifies it more.

android/analytics.md
nikhil-dabhade: Thanks for catching. Will change it.

nikhil-dabhade: I think it's a fair expectation. Since Amplify is in preview it would not be an apples to apples comparison. In addition, the note may not be enough space to explain the difference. However, we can add the difference to Getting started pages. Thoughts?

android/datastore.md
nikhil-dabhade: DataStore is not present for SDKs. There is no equivalent SDK for it. Thus, did not add a note for SDK here.

Issue #1127, opened 2020-01-31
Amplify OAuth documentation not updated with SignInWithApple


t-moedano:
**Describe the bug**
If you look at Amplify authorization documentation and tutorials there is any information about SignInWithApple but the feature is implemented

**To Reproduce**
1. Go to https://aws-amplify.github.io/docs/js/authentication#oauth-and-federation-overview
2. See there is not instruction about SignInWithApple 
3. Go to https://aws-amplify.github.io/amplify-js/api/enums/cognitohosteduiidentityprovider.html
4. See that does not show that one provider options is 'SignInWithApple' although this option is a possible parameter in federatedSignIn method

**Expected behavior**
See instructions/information about the use of SignInWithApple option in AWS Amplify.

**Additional context**
SignInWithApple it is a important feature to be explained. Apple requires that new social logins contain Sign In With Apple option in order to approve new apps. It is good that AWS Amplify already support this feature but I didn't see any place AWS Amplify documentation talking about it.

kneekey23:
hi @t-moedano have you tried following these instructions here? https://aws.amazon.com/blogs/security/how-to-set-up-sign-in-with-apple-for-amazon-cognito/

t-moedano:
Hello @kneekey23 ! Yes, I also followed this guide when building my application. But what Im pointing is that the Amplify documentation should be updated at least pointing to other Apple related AWS guide. (As does with facebook and google)

Just looking at Amplify documentation it doesn’t seem that Amplify supports SignInWithApple but it really does. So, in the same way Amplify pages explain and give examples with Facebook and Google, should also include Sign In With Apple.

And also, the API reference itself seems to be outdated because it doesnt include SignInWithApple parameter as a possible option for federatedSignIn method.

kneekey23:
Closing re: above mentioned PR

PR #1220, opened 2020-03-10
Reference Apple Sign In blog and hostedUI options

lawmicha:
*Description of changes:*

- Add a reference to the Cognito blog for setting up hosted UI for Apple SIgn In.
- Make sure developers know what to use in `HostedUIOptions` by passing in "SignInWithApple" to launch AppleSignIn directly.

*Releated*
- currently, CLI does not support Apple auth flow, details here: https://github.com/aws-amplify/amplify-cli/issues/3434
- developer requesting documentation https://github.com/aws-amplify/docs/issues/1127


ios/authentication.md
```diff
@@ -1014,6 +1014,14 @@ Note: your user pool domain is something like: `domain_prefix-<env-name>.auth.<r
 ![Image]({{common_media}}/cognitoHostedUI/amazon5.png)
 5. Choose Save.
 
+##### Setting up Hosted UI Domain with Apple
```
jpignata: By "with Apple" do you mean "Sign in with Apple?" If so, is it worth it to be explicit?

ios/authentication.md
```diff
@@ -1014,6 +1014,14 @@ Note: your user pool domain is something like: `domain_prefix-<env-name>.auth.<r
...
+
+1. The easiest way to get HostedUI configuration in your app is through automated set up with CLI. Choose a supported identity provider in your set up (Google and/or Facebook, or Login with Amazon) you would like to use. 
+
+2. Once you have the configuration for your hosted UI in your application, learn more about [How to set up Sign in with Apple for Amazon Cognito](https://aws.amazon.com/blogs/security/how-to-set-up-sign-in-with-apple-for-amazon-cognito/).
```
jpignata: Does this step elide a bunch of other steps that the user needs to pick from the blog post, a la "draw the rest of the owl?" If so, any value in hoisting those steps here?

lawmicha: so some context here is that Amplify CLI does not support Apple sign in. Cognito does and our SDK does which is why we should have some steps in the docs.

missing steps i can think of that i can add is `amplify console auth` to open your UserPool in AWS Console. there might be a disconnect from the CLI to getting to their cognito user pool, which is why they need to run that command or navigate there (and then they need to know which userpool was provisioned) manually.

so i'm not against hoisting the steps here, it would increase the page size from the already really large page, so maybe we just need a separate page and link it from here. the other question is how long before CLI priotizies apple sign in? asking over in https://github.com/aws-amplify/amplify-cli/issues/3434 as that may influence our decision here. it looks like the blog was updated before, and so hoisting means we will also take ownership of cognito steps and updates, which is fair to do, as we already guide customers to amplify docs over cognito docs when using our SDK.

lawmicha:
@brene regarding frontporting to v2-preview, https://github.com/aws-amplify/docs/pull/1223#pullrequestreview-372465238 wanted to clarify, are we doing this for all PR's to master?

ios/authentication.md
lawmicha: previous sections say "with Amazon", "with Google", however i have heard apple's referenced as "Apple Sign In". so i'm not sure
"Setting up Hosted UI Domain for Sign In with Apple"
"Setting up Hosted UI Domain for Apple Sign In"

ios/authentication.md
jpignata: Thanks for the context. I read the page and understand more fully now what this would entail. Assuming it's on the near term roadmap, my vote would be to wait until the CLI supports it directly and write the (much reduced) set of steps here.

ios/authentication.md
royjit: "Setting up Hosted UI Domain for Sign In with Apple"

brene:
> @brene regarding frontporting to v2-preview, https://github.com/aws-amplify/docs/pull/1223#pullrequestreview-372465238 wanted to clarify, are we doing this for all PR's to master, ie. this Apple docs PR?

Yes. In the interim we need to bifurcate all documentation.

lawmicha:
> > @brene regarding frontporting to v2-preview, [#1223 (review)](https://github.com/aws-amplify/docs/pull/1223#pullrequestreview-372465238) wanted to clarify, are we doing this for all PR's to master, ie. this Apple docs PR?
> 
> Yes. In the interim we need to bifurcate all documentation.

okay, sounds good! here it is: https://github.com/aws-amplify/docs/pull/1251

PR #1221, opened 2020-03-10
I think it should say below steps? Maybe I'm wrong

waldothedeveloper:
If you have used Amplify CLI to create your API, you can enable custom headers by following ABOVE steps ???????
changed to: 
If you have used Amplify CLI to create your API, you can enable custom headers by following BELOW steps:

*Issue #, if available:*

*Description of changes:*


waldothedeveloper:
> Reading the page, I agree that this is incorrect. Thanks!
> 
> Your change to "below steps" also doesn't read very well. How about "these steps" or "the below steps" or something similar?

I would go with "these steps" in my opinion.

PR #1226, opened 2020-03-11
iOS - Remove AWSCore from GettingStarted, update AppSync 3.1.0

lawmicha:
Since moving AppSync pod install up to the top, we no longer need to have AWSCore explicit as AppSync will pull in the right version.

If our guidance in the getting started guide is to test out the API functionality using AppSync then we can just let it resolve to the required AWSCore dependency.


mlabieniec:
Will need to update other areas, can you add them to this pr?

PR #1229, opened 2020-03-11
feat: search

harrysolovay:


client/src/docs-ui/secondary-nav/secondary-nav.tsx
```diff
@@ -14,6 +14,18 @@ export class DocsSecondaryNav {
   /*** the current filter state */
   @Prop() readonly selectedFilters: SelectedFilters;
 
+  componentDidRender() {
+    if (Build.isBrowser) {
+      // @ts-ignore
+      docsearch({
+        apiKey: "25626fae796133dc1e734c6bcaaeac3c",
```
nikhil-dabhade: Does the API key need to be hardcoded ? Seems like this could be a security issue ?

harrysolovay: This API key isn't our own. It's the Algolia test API. But no, it doesn't matter, even if it was ours.

ericclemmons: For reference, the current docs also have this:
https://github.com/aws-amplify/docs/blob/98c85368c7f19ae4b9808ca45dffea9ebaf41dad/theme/_includes/site-footer.html#L53-L62

nikhil-dabhade: Verified with Harry that this is a test API key which is public, given by Algolia and cannot be used to impersonate us.

PR #1230, opened 2020-03-11
More mobile tweaks

harrysolovay:
* fix border radius on webkit devices
• video width (iframes)


jamesonwilliams:
In the future, we should be using better documentation on PRs. "More mobile tweaks" doesn't help a lot.

PR #1236, opened 2020-03-16
LoginWithAmazon refreshes user token

ericclemmons:
*Issue #, if available:* Closes https://github.com/aws-amplify/amplify-js/issues/5082

*Description of changes:*

Starting Friday, I tested `LoginWithAmazon` with my `photo-albums-e2e` sample app: continuously checking the values of `currentAuthenticatedUser()` and `currentUserPoolUser()` after a few hours, throughout the weekend, and again on Monday.

My tokens continued to be refreshed with access to the app (CRUD) functioning as expected.  `user` objects were returned as well.

**Preview Link**: https://pr-1236.d3jqoae19jbaz1.amplifyapp.com/js/authentication#oauth-and-hosted-ui

---

### DIFF
--- a/js/authentication.md
+++ b/js/authentication.md
@@ -877,7 +877,7 @@ OAuth support in Amplify uses Cognito User Pools and supports federation with so
 
 #### OAuth and Hosted UI
 
-After configuring the OAuth endpoints, you can use them or the Hosted UI with `Auth.federatedSignIn()`. Passing *LoginWithAmazon*, *Facebook*, or *Google* will bypass the Hosted UI and federate immediately with the social provider as shown in the below React example. If you are looking to add a custom state, you are able to do so by passing a `string`
+After configuring the OAuth endpoints, you can use them or the Hosted UI with `Auth.federatedSignIn()`. Passing `LoginWithAmazon`, `Facebook`, or `Google` will bypass the Hosted UI and federate immediately with the social provider as shown in the below React example. If you are looking to add a custom state, you are able to do so by passing a `string`
 (e.g. `Auth.federatedSignIn({ customState: 'xyz' })`) value and listening for the custom state via Hub
 
 ```javascript
@@ -940,7 +940,7 @@ const { token } = federatedInfo;
 
 **Refreshing JWT Tokens**
 
-By default, AWS Amplify will automatically refresh the tokens for Google and Facebook when the app is in the web environment, so that your AWS credentials will be valid at all times. But if you are using another federated provider, or the app is running in React Native, you will need to provide your own token refresh method:
+By default, AWS Amplify will automatically refresh the tokens for `Google`, `Facebook`, and `LoginWithAmazon` when the app is in the web environment, so that your AWS credentials will be valid at all times. But if you are using another federated provider, or the app is running in React Native, you will need to provide your own token refresh method:
 ```javascript
 import { Auth } from 'aws-amplify';
 


Amplifiyer:
Do we know how this is actually working? Looking at the code, amplify [only supplies refresh handlers](https://github.com/aws-amplify/amplify-js/blob/a047ce73abe98c3bf82e888c3afb4d2f911805f3/packages/core/src/Credentials.ts#L25-L26) for google and facebook and later while refreshing the tokens, should [ideally fail ](https://github.com/aws-amplify/amplify-js/blob/a047ce73abe98c3bf82e888c3afb4d2f911805f3/packages/core/src/Credentials.ts#L122-L125)when the federated provider is LoginWithAmazon.

I might be looking at the wrong tokens or at wrong place, so just interested to see where the refresh really happens.

ericclemmons:
@Amplifiyer That's exactly what I was hoping to find out.  In practice, it seems to work (my user sessions are lasting longer than 60 minutes), despite not having handlers.

Without traipsing through code paths, what's the best way to validate that tokens are actually being refreshed?  (I was looking at it from a functional standpoint, which passed somehow...)

ericclemmons:
Adding onto my prior post, I was looking to reproduce the _failure_ state (where `LoginWithAmazon` **does not** refresh tokens).

I'm operating off the current premise that tokens last 60 minutes / 1 hour.

Amplifiyer:
I am under the same impression that we force refresh tokens after 50 mins now (since elorzafe's changes and typically [last 60 mins](https://developer.amazon.com/docs/login-with-amazon/access-token.html)) and so Credentials.refreshFederatedToken() should be getting called. I think you maybe able to see the expiration of the JWT token in your local cache. You can also possibly enable debug logs and check the debug logs of `Credentials.refreshFederatedToken`

I've a hunch that this [script](https://github.com/aws-amplify/amplify-js/blob/4bd5c7ebef0ad223d4c05a452e696242927a750f/packages/aws-amplify-react/src/Auth/Provider/withAmazon.tsx#L131) might be taking care of the refreshes, but I can't find any [documentation](https://developer.amazon.com/docs/login-with-amazon/javascript-sdk-reference.html) to back it up either

ericclemmons:
You're right about the 50/60m change.  Interestingly, I'm not using those buttons, so that [script](https://github.com/aws-amplify/amplify-js/blob/4bd5c7ebef0ad223d4c05a452e696242927a750f/packages/aws-amplify-react/src/Auth/Provider/withAmazon.tsx#L131) isn't part of the equation 🤔

Going to escalate on this question.  We don't want to misrepresent facts, but it's crucial that I understand whether "working" for me is validating that refresh tokens are working or not.

Amplifiyer:
I believe it is a very strong data point in validating that refreshing the access token is working (is it same as saying refresh tokens are working?). It makes me slightly nervous that we are putting this in public docs without knowing `why` they are working 🧐

I'm good for this PR if you believe we have strong evidence that it should work for majority of our customers.

ericclemmons:
@Amplifiyer Thanks for your valuable feedback.  I'm in complete agreement: our docs should not make any promises that we can't backup with an implementation.

I'm going to get to the bottom of this...

ericclemmons:
Going to close this & revisit in https://github.com/aws-amplify/amplify-js/issues/5082.

This portion of the docs are specific to IDP:
> https://aws-amplify.github.io/docs/js/authentication#token-refresh

PR #1306, opened 2020-03-22
Update setup.md to be more explicit. 

omenking:
* I can understand why we have a package.json file with very specific versions though we can encourage users to install these packages themselves so they are using the latest versions to avoid security vulnerabilities but also still provide an example so if they are having issues they can just copy ours with the fixed versions.
* We should set explicitly the host and port for devServer. Users are generally likely going to need to change these settings for their environment and this will save them time trying to hunt down these options.
* For cloud IDEs such a Cloud9 you need to use port `0.0.0.0`. This will save users significant time debugging connectivity issues.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/start/getting-started/fragments/vanillajs/setup.md
+++ b/docs/start/getting-started/fragments/vanillajs/setup.md
@@ -18,7 +18,37 @@ The app directory structure should be:
         |- app.js
 ```
 
-Add the following to the `package.json` file:
+Intitalize a new `package.json` file using the following command:
+```
+npm init -y
+```
+
+We'll need to install dev dependencies 
+
+```
+npm i \
+  webpack 
+  webpack-cli \
+  copy-webpack-plugin \
+  webpack-dev-server \
+  --save-dev
+```
+We'll need to install amplify dependencies
+
+```
+npm install @aws-amplify/api dependencies --save
+```
+
+We'll need to add the following scripts to our package.json.
+
+```
+  "scripts": {
+    "start": "webpack && webpack-dev-server --mode development",
+    "build": "webpack"
+  }
+```
+
+Your final `package.json` should something similar to this:
 
 ```javascript
 {
@@ -41,6 +71,7 @@ Add the following to the `package.json` file:
   }
 }
 ```
+
 
 ### Install local development dependencies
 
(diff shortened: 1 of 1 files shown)


nikhil-dabhade:
@brene @sammartinez This is in the Js getting started guide. Can you take a look and validate the changes?

sammartinez:
> @brene @sammartinez This is in the Js getting started guide. Can you take a look and validate the changes?

Verifying now

docs/start/getting-started/fragments/vanillajs/setup.md
```diff
@@ -123,6 +156,8 @@ module.exports = {
 };
 ```
 
+By default `webpack-dev-server` will start on port `8080`. You may need to change devServer host to `0.0.0.0` if you are running your web-server in a Cloud IDE such as AWS Cloud9.
```
brene: I think this is too much detail for a getting started. My thought: If people are aware of Cloud IDEs they probably will know how to setup webpack. @sammartinez - thoughts?

brene: Chatted with Sam offline. This seems fine. I'll resolve this comment.

sammartinez: This should be fine, but we may want to recommend locking down to your IP so this isnt open to the world

docs/start/getting-started/fragments/vanillajs/setup.md
```diff
@@ -18,7 +18,37 @@ The app directory structure should be:
...
-Add the following to the `package.json` file:
+Intitalize a new `package.json` file using the following command:
+```
+npm init -y
+```
+
+We'll need to install dev dependencies 
+
+```
+npm i \
```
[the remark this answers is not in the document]
sammartinez: @omenking can you verify this works in Windows?

brene:
@omenking - let me know if you have any questions regarding my comments.

brene:
@omenking - just checking in if you can address our comments

docs/start/getting-started/fragments/vanillajs/setup.md
brene, line 159: Can you add a ">" around this. e.g.
> By default `webpack-dev-server` will start on port `8080`. You may need to change devServer host to `0.0.0.0` if you are running your web-server in a Cloud IDE such as AWS Cloud9.

PR #1319, opened 2020-03-24
fix: how links are parsed and transformed in capi

harrysolovay:
* hash string support
* stability improvements


client/src/docs-ui/chat-button/chat-button.style.ts
```diff
@@ -25,6 +25,6 @@ export const gitterChatStyle = css`
   }
 
   a {
-    color: var(--font-color);
+    color: var(--color-white);
```
jordanranz: Let's make this a font variable like `--font-color-contrast` instead of using `--color-white`

capi/src/visitors/image-references.ts
```diff
@@ -1,6 +1,6 @@
 import * as t from "../types";
 
-const IS_URL_REGEX = /^https?:\/\//i;
+const IS_ABSOLUTE_REGEX = /^https?:\/\//i;
```
jordanranz: Can we name this `IS_URL_ABSOLUTE_REGEX`

capi/src/visitors/links.ts
```diff
@@ -1,90 +1,117 @@
 import * as t from "../types";
 import * as path from "path";
+import {parse} from "url";
 
-const IS_URL_REGEX = /^https?:\/\//i;
+const IS_ABSOLUTE_REGEX = /^https?:\/\//i;
```
jordanranz: same comment as above

capi/src/visitors/image-references.ts
harrysolovay: Done

capi/src/visitors/links.ts
harrysolovay: Done

PR #1338, opened 2020-03-26
New UI, Old docs

jordanranz:
*Description of changes:*
- Amplify UI Components documentation


_config.yml
```diff
@@ -55,806 +55,810 @@ plugins:
...
+      - title: Amplify iOS
+        excerpt: Version of the docs
+        cta: Read more
+        class: aws-icon-ps-60-circular-circuit
+        url: "/ios/start"
+      - title: iOS SDK 2.0
+        excerpt: Version of the docs
+        cta: Read more
+        class: aws-icon-ps-60-circular-circuit
+        url: "/sdk/ios/start"
```
sammartinez, line 619: Seeing that most of these changes look like prettier doing something, do we want to do the same config change we did for JS last night?

_config.yml
```diff
@@ -55,806 +55,810 @@ plugins:
...
+        item_external_url: "https://aws-amplify.github.io/media/ui_library"
+        id: ui
+    ios:
+      theme: "dark"
+  footer:
+    content:
+      logo:
+        text: AWS Amplify is supported by Amazon Web Services
+        image: true
+      copyright: © 2018-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
jordanranz: sure

_config.yml
jordanranz: There is no prettier in docs v1 so I don't believe so. Not sure why my VS code formatted everything..

jordanranz: I think this is valid formatting though

sammartinez: Agreed it was more of a callout if it was something with Prettier, which it isn't, and also just seeing what it was.

sammartinez: This was non-blocking btw

PR #1358, opened 2020-03-27
Add redirects for V1 Android library pages

jpignata:
*Issue #, if available:* n/a

*Description of changes:* We've been committing all of our updates to V2 and we do not want multiple versions of our docs out there as it is confusing to customers.

This change adds a redirect layout which:

- Indicates to Google and other search engines via a link[rel=canonical] tag where the new content can be found
- Redirects the user via a meta[http-equiv=refresh] tag and has a fallback script tag to change the window's location.

The following pages will be redirected to V2:

- Android Amplify Getting Started
- Android Amplify API
- Android Amplify Authentication
- Android Amplify DataStore
- Android Amplify Storage
- Android Amplify Analytics


jordanranz:
This looks good. I am guessing that all of the removed content is represented in the new docs. Also, is the redirect.html used for the canonical bit?

android/authentication.md
```diff
@@ -1,1769 +1,3 @@
 ---
-title: Authentication
+redirect_to: https://docs.amplify.aws/lib/auth/getting-started?platform=android
```
sammartinez: How would you redirect from here?

jordanranz: Pretty sure https://aws-amplify.github.io/docs/sdk/android/authentication -> https://docs.amplify.aws/lib/auth/getting-started?platform=android right?

PR #1364, opened 2020-03-28
fix: nav back bug

harrysolovay:
So that navigating back works in all situations. Without extra lifecycle stages triggered. And make all hoverable parts of TOC items clickable.


client/src/docs-ui/page/page.tsx
```diff
@@ -90,43 +88,52 @@ export class DocsPage {
...
-            }),
-          );
-          this.computeFilter();
-        }
-      } catch (exception) {
-        track({
-          type: AnalyticsEventType.PAGE_DATA_FETCH_EXCEPTION,
-          attributes: {url: location.href, exception},
-        });
+  getPage = async () => {
```
jordanranz: Can we change the name of this member function? maybe `getPageData`

PR #1368, opened 2020-03-28
Visual tweaks to docs-version-switch

jpignata:
This is just a spike to illustrate a few things I found suboptimal about the current docs-version-switch design to encourage conversation. This isn't presented as a solution to be merged, hence I am opening it as a draft.

*Issue #, if available:* n/a

*Description of changes:* To my eye, the style of the docs-version-switch has a few visual issues that impair its usability:

1. The light grey used as the active state does not immediately make it obvious that area is a toggle; the hover state is an imperceptibly lighter shade of grey which seems like an error.
1. The options presented are "Libraries" which is curious as it is under the global navigation of "Libaries," with a subtitle of "preview", and "SDK", which is presented as stable. It's not immediately clear to me what those options indicate: what we're aiming to communicate is preview vs. current versions rather than stability.
1. The SDK is presented as stable which indicates it is preferred, and the "Libraries" navigation item defaults to it, but it is presented as the second option in the switch.

To address the above, this change:

- Replaces the word "Libraries" with "Amplify"
- Replaces the word "stable" with "current"
- Swaps the two options presenting SDK first
- Uses a light orange background for the active state and changes the font weight
- Removes the hover state from the switch


jpignata:
@brene made a few tweaks, which I think is a great improvement:

swaminator:
What i'm seeing in the [preview](https://pr-1368.d39agkzf0cwtdz.amplifyapp.com/lib?platform=android) does not resemble what I see in the image above. The text is showing up as blue instead of white which looks weird.

jpignata:
Thanks @swaminator. Ignore the specifics of the preview -- I made these changes in a 15 minute timebox as a strawman proposal as opposed to sometime I am proposing we merge directly. I'll work with the team next week to make a better proposal and circle back for sign off.

swaminator:
> To address the above, this change:
> 
> * Replaces the word "Libraries" with "Amplify"
> * Replaces the word "stable" with "current"
@jpignata 
I agree with your point on confusion when clicking Libraries to see Libraries and SDK. However I feel strongly that we should not use the word "Amplify" as it is overloaded. Was looking at what other tools do on NPM. How about we have the following in the switcher for iOS and Android: 
* Latest (points to the SDK docs)
* Preview (points to the Amplify Library docs)

That addresses both your concerns.

jpignata:
@swaminator says:

> How about we have the following in the switcher for iOS and Android:
> 
> * Latest (points to the SDK docs)
> * Preview (points to the Amplify Library docs)

I think that's reasonable and addresses my concerns, but would like some others' input. We do need to get our nomenclature straight amongst ourselves as I'm dissatisfied to how we elucidate the differences between the two libraries or SDKs throughout the documentations. We need a canonical noun to refer to "old but current" and another to refer to "next." 

cc: @mlabieniec @sammartinez

mlabieniec:
> > To address the above, this change:
> 
> > 
> 
> > * Replaces the word "Libraries" with "Amplify"
> 
> > * Replaces the word "stable" with "current"
> 
> 
> 
> @jpignata 
> 
> I agree with your point on confusion when clicking Libraries to see Libraries and SDK. However I feel strongly that we should not use the word "Amplify" as it is overloaded. Was looking at what other tools do on NPM. How about we have the following in the switcher for iOS and Android: 
> 
> * Latest (points to the SDK docs)
> 
> * Preview (points to the Amplify Library docs)
> 
> 
> 
> That addresses both your concerns.
> 
> 
> 
> 

I don’t agree with Amplify being overloaded. Right now it’s super confusing. How about Framework/SDK? I don’t like Latest / Preview so much since it hints at both are amplify which is definitely not the case and additionally confusion.

swaminator:
> I don’t agree with Amplify being overloaded. Right now it’s super confusing. How about Framework/SDK? I don’t like Latest / Preview so much since it hints at both are amplify which is definitely not the case and additionally confusion.

@mlabieniec I'm on customer calls where they are using different tools within Amplify (Console, CLI, Libraries) and they've had feature requests where we've had to spend time figuring out which product it actually belonged to. I see your concern on confusion wrt `Latest/Preview` for existing customers. 

Why don't we slightly tweak what is currently in production: SDK (current)/Libraries (preview). So when an Android/iOS user clicks on Libraries, they land on the SDK (current), they see info on what the SDK is vs the Libraries in the landing page, and can switch to the Libraries. Repeating information isn't a bad thing.

jordanranz:
I really like the look and feel of the switcher in comparison with the grey color. I'll let you @jpignata and @mlabieniec figure out the verbage.

jpignata:
Closing in favor of https://github.com/aws-amplify/docs/pull/1392 and we'll revisit the verbiage.

PR #1374, opened 2020-03-29
Fix broken Connect component example

Weetbix:
*Description of changes:*
Currently this example usage of Connect will always show the "error" state, as the connect callback is called with `errors: []` when successfully ran. 

This means people using this example code will see "error" in their app even though the request was successful and data has been retrieved. 


Weetbix:
Bump to stop the auto close...  PR still valid

PR #1418, opened 2020-04-02
fix: hash link bug

harrysolovay:
Task labeled "#Links that just won't work!!! 😡"


client/src/utils/get-nav-height.ts
```diff
@@ -0,0 +1,6 @@
+import {Breakpoint} from "../amplify-ui/styles/media";
+
+export const getNavHeight = (rems?: boolean) => {
+  const value = innerWidth > Breakpoint.LAPTOP * 16 ? 3 : 6.25;
```
jordanranz, line 4: Comment this

client/src/docs-ui/page/page.tsx
```diff
@@ -105,6 +105,15 @@ export class DocsPage {
     return this.getPageData();
   }
 
+  componentDidLoad() {
+    const {hash} = location;
+    if (hash) {
+      setTimeout(() => {
```
jordanranz, line 111: We need to solve this for the menu bug too. Chatting offline..

client/src/utils/scroll-to-hash.ts
```diff
@@ -0,0 +1,13 @@
+import {headingIsVisible} from "./heading-is-visible";
+import {getElementTop} from "./get-element-top";
+import {getNavHeight} from "./get-nav-height";
+
+export const scrollToHash = (hash: string): void => {
+  const targets = Array.from(document.querySelectorAll(hash)).filter((e) =>
```
jordanranz, line 6: We only care about hashes inside of page content right? Do we have to do a document query here?

client/src/utils/get-nav-height.ts
jordanranz: Actually let's just add an ID to the navbar then query this id. We can then use the height of the element

client/src/utils/scroll-to-hash.ts
harrysolovay: Yea––there might be multiple occurrences of the same ID

client/src/utils/get-nav-height.ts
```diff
@@ -0,0 +1,28 @@
...
+      return cachedRemsResult;
+    } else if (cachedPxResult) {
+      return cachedPxResult;
+    }
+  }
+
+  const nav = document.getElementById("secondary-nav");
+
+  if (nav) {
+    cachedAt = innerWidth;
```
jordanranz, line 19: What is this var for

PR #1420, opened 2020-04-02
chore: configured prerendering

harrysolovay:
*Issue #, if available:*

*Description of changes:*


jpignata:
One problem with --prerender is the code we've added to docs-internal-link to do URL resolution. In prerender, window.localtion is undefined, thus the split, pop thing we're doing there doesn't work. This causes the URLs to be set with a scheme, which causes Stencil to believe the URL's path is actually a host, and skips it in prerendering.

Anyway, I fixed that, but there are myriad other issues. Our decision to use query strings for platform for things like /lib and /sdk is causing entire sections of the site not to be crawled in the prerender process. How are we supposed to prerender files if the same path is used for wildly different content? etc.

harrysolovay:
@jpignata we want to support multiple filters down the line. Imagine a page that has platform and framework filtering, as well as version filtering.

We'd prefer to support this:

`http://docs.amplify.aws?platform=js&framework=react&version=3.0.5`

Than this:

`http://docs.amplify.aws/q/platform/js/framework/react/version/3.0.5`

I agree, this computation should happen elsewhere, but I don't believe this to be blocking. The end goal is certainly to get the correct filtration on first paint. In the meantime, we'll implement a spinner (shouldn't show for no more than a few milliseconds while the filter state loads into memory).

jpignata:
> @jpignata we want to support multiple filters down the line. Imagine a page that has platform and framework filtering, as well as version filtering.
> 
> We'd prefer to support this:
> 
> `http://docs.amplify.aws?platform=js&framework=react&version=3.0.5`
> 
> Than this:
> 
> `http://docs.amplify.aws/q/platform/js/framework/react/version/3.0.5`

The former form as implemented is suboptimal for several reasons, but let’s focus on the one I’m concerned about in this pull: how does your proposed site structure work with prerendering?

harrysolovay:
We initially had it implemented as the former (with filters existing as pieces of the path, rather than as query params). Let's chat about this in greater depth with the Nikhils and Jordan offline.

PR #1421, opened 2020-04-03
Android Lib - Storage - Getting Started updates

richardmcclellan:

### DIFF
--- a/docs/lib/storage/fragments/android/getting-started.md
+++ b/docs/lib/storage/fragments/android/getting-started.md
@@ -81,16 +81,17 @@ android {
 Add the following code to the bottom of your MainActivity `onCreate` method (ideally this would go in your Application class but this works for getting started quickly):
 
 ```java
-  AWSMobileClient.getInstance().initialize(getApplicationContext(), new Callback<UserStateDetails>() {
+    try {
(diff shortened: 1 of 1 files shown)


docs/lib/storage/fragments/android/getting-started.md
```diff
@@ -81,16 +81,17 @@ android {
 Add the following code to the bottom of your MainActivity `onCreate` method (ideally this would go in your Application class but this works for getting started quickly):
 
 ```java
-  AWSMobileClient.getInstance().initialize(getApplicationContext(), new Callback<UserStateDetails>() {
```
jpignata, line 84: My take on this is that it's more confusing for the user to see two success or error messages or a mix of the two, which the first example prevents by waiting for AWSMobileClient to be initialized before configuring Amplify and sending one message. I'm not sure this optimization is worth it given this will disappear in a couple of weeks. Would be interested to hear @TrekSoft weigh in, too.

richardmcclellan: Valid point.  I didn't realize AWSMobileClient is going away in a few weeks - given that, I agree it's not really worth worrying about this right now, so I'll go ahead and close this PR!

PR #1429, opened 2020-04-03
Update Android Getting Started intro/dependencies

jamesonwilliams:
The Android Getting Started guide made reference to the beta version of the code-generator plugin, and it made reference to an out of date version of the Android Gradle Plugin. Both versions have been updated to the latest.

The wording in the first paragraph of the Getting Started guide has also received another pass, towards improving its clarity of wording.


docs/lib/datastore/fragments/android/getting-started.md
```diff
@@ -1,10 +1,10 @@
 ## Model Generation
 
-The first step to integrate DataStore into your app is to model your data with a GraphQL schema, and *generate models* from that schema. The same GraphQL may be used to generate models for Android, iOS, and JavaScript. The same schemas are also used by the network protocol when synchronizing data with the cloud. Models can be generated by running a Gradle task in Android Studio, or by executing a command with the Amplify CLI.
+The first step towards integrating DataStore into your app is to model your data. To do so, you'll define a GraphQL schema. Amplify will then *generate Java classes* from that schema. The same GraphQL may be used to generate models for Android, iOS, and JavaScript. The same schema are also used by the DataStore's network protocol when synchronizing data with the cloud. Models can be generated by running a Gradle task in Android Studio, or by executing a command with the Amplify CLI.
```
jpignata, line 3: "The same schema is also" perhaps?

jpignata: It's also weird how we start two consecutive statements with "the same." Should GraphQL be a noun? Or should this be consolidated into:

> The same schema may be used to generate models across Android, iOS, and JAvaScript, and is used by DataStore's network protocol when synchronizing data with the cloud.

jpignata: Random though: is it stylized DataStore? IIRC, the left navigation says Datastore right now.

jamesonwilliams: I like the compound you suggest and will use it.

`DataStore` with a capital `S` is the stylization that is used in the source code. I think we should update the left-nav, to the same.

jpignata:
Small nit: are you certain the plural form of schema is schema? I've seen schemas commonly and haven't run into this form.

jamesonwilliams:
> Small nit: are you certain the plural form of schema is schema? I've seen schemas commonly and haven't run into this form.

I wasn’t sure. Here’s what Wikipedia says:

> The word schema comes from the Greek word σχήμα (skhēma), which means shape, or more generally, plan. The plural is σχήματα (skhēmata). In English, both schemas and schemata are used as plural forms.

It should be clear to all readers what “schemas” means. And I’ve done due diligence now to ensure I’ve not committed a grammatical sin. So - may it go forth as “schemas”! Update coming...

PR #1439, opened 2020-04-04
Optimize Android Datastore Examples for Screen Width

jamesonwilliams:
On a MacBook, there is very little width for the main content of the
documentation, on the V2 docs site.
    
To present a concise and readable example to the user, a balance needs
to be found between descriptive and meaningful phrasing, and getting to
the point.
    
Requiring the user to scroll left/right to understand basic examples
isn't helping.
    
Some basic variable renaming / formatting changes are made to push the
content of the examples to the left of the screen.


docs/lib/datastore/fragments/android/data-access.md
```diff
@@ -10,8 +10,8 @@ Post post = Post.builder()
     .build();
 
 Amplify.DataStore.save(post,
-    saved -> Log.i("DataStore", "Saved a post: " + saved.item()),
-    saveFailure -> Log.e("DataStore", "Save failed.", saveFailure)
+    saved -> Log.i("DataStore", "Saved a post."),
```
jpignata, line 13: ("While you're here" comment.) Consistency, hobgoblin, little minds: why not use the same tag on each log line? I'm not sure I see the value in some being set to "QuickStart" or "GetStarted" or whatever.

docs/lib/datastore/fragments/android/data-access.md
```diff
@@ -45,17 +40,12 @@ You can provide additional filters to your query using a **query predicate**. Th
...
-            Log.i("DataStore", "Rating: " + post.getRating());
-            Log.i("DataStore", "Status: " + post.getStatus());
-        }
-    },
-    queryFailure -> Log.e("DataStore", "Query failed.", queryFailure)
-);
+Amplify.DataStore.query(Post.class, Post.RATING.gt(4), goodPosts -> {
+    while (goodPosts.hasNext()) {
+        Post post = goodPosts.next();
+        Log.i("DataStore", "Title: " + post.getTitle());
```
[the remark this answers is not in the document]
jamesonwilliams: Yes, you're right! Good catch. I'll add it back.

docs/lib/datastore/fragments/android/data-access.md
jamesonwilliams: Agreed, also. Let's go with `"QuickStart"`, uniformly, everywhere.

jamesonwilliams:
> Small comments inline for your consideration.
> 
> Overall I think these changes make the examples clearer and more digestible. I do wonder though about the impetus for them: are you seeing the right navigation? As it's a responsive site, it may make sense for us to spend some time and optimize for specific, common resolutions (i.e. whatever yours is). If you play with the width and make it slightly narrower, does the right hand navigation disappear, revealing more content? Should we consider making the left navigation collapsible?

I do see the right-nav, yea. I think that's the experience issue, on my viewport.

I agree there are a number of UX solutions to this. Combining the left and right nav, making the left and/or right navs collapsible. Fixing the navs to the content, instead of the window...

Let me actually log an issue on the docs site for this so we can track it in further detail. I'll come back and update this comment with a link, when I have that.

docs/lib/datastore/fragments/android/data-access.md
jamesonwilliams: Eh, actually, I settled on `"GetStarted"`, which sounds less gimmicky. `Quick` looks like its about to drop its `c`, and do some telemarketing as a wonder-product.

PR #1440, opened 2020-04-05
Add Schema heading in Datastore guide

motleydev:
I kept missing this step because I assumed it was related to React Native as that was the previous heading. This helps navigate the steps better.

*Issue #, if available:*

*Description of changes:*

Added a heading around Modifying the Schema

### DIFF
--- a/js/datastore.md
+++ b/js/datastore.md
@@ -42,6 +42,7 @@ cd AmplifyDatastoreRN
 npx amplify-app@latest
 ```
 
+## Modifying the Schema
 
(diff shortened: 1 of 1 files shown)


swaminator:
@motleydev thanks for the submission. We are migrating to our v2 docs where we have addressed your feedback. Please refer to: https://docs.amplify.aws/lib/datastore/getting-started?platform=js#update-the-schema

If you feel that it doesn't address your solution, feel free to reopen the ticket.

PR #1459, opened 2020-04-06
large updates to JavScript Library GraphQL API category docmentation

dabit3:
Updates to Amplify GraphQL JavaScript Client documentation.


docs/lib/graphqlapi/concepts.md
```diff
@@ -0,0 +1,42 @@
...
+
+</amplify-callout>
+
+## Using AWS AppSync
+
+AWS AppSync helps you build data-driven apps with real-time and offline capabilities. Learn more about [AWS AppSync](https://aws.amazon.com/appsync/) by visiting [AWS AppSync Developer Guide](https://docs.aws.amazon.com/appsync/latest/devguide/welcome.html).
+
+The Amplify Framework offers three SDK options for AppSync.
+
+__[Amplify GraphQL client](~/lib/graphqlapi/query-data.md)__ - a light weight option if you're looking for a simple way to leverage GraphQL features and do not need the offline capabilities or caching. If you need those features, please look at [Amplify DataStore](https://aws-amplify.github.io/docs/js/datastore).
```
sammartinez, line 30: Do we want to point to our new docs for DataStore?

dabit3: Good catch, just pushed this update.

docs/lib/graphqlapi/fragments/graphql-from-node.md
```diff
@@ -0,0 +1,86 @@
+You can also call an AppSync GraphQL API from a NodeJS app or a Lambda function. For example, deploy a simple `Todo` model with the following schema in the `amplify add api` flow:
```
sammartinez, line 1: @mlabieniec / @undefobj  What do you think about adding this, we are pretty explicit about not having Node callouts in Amplify. This would be new.

undefobj: We have Node callout in the existing docs which is what Nader is porting over, not a new thing. What we're explicit is that Amplify JS isn't supported in Node. The multi-auth feature was actually added for this reason.

dabit3: Do you think I should also add an example of calling with an API key?

PR #1460, opened 2020-04-07
Update GraphQL API getting started to gold standard

richardmcclellan:
https://app.asana.com/0/1158799035447930/1169671161142084


docs/lib/graphqlapi/fragments/android/getting-started.md
```diff
@@ -1,41 +1,17 @@
-You can integrate with Amplify framework using the following steps:
+> ### Prerequisites
+> * An Android project targeting Android API level 16 (Android 4.1) or above.
```
jpignata, line 2: Can we be consistent with punctuation on bullets? I'd prefer without, but "both" is the wrong answer. Also! I wonder if we can link to Google's Android documentation here directly to their getting started steps.

docs/lib/graphqlapi/fragments/android/getting-started.md
```diff
@@ -1,41 +1,17 @@
...
+> * [Install and configure](~/cli/start/install.md) the Amplify CLI
 
-1. Setup the API endpoint and authentication information in the client side configuration.
-2. Generate Java Model classes from the API schema.
-3. Write app code to run queries, mutations and subscriptions.
+## GraphQL API with Amplify
 
-## Set up your backend
+The Amplify GraphQL API category provides a solution for making HTTP requests to GraphQL endpoints.  It supports [AWS AppSync](https://aws.amazon.com/appsync/) or any other GraphQL backend.
```
jpignata, line 7: This actually isn't true at the moment. We only support AWS AppSync, I believe.

jpignata, line 7: "Amplify GraphQL API category" isn't the right terminology here probably. We should say "The Amplify API category provides a solution for making HTTP requests to REST and GraphQL endpoints. For GraphQL, it supports ..." something like that.

docs/lib/graphqlapi/fragments/android/getting-started.md
brene: Same. prefer without punctuation on bullets

docs/lib/graphqlapi/fragments/android/getting-started.md
brene: @jpignata - I was just chatting with @nikhil-dabhade about this and he pointed that the JS libraries do support other GQL backends as well. Is this Android-specific?

jpignata: Yes, JavaScript does, but neither iOS or Android do at the moment, I believe.

docs/lib/graphqlapi/fragments/android/getting-started.md
richardmcclellan: Good suggestions!  I will:
 - remove punctuation
 - add link to Google Android docs

docs/lib/graphqlapi/fragments/android/getting-started.md
richardmcclellan: I actually copied this from the JS page, so good to know that iOS and Android don't support it yet :) Will remove.

richardmcclellan: Makes sense!

PR #1465, opened 2020-04-07
Default and impossible paths

harrysolovay:
* when on a JS lib page, and switching platforms, SDK is now the default
* /sdk js option -> /lib js
* /lib ios and android option -> sdk ios and android


client/src/docs-ui/choose-anchor/choose-anchor.tsx
```diff
@@ -30,7 +30,14 @@ export class DocsChooseAnchor {
             Object.entries(filterMetadataByOptionByName[filterKey]).map(
               ([filterValue, {label, graphicURI}]) => {
                 const route =
-                  this.page && `${this.page.route}?${filterKey}=${filterValue}`;
+                  this.page?.route &&
+                  filterValue === "js" &&
+                  this.page.route.startsWith("/sdk")
```
jordanranz, line 35: Let's break these out to their own functions

harrysolovay: Done

PR #1474, opened 2020-04-08
New url scheme

harrysolovay:
Not yet ready for a review.


client/src/docs-ui/universal-nav/universal-nav.style.ts
```diff
@@ -95,11 +95,3 @@ export const hideAboutLinkStyle = css`
     display: initial;
   }
 `;
-
-export const hideOpenSourceLinkStyle = css`
-  display: none;
-
-  ${MQMobile} {
```
jordanranz, line 102: You should be able to remove the MQMobile import here now

PR #1477, opened 2020-04-08
Instructions and code sample: text to speech

wooj2:
I found the code sample did not actually emit audio :(.  Seems like `player` needs to be saved in an instance variable and stick around in memory in order for the audio to be played.  Previously player was going out of scope, and the audio was not being played.

### DIFF
--- a/docs/lib/predictions/fragments/ios/text-speech.md
+++ b/docs/lib/predictions/fragments/ios/text-speech.md
@@ -64,4 +64,4 @@ func textToSpeech(text: String) {
     })
 }
 ```
-As a result of running this code, you should hear audio of the text being emitted from your device.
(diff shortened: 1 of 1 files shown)


docs/lib/predictions/fragments/ios/text-speech.md
```diff
@@ -2,38 +2,66 @@
...
+      if let player = self.player {
+        player.play()
+      }
+      default:
+        print("")
+      }
+    })
 }
 ```
+As a result of running this code, you should hear audio of the text being emitted from your device.
```
jpignata, line 67: Should, or will? :trollface:

wooj2: updating :)

PR #1481, opened 2020-04-08
Update identify-text.md

wooj2:


docs/lib/predictions/fragments/ios/identify-text.md
```diff
@@ -35,6 +54,29 @@ func detectText(_ image: URL) {
...
+let xBoundingBox = word.boundingBox.origin.x * self.imageView.frame.size.width
+let yBoundingBox = word.boundingBox.origin.y * self.imageView.frame.size.height
+let widthBoundingBox = word.boundingBox.size.width * self.imageView.frame.size.width
+let heightBoundingBox = word.boundingBox.size.height * self.imageView.frame.size.height
+let boundingBox = UIView(frame: CGRect(x: xBoundingBox, y: yBoundingBox, width: widthBoundingBox, height: heightBoundingBox))
+... // stylize bounding box
+self.imageView.addSubview(boundingBox)
+```
+Note that Rekognition places (0,0) at the top left and Core ML places (0,0) at the bottom left, we have flipped the y axis of the CoreML bounding box for you since iOS starts (0,0) from the top left
+
```
wooj2, line 79: Not sure if this is still relevant, can someone help me understand if this is even relevant to say anymore?

kneekey23: its relevant and this addition is actually more helpful since the sample apps arent available yet

docs/lib/predictions/fragments/ios/identify-text.md
```diff
@@ -35,6 +54,29 @@ func detectText(_ image: URL) {
...
+To get results that utilize on-device capabilities (Core ML), without combining results from the backend, you can use the following to pass into the `options` argument of the `Amplify.Predictions.identify` function.
+```swift
+let options = PredictionsIdentifyRequest.Options(defaultNetworkPolicy: .offline, pluginOptions: nil)
+```
+
+Note that bounding boxes in `IdentifyTextResult` are returned as ratios.  If you would like to place bounding boxes on individual recognized words that appear in the image, use the following method to calculate a frame for a single bounding box.
+```swift
+let image = UIImage(...)
+self.imageView = UIImageView(image: self.image)                        
+...
```
brene, line 66: Let's try to avoid un-runnable code.  Can we put a comment here to educate the user what kind of code will be in this area?

docs/lib/predictions/fragments/ios/identify-text.md
```diff
@@ -35,6 +54,29 @@ func detectText(_ image: URL) {
...
+let data = result as! IdentifyTextResult
+let word = data.words!.first!
+...
```
brene, line 69: same as above

docs/lib/predictions/fragments/ios/identify-text.md
```diff
@@ -35,6 +54,29 @@ func detectText(_ image: URL) {
...
+let xBoundingBox = word.boundingBox.origin.x * self.imageView.frame.size.width
+let yBoundingBox = word.boundingBox.origin.y * self.imageView.frame.size.height
+let widthBoundingBox = word.boundingBox.size.width * self.imageView.frame.size.width
+let heightBoundingBox = word.boundingBox.size.height * self.imageView.frame.size.height
+let boundingBox = UIView(frame: CGRect(x: xBoundingBox, y: yBoundingBox, width: widthBoundingBox, height: heightBoundingBox))
+... // stylize bounding box
```
brene, line 75: same

docs/lib/predictions/fragments/ios/identify-text.md
```diff
@@ -35,6 +54,29 @@ func detectText(_ image: URL) {
...
 ```
 
+To get results that utilize on-device capabilities (Core ML), without combining results from the backend, you can use the following to pass into the `options` argument of the `Amplify.Predictions.identify` function.
+```swift
+let options = PredictionsIdentifyRequest.Options(defaultNetworkPolicy: .offline, pluginOptions: nil)
+```
+
+Note that bounding boxes in `IdentifyTextResult` are returned as ratios.  If you would like to place bounding boxes on individual recognized words that appear in the image, use the following method to calculate a frame for a single bounding box.
+```swift
+let image = UIImage(...)
```
brene, line 64: what's the expectation here instead of `...`?  maybe to be a bit more explicit `UIImage(NAME_OF_YOUR_ASSET_FILE)`

docs/lib/predictions/fragments/ios/identify-text.md
brene: essentially put the comment indicator up front.
`// ... stylize bounding box`

docs/lib/predictions/fragments/ios/identify-text.md
```diff
@@ -2,24 +2,43 @@
...
+Amplify will make calls to both Amazon Textract and Rekognition depending on the type of text you are looking to identify (i.e. image or document).
+
+If you are detecting text from an image you would send in `.plain` as your text format as shown below.  Using `.plain` with `PredictionsIdentifyRequest.Options()` combines results from on device results from Core ML and AWS services to yield more accurate results.
 
 ``` swift
+import Amplify
+
+...
```
brene, line 40: I don't think the `...` is necessary. or change it to 
```
import Amplify

// ... other code in your class

func ...
```

docs/lib/predictions/fragments/ios/identify-text.md
wooj2: Hmm.. seems pretty clear to me.  Do you have any proposals as to what would make it clearer?

brene: As a user, copy and pasting this content would also require me to manually remove multiple lines of ... just to run the code quickly to verify if it’s what I’m looking for. At least move the ... behind a comment

wooj2: Good point, but maybe the bigger question is that maybe we should remove all of this content and leave it just for a sample app?  Who cares about bounding rectangles?

docs/lib/predictions/fragments/ios/identify-text.md
wooj2: sure.

docs/lib/predictions/fragments/ios/identify-text.md
wooj2: removing, we'll figure out a place to put this info, since this is common to multiple places in predictions.

wooj2:
Added task to find place for where bounding rectangle info should go:
https://github.com/aws-amplify/docs/issues/1489

PR #1484, opened 2020-04-08
Update identify entity removed bounding box info

wooj2:
* Removing bound box info.  If this info is worthy, we should have it in an appendix so it applies to all sections.
* I was unable to test the portion around "match entities from a pre-created "Amazon Recognition Collection" -- does anyone have an example I can use?  The documentation that we link to isn't very helpful in this context.


docs/lib/predictions/fragments/ios/identify-entity.md
```diff
@@ -2,23 +2,38 @@
...
 ```
+Run `amplify push` to create the resources in the cloud
+
 
 ## Working with the API
 
-In order to match entities from a pre-created [Amazon Rekognition Collection](https://docs.aws.amazon.com/rekognition/latest/dg/collections.html), make sure there is a `collectionId` set in your `amplifyconfiguration.json` file. If there is no `collectionId` set in the `amplifyconfiguration.json` file, then this call will just detect entities in general with facial features, landmarks, etc. Bounding boxes for entities are returned as ratios so make sure if you would like to place the bounding box of your entity on an image that you multiple the x by the width of the image, the y by the width of the image, and both height and width ratios by the image's respective height and width.
+In order to match entities from a pre-created [Amazon Rekognition Collection](https://docs.aws.amazon.com/rekognition/latest/dg/collections.html), make sure there is a `collectionId` set in your `amplifyconfiguration.json` file. If there is no `collectionId` set in the `amplifyconfiguration.json` file, then this call will just detect entities in general with facial features, landmarks, etc.
```
wooj2, line 36: This part was totally confusing  --anyone have any example i can use, and then i can work on rewriting it?

wooj2:
Opened an issue to figure out where bounding rectangle & coordinate info should go:
https://github.com/aws-amplify/docs/issues/1489

docs/lib/predictions/fragments/ios/identify-entity.md
wooj2: After about 1-2 hours of troubleshooting and reading through docs, I was unable to get this portion to work.

I went through instructions of creating a collection, adding a face to a collection, creating a s3 bucket, creating a user with the right permissions specifying the configuration in my amplifyconfiguration.json, but in all cases, I can not get this to work.

In all cases, the calls to identify resolved to a `IdentifyEntitiesResult` rather than the expected `IdentifyEntityMatchesResult`.

Couple options:
1.  Just change the example to only return "IdentifyEntitiesResult", and don't take users through the `IdentifyEntityMatchesResult` as this has a more involved setup.

2.  Have the original devs look into this and update documentation appropriately.

3.  Something else?

jpignata: I vote #1 for now. We should create a bug in the backlog for the `IdentifyEntityMatchesResult` path and add documentation somewhere later in the document. That documentation should be far clearer about how to set this up (i.e., show me how to set a `collectionId` rather than telling me it should be set. Set where?)

wooj2: @jpignata  -- sorry for the delay, Sounds good, i'll update the document!  To make sure this gets addressed, i created a ticket in pivotal tracker.

PR #1487, opened 2020-04-09
Update identify-celebrity.md

wooj2:


docs/lib/predictions/fragments/ios/identify-celebrity.md
```diff
@@ -5,32 +5,50 @@ If you haven't already done so, run `amplify init` inside your project and then
...
+      print(error)
+    default:
+      print("")
+    }
+  })
 }
-```
\ No newline at end of file
+```
+As a result of passing in a URL of an image of a celebrity, you should see the corresponding celebrity name printed to the screen along with additional metadata.
```
jpignata, line 54: Should or will? Maybe provide a URL to a picture of Jeff Bezos or something in the code? Though, maybe a bad idea since the others don't have examples driving the code.

kneekey23: agree ^

wooj2: This isn't something I feel strongly about, and happy to change it to whatever is agreed to, but not exactly sure what to change the code to.  The options i came up with seem to detract from the original intent of the sample code:

Option 1: add a function:
```swift
@IBAction func didTapButton(_ sender: Any) {
        let imageOfBezos = URL(string:"http://bezos..")!
        detectCelebs(imageOfBezos)
}
```
I don't like this because
a. I feel it detracts from the code we are trying to show
b.  Additional reasons below

Option 2: Without a function
```swift
let imageOfBezos = URL(string:"http://bezos..")!
detectCelebs(imageOfBezos)
```
I don't like this either, because:
a.  we don't have a URL which will be guaranteed
b.  Having to use a force unwrap is poor practice
c.  It's now out of a function, which i've gotten push back in another PR.

Anyways -- happy to leave it as is, or anything you guys want to change it to.

jpignata: Feel free to ignore. Just an idea. :) Perhaps in the "should" line we can provide a stable example URL and expected output.

brene: How about using Jeff's twitter photo link? https://pbs.twimg.com/profile_images/669103856106668033/UF3cgUk4_400x400.jpg 

Anyway no strong feelings on this either.

wooj2: I'm going to ignore for now given there isn't clear consensus.  Feel free to open a ticket or gather the team to solicit ideas on how to make this better.

Regarding twitter -- again, we do not control this endpoint, so if the URL were to change, our example would be broken.

PR #1544, opened 2020-04-10
Reload page on changes in development

jpignata:
*Issue #, if available*: n/a

*Description of changes:* Reload page on changes in development

For whatever reason, the default reload behavior of Hot Module
Replacement is not working within this project which results in the
content author being forced to tediously reload on changes. This change
configures the behavior to `pageReload` which will do a full refresh of
the page on rebuild.

See: https://stenciljs.com/docs/dev-server#dev-server-config

By submitting this pull request, I confirm that you can use, modify, copy, and
redistribute this contribution, under the terms of your choice.

### DIFF
--- a/client/stencil.config.ts
+++ b/client/stencil.config.ts
@@ -3,6 +3,9 @@ import nodePolyfills from "rollup-plugin-node-polyfills";
 
 export const config: Config = {
   globalScript: "src/global/global.ts",
+  devServer: {
+    reloadStrategy: 'pageReload'
+  },
   nodeResolve: {
     browser: true,
   },


brene:
+1 @jordanranz

harrysolovay:
@jpignata b e a u t i f u l . Any idea why it triggers two refreshes instead of just one?

In the future––once we make the content accessible through a GraphQL API––we can utilize subscriptions for a cleaner dev workflow.

jpignata:
> @jpignata b e a u t i f u l . Any idea why it triggers two refreshes instead of just one?

None. I noticed that, but didn't dig into it. On the later list above it: why doesn't hmr work?

Issue #1545, opened 2020-03-20
How can DataStore observe changes on one item?


waltermvp:
** Which Category is your question related to? **
DataStore / API

** What AWS Services are you utilizing? **
DynamoDB Appsync

** Provide additional details e.g. code snippets **
The following observes the entiire Events table:

```
    const subscription = DataStore.observe(Event).subscribe(msg => {
      console.log(msg.model, msg.opType, msg.element);
      onQuery();
    });
```

How can I observe changes for one specific Event item using an `id`?

PR #1548, opened 2020-04-10
by ID example of JS DataStore Observe, re issue #1545

dabit3:
*Issue #, if available:*
#1545

### DIFF
--- a/docs/lib/datastore/fragments/js/data-access/observe-snippet.md
+++ b/docs/lib/datastore/fragments/js/data-access/observe-snippet.md
@@ -16,6 +16,6 @@ const subscription = DataStore.observe(Post, id).subscribe(msg => {
 
 <amplify-callout>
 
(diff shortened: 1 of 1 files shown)


docs/lib/datastore/fragments/js/data-access/observe-snippet.md
```diff
@@ -4,8 +4,18 @@ const subscription = DataStore.observe(Post).subscribe(msg => {
 });
 ```
 
+Observing changes of a single item by ID.
+
+```js
+const id = '69ddcb63-7e4a-4325-b84d-8592e6dac07b';
```
drochetti, line 10: nit: we've been using `"123"` as a representation of ids in the rest of the examples, maybe we should keep it consistent with those?

dabit3: Hey Daniel, thank you for pointing this out, I'll update in my next PR.

PR #1557, opened 2020-04-11
Hide copy button for console snippets

jpignata:
*Issue #, if available:* n/a

*Description of changes:* Hide copy button for console snippets

Right now, we show the copy button for output that's meant to be
expository rather than a block we expect a developer to copy and paste.
The most common example is CLI output: while this is useful to see to
understand what options should be selected, there's no value in copying
it into your clipboard.

This change suppresses the copy button if the language is set to
"console" which is defined in Linguist. See: https://bit.ly/2XsAkBq

### DIFF
--- a/capi/src/init-node/index.ts
+++ b/capi/src/init-node/index.ts
@@ -60,7 +60,7 @@ const highlight = (code: string, language: string): string => {
     languageIsSet ? `-${language}` : ""
   }">${highlighted}</div>`;
 
-  return `<amplify-code-block line-count="${String(
+  return `<amplify-code-block language="${language}" line-count="${String(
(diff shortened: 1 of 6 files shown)


jpignata:
**Example**

https://pr-1557.d39agkzf0cwtdz.amplifyapp.com/lib/analytics/getting-started?platform=android

jpignata:
@jordanranz @harrysolovay Ping? No good?

jpignata:
After I get a :thumbsup: from Harry and Jordan, I'll add console as the language to all output snippets.

harrysolovay:
@jpignata lgtm!

PR #1564, opened 2020-04-13
Swap SDK and Libraries in Slider

jpignata:
*Issue #, if available:* n/a

*Description of changes:* Swap SDK and Libraries

Once Libraries are GA we can swap them back (and remove the parentheticals) but at the moment the default option being second doesn't make a lot of sense.

### DIFF
--- a/client/src/docs-ui/menu/menu.tsx
+++ b/client/src/docs-ui/menu/menu.tsx
@@ -32,14 +32,14 @@ export class DocsMenu {
       return (
         <docs-version-switch
           leftOption={{
-            title: "Libraries",
-            subTitle: "(preview)",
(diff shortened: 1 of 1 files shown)


brene:
I like the swap but am not sure about "current". It doesn't provide the same clear guidance as "stable".

jpignata:
This might be a matter of taste. Happy to let it drop if nobody has a strong opinion, and just go with what's on the page now.

jpignata:
[Submitted for your approval](https://scifi.stackexchange.com/questions/127987/in-which-episodes-does-rod-serling-say-submitted-for-your-approval): I ditched the wording change, but kept the order change.

PR #1581, opened 2020-04-15
GraphQL API Sample Code Update

richardmcclellan:
*Description of changes:*
Since `api` does not require `auth`, the sample code should not include initialization of AWSMobileClient, because AWSMobileClient requires `auth`.   Initializing AWSMobileClient without auth configured results in a RuntimeException.

### DIFF
--- a/docs/lib/graphqlapi/fragments/android/getting-started.md
+++ b/docs/lib/graphqlapi/fragments/android/getting-started.md
@@ -144,9 +144,7 @@ When the build is successful, it adds two gradle tasks to you project - `modelge
 Add the following imports at the top of your MainActivity and code at the bottom of the `onCreate` method (ideally this would go in your Application class but this works for getting started quickly):
 
 ```java
-import com.amazonaws.mobile.client.AWSMobileClient;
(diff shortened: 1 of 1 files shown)


docs/lib/graphqlapi/fragments/android/getting-started.md
```diff
@@ -157,22 +155,13 @@ public class MainActivity extends AppCompatActivity {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_main);
 
-        AWSMobileClient.getInstance().initialize(getApplicationContext(), new Callback<UserStateDetails>() {
```
drochetti, line 160: Is that a problem in other languages' snippets as well? If so we should either fix them too or follow up with owners of those platforms.

richardmcclellan: Good question.  I just checked iOS and JS, and they do not have this same issue.

richardmcclellan:
I just checked the code samples on subsequent pages, and don't see anything that assumes auth is present, so think this change should be fine!

Issue #1489, opened 2020-04-09
Find place to put information on bounding rectangle & coordinate info


wooj2:
as part of this PR:
https://github.com/aws-amplify/docs/pull/1481

It's a bit confusing on where to put information with regards to creating bounding rectangles, since it applies to the following pages:
https://docs.amplify.aws/lib/predictions/identify-entity?platform=ios
https://docs.amplify.aws/lib/predictions/label-image?platform=ios
https://docs.amplify.aws/lib/predictions/identify-celebrity?platform=ios

The content that describes at a high level how to deal with bounding boxes was originally incorrect:

---
 Bounding boxes for entities are returned as ratios so make sure if you would like to place the bounding box of your entity on an image that you multiple the x by the width of the image, the y by the width of the image, and both height and width ratios by the image’s respective height and width.

And proposed to change to something along the lines of:

---

Note that bounding boxes in `IdentifyTextResult` are returned as ratios.  If you would like to place bounding boxes on individual recognized words that appear in the image, use the following method to calculate a frame for a single bounding box.

```swift
let image = UIImage(...)
self.imageView = UIImageView(image: self.image)                        
...
let data = result as! IdentifyTextResult
let word = data.words!.first!
...
let xBoundingBox = word.boundingBox.origin.x * self.imageView.frame.size.width
let yBoundingBox = word.boundingBox.origin.y * self.imageView.frame.size.height
let widthBoundingBox = word.boundingBox.size.width * self.imageView.frame.size.width
let heightBoundingBox = word.boundingBox.size.height * self.imageView.frame.size.height
let boundingBox = UIView(frame: CGRect(x: xBoundingBox, y: yBoundingBox, width: widthBoundingBox, height: heightBoundingBox))
... // stylize bounding box
self.imageView.addSubview(boundingBox)
```
Note that Rekognition places (0,0) at the top left and Core ML places (0,0) at the bottom left, we have flipped the y axis of the CoreML bounding box for you since iOS starts (0,0) from the top left

wooj2:
Resolving, updated code sample in #1600

PR #1600, opened 2020-04-17
commiting change to reflect matching entities from a collection and update bounding boxes

kneekey23:
*Description of changes:*
Corrected documentation to adequately set this up and get it working so adding it back in. Tested and it works so no actual bug just a docs error. 🙃 Additionally added bounding box info back to these two pages since the information is a bit different in both places addressing issue #1489 


docs/lib/predictions/fragments/ios/identify-entity.md
```diff
@@ -33,6 +33,26 @@ Run `amplify push` to create the resources in the cloud
 
 ## Working with the API
 
+In order to match entities from a pre-created [Amazon Rekognition Collection](https://docs.aws.amazon.com/rekognition/latest/dg/collections.html), make sure these two properties,`collectionId` and `maxEntities`, are set in your `amplifyconfiguration.json` file. `collectionId` should be the  name of your collection that you created either with the CLI or the SDK and `maxEntities` should be a number greater than `0` or less than or equal to `50` which is the max number of entities Rekognition can detect from a collection. If there is no `collectionId` set in the `amplifyconfiguration.json` file or `maxEntities` is set to be `0`, then this call will just detect entities in general with facial features, landmarks, etc. Bounding boxes for entities are returned as ratios so make sure if you would like to place the bounding box of your entity on an image that you multiple the x by the width of the image, the y by the height of the image, and both height and width ratios by the image's respective height and width.
```
wooj2, line 36: * Instead of "make sure these two properties...", maybe use: "ensure that both collectionId and maxEntities are set in your amplify... file"
* The value of `collectionId` should be the name of the ..
* separate sentence between explaining the values for `collectionId` and `maxEntities` (it's currently a run on sentence)
* "should be a number greater than.." .. replace with: "should be a number greater than 0 and less than 51."
* change "If there is no `collectionId` set in the `amplifyconfiguration.json` file or `maxEntities` is set to be `0`" to "If both collectionId and maxEntities do not have valid values in `amplifyconfiguration.json` then, this call will instead detect entities..
* I think we should remove info about bounding box... and find a common place for it.  I opened a ticket for it here:
  * https://github.com/aws-amplify/docs/issues/1489

kneekey23: k fixed all

docs/lib/predictions/fragments/ios/identify-text.md
```diff
@@ -50,6 +50,25 @@ func detectText(_ image: URL) {
...
+	let data = result as! IdentifyTextResult
+	let word = data.words!.first!
+	...
+	let xBoundingBox = word.boundingBox.origin.x * self.imageView.frame.size.width
+	let yBoundingBox = word.boundingBox.origin.y * self.imageView.frame.size.height
+	let widthBoundingBox = word.boundingBox.size.width * self.imageView.frame.size.width
+	let heightBoundingBox = word.boundingBox.size.height * self.imageView.frame.size.height
+	let boundingBox = UIView(frame: CGRect(x: xBoundingBox, y: yBoundingBox, width: widthBoundingBox, height: heightBoundingBox))
+	... // stylize bounding box
+	self.imageView.addSubview(boundingBox)
```
wooj2, line 67: Ok, so this code as pointed out by @brene is difficult to copy and paste.  So lets do this
1.  Update the code above to have a completion block like the following:
```
    func detectText(_ image: URL, completion: @escaping ([IdentifiedWord]) -> Void) {
        _ = Amplify.Predictions.identify(type: .detectText(.plain), image: image, options: PredictionsIdentifyRequest.Options(), listener: { (event) in
            switch event {
            case .completed(let result):
                let data = result as! IdentifyTextResult
                completion(data.words!)
            case .failed(let error):
                print(error)
            default:
                print("")
            }
        })
    }
```
2. Update the bounding box example to be something like this:
```
    @IBAction func didTapButton(_ sender: Any) {
        let imageURL = URL(string: "https://imageWithText")
        let data = try? Data(contentsOf: imageURL!)
        let image = UIImage(data: data!)
        let imageView = UIImageView(image: image)
        self.view.addSubview(imageView)

        detectText(imageURL!, completion: { words in
            let word = words.first!
            DispatchQueue.main.async {
                let xBoundingBox = word.boundingBox.origin.x * imageView.frame.size.width
                let yBoundingBox = word.boundingBox.origin.y * imageView.frame.size.height
                let widthBoundingBox = word.boundingBox.size.width * imageView.frame.size.width
                let heightBoundingBox = word.boundingBox.size.height * imageView.frame.size.height
                let boundingBox = UIView(frame: CGRect(x: xBoundingBox, y: yBoundingBox, width: widthBoundingBox, height: heightBoundingBox))
                boundingBox.backgroundColor = .red
                imageView.addSubview(boundingBox)
            }
        })
    }
```

kneekey23: updated

PR #1601, opened 2020-04-17
Nader/js auth updates

dabit3:
- Adds Angular and Vue examples for UI components in authentication guide for JS
- Fixes React export in example
- Fixes incorrect Vue configuration in getting started guide


brene:
LGTM. @sammartinez - any comments?

PR #1614, opened 2020-04-21
Add "read" operation to editors example

kay-is:
An editor should be able to `read` what they `update`.

*Description of changes:*

The example gives editors only `update` rights.

While it's okay to give anyone rights as you like, it seems to me that editors should also be able to `read` what they want to `update`.

I was a bit confused, how Amplify would calculate `read` permissions for an editor with the given example rules.


brene:
@undefobj - can you review this?

brene:
@kay-is - thanks for the contribution!

PR #1631, opened 2020-04-23
a couple of updates to DataStore docs

dabit3:
- Added reference for how to import datastore.
- Moved real-time into it's own section.


docs/lib/datastore/data-access.md
```diff
@@ -3,6 +3,14 @@ title: Manipulating data
...
 ---
 
+## Getting started
+
+To get started, first import the `DataStore` API:
+
+```js
+import { DataStore } from '@aws-amplify/datastore';
+```
+
```
drochetti, line 13: this needs to go into a JS fragment, otherwise it will show for iOS and Android as well.

The Getting Started page mentions the imports I believe, is it something we could potentially standardize?

docs/lib/datastore/real-time.md
```diff
@@ -0,0 +1,12 @@
...
+description: Learn more about how DataStore handles data changes in real-time.
+---
+
+## Observe Data in real-time
+
+You can subscribe to changes on your Models. This reacts dynamically to updates of data to the underlying Storage Engine, which could be the result of GraphQL Subscriptions as well as Queries or Mutations that run against the backing AppSync API if you are synchronizing with the cloud.
+
+<inline-fragment platform="js" src="~/lib/datastore/fragments/js/data-access/observe-snippet.md"></inline-fragment>
+<inline-fragment platform="ios" src="~/lib/datastore/fragments/ios/data-access/observe-snippet.md"></inline-fragment>
+<inline-fragment platform="android" src="~/lib/datastore/fragments/android/data-access/observe-snippet.md"></inline-fragment>
```
drochetti, line 12: I like the fact we have a section dedicated to real-time. My suggestion would be to move these fragments to their own folder (move from `data-access/` to `real-time/` to match the page it belongs to)

docs/lib/datastore/data-access.md
dabit3: Thanks, updates pushed based on this feedback.

docs/lib/datastore/real-time.md
dabit3: Thanks, updates pushed based on this feedback.

PR #1632, opened 2020-04-23
Subtle fixes

harrysolovay:
* fix floating action button dismissal issue on mobile
* clean up feedback template link
* ensure TOC doesn't show when there aren't any h2s or h3s
* add space between next & previous links on mobile to prevent this:

Screen Shot 2020-04-23 at 9 51 27 AM

* add subtle shadow to indicate when secondary nav section is scrollable

Screen Shot 2020-04-24 at 6 51 03 PM

* hide menu on mobile when item or filter selected


client/src/docs-ui/repo-actions/repo-actions.tsx
```diff
@@ -11,10 +11,10 @@ export class DocsRepoActions {
     return (
       <Host class={repoActionsStyle}>
         <amplify-external-link
-          href={`https://github.com/aws-amplify/docs/issues/new?title=[feedback]&labels=v2&body=${encodeURI(
-            `**Page**: [\`${location.href}\`](${location.href})
+          href={`https://github.com/aws-amplify/docs/issues/new?title=Feedback&body=${encodeURI(
```
jordanranz, line 14: I think the square bracket here is better, it should be a capital 'F' tho. Also you should make this common if possible since it is used in multiple places.

client/src/amplify-ui/sidebar-layout/sidebar-layout-toggle/sidebar-layout-toggle.tsx
```diff
@@ -24,15 +21,23 @@ export class AmplifySidebarLayoutToggle {
    */
   getClass = () => {
     return {
-      [unclickableStyle]: !!this.inView,
       [sidebarLayoutSidebarToggleStyle]: true,
       ...(this.inViewClass ? {[this.inViewClass]: !!this.inView} : {}),
     };
   };
 
+  safeToggle = (e: Event) => {
```
jordanranz, line 29: `safeToggle` is unclear why it is "safe" can you rename or add a comment

client/src/docs-ui/repo-actions/repo-actions.tsx
brene: Agree square brackets are nice

jordanranz: Additionally, I believe this needs to change in `feedback-callout`. All the more reason to have a common const or function for this

PR #1637, opened 2020-04-23
Update protected access for amplify storage

wooj2:


docs/lib/storage/fragments/ios/configureaccess.md
```diff
@@ -1,19 +1,26 @@
...
 }
 ```
 
-When uploading a file with `protected` access level, users can only read the file and only the user which created the file can delete it.
+When uploading a file with `.protected` access level:
+* Owners have the ability to read, write, and delete access
+* Other users only have read access
 
-Another user that wants to read the file can specify the user that created it:
+If another user wants to read the file, you will need to include the `targetIdentityId` parameter:
```
jpignata, line 18: Above it says other users only have read access, but this states users need to be included in the `targetIdentifyId` parameter to gain access.. I'm a bit confused. "If another user wants to read the file" -- does this mean to say if I want to grant read access to a specific user?

docs/lib/storage/fragments/ios/configureaccess.md
```diff
@@ -1,19 +1,26 @@
...
+let dataString = "My Data"
+let data = dataString.data(using: .utf8)!
+let options = StorageUploadDataRequest.Options(accessLevel: .protected)
+Amplify.Storage.uploadData(key: "userProtectedKey", data: data, options: options) { (event) in
     // handle result
 }
 ```
 
-When uploading a file with `protected` access level, users can only read the file and only the user which created the file can delete it.
+When uploading a file with `.protected` access level:
```
jpignata, line 14: Is this defining what the protected access level is? The "when uploading" part makes it less clear.

docs/lib/storage/fragments/ios/configureaccess.md
```diff
@@ -1,19 +1,26 @@
 ## Protected access
 
-Create an options object with the protected access level to restrict access for certain objects.
+Create an `Options` object with the `.protected` access level to restrict access for your key.
```
jpignata, line 3: Niggle: The key refers to the location of an object, or, to be storage system agnostic, a file. What we're restricting access to is the file or object, not the key.

docs/lib/storage/fragments/ios/configureaccess.md
wooj2: Chatted offline: need to update to be more clear about if user1 uploads a protected file, then user2 can access this file in the following manner.

docs/lib/storage/fragments/ios/configureaccess.md
wooj2: Chatted offline: need to reference the code sample above that we are uploading a file, and as a result of using the code above that the .protected access level is that owerns will have the ability to read/write/delete etc & other users will have read access (assuming they use the .protected flag and targetIdentityId

PR #1648, opened 2020-04-27
feat: implement sitemap generation

harrysolovay:
Utilizes Capi to do so


capi/src/write/sitemap.ts
```diff
@@ -0,0 +1,42 @@
...
+  for (const [, page] of ctx.pageBySrcPath.entries()) {
+    routes.set(page.route, true);
+    if (page.versions) {
+      Object.entries(page.versions).forEach(([, version]) =>
+        routes.set(version, true),
+      );
+    }
+  }
+
+  const src = `<?xml version="1.0" encoding="utf-8" standalone="yes" ?>
```
jordanranz, line 29: What does standalone="yes" do?

capi/src/write/sitemap.ts
```diff
@@ -0,0 +1,42 @@
...
+  return ` <url>
+    <loc>http://docs.amplify.aws${route}</loc>
+    <lastmod>${yyyyMmDd}</lastmod>
+    <changefreq>hourly</changefreq>
+    <priority>.5</priority>
+  </url>`;
+}
+
+export async function sitemap(config: t.Config, ctx: t.Ctx): Promise<void> {
+  const routes = new Map<string, true>();
```
jordanranz, line 18: Why a map here instead of an array?

harrysolovay: It's fast, and ensures no duplicates

capi/src/write/sitemap.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+  }
+
+  const src = `<?xml version="1.0" encoding="utf-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+  ${[...routes.keys()].map(createSitemapRoute).join("\n")}
+</urlset>
+`;
+
+  const destPathPieces = config.outDir.split(path.sep);
```
jordanranz, line 35: If something in the route assignment broke it seems we would just generate an empty sitemap file. Should we throw some kind of error instead?

harrysolovay: Maybe I don't understand correctly...

destPathPieces being of length 0 won't happen with the Amplify Console environment... and if anyone is running Capi locally, and they have the project at their filesystem root (before library, before user)... then they're begging for trouble.

capi/src/write/sitemap.ts
```diff
@@ -0,0 +1,44 @@
...
+    if (page.versions) {
+      Object.entries(page.versions).forEach(([, version]) =>
+        routes.set(version, true),
+      );
+    }
+  }
+
+  if (routes.size > 0) {
+    const src = `<?xml version="1.0" encoding="utf-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
```
jordanranz, line 31: Can we fix the formatting here

capi/src/write/sitemap.ts
```diff
@@ -0,0 +1,44 @@
...
+  ${[...routes.keys()].map(createSitemapRoute).join("\n")}
+</urlset>
+`;
+
+    const destPathPieces = config.outDir.split(path.sep);
```
jordanranz, line 36: If `outDir` is undefined this will break

capi/src/write/sitemap.ts
```diff
@@ -0,0 +1,44 @@
...
+    destPathPieces.pop();
```
jordanranz, line 37: what is this `pop` doing?

capi/src/write/sitemap.ts
```diff
@@ -0,0 +1,44 @@
...
+    routes.set(page.route, true);
+    if (page.versions) {
+      Object.entries(page.versions).forEach(([, version]) =>
+        routes.set(version, true),
+      );
+    }
+  }
+
+  if (routes.size > 0) {
+    const src = `<?xml version="1.0" encoding="utf-8"?>
```
jordanranz, line 30: Let's use uppercase UTF-8

capi/src/write/sitemap.ts
```diff
@@ -0,0 +1,44 @@
+import * as t from "../types";
+import * as path from "path";
+import * as fs from "fs-extra";
+
+const now = new Date();
+const yyyyMmDd = `${now.getFullYear()}-${now.getMonth()}-${now.getDate()}`;
+
+function createSitemapRoute(route: string): string {
+  return ` <url>
+    <loc>http://docs.amplify.aws${route}</loc>
```
jordanranz, line 10: This should be `https`

capi/src/write/sitemap.ts
harrysolovay: `outDir` cannot be undefined

capi/src/write/sitemap.ts
harrysolovay: Moved this into an optional util in the config

capi/src/write/sitemap.ts
harrysolovay: Fixed

capi/src/write/sitemap.ts
harrysolovay: Moved this into config

capi/src/write/sitemap.ts
harrysolovay: Fixed

capi/src/write/sitemap.ts
harrysolovay: Done

tasks/build.ts
```diff
@@ -43,4 +43,9 @@ c.API({
   outDir: "client/src/api",
   publicDir: "client/src/assets",
   watch,
+  getSitemapDestDir(config) {
```
jordanranz, line 46: why not just `appDir: "client/src"`

harrysolovay: My thinking is it should be based on the configured dest path.

PR #1649, opened 2020-04-27
feat: implement code-block-switcher component

harrysolovay:
When wanting to group code blocks in markdown, we can now do something like the following:

````md
<amplify-code-block-switcher tab-heading-list="JavaScript,TypeScript,Rust">

```js
const a = "a";
```

```ts
const a: "a" = "a";
```

```rust
let mut a = String::from("a");
```

</amplify-code-block-switcher>
````

### CODE REVIEW

File: client/src/amplify-ui/code-block-switcher/code-block-switcher.style.ts [hunk 1]
```diff
@@ -0,0 +1,41 @@
...
+    margin-top: 0;
+  }
+`;
+
+export const activeTabStyle = css`
+  font-weight: 700;
+`;
+
+export const tabStyle = css`
+  color: var(--color-white);
```
jordanranz, line 23: Remember to use variables here and below

client/src/amplify-ui/code-block-switcher/code-block.spec.ts
```diff
@@ -0,0 +1,25 @@
+import {AmplifyCodeBlockSwitcher} from "./code-block-switcher";
```
jordanranz, line 1: Code block and code block switch tests should be different

docs/lib/auth/overview.md
```diff
@@ -3,6 +3,18 @@ title: Concepts
 description: Learn more about the foundational auth concepts for cloud-based application and how they work with Amplify Framework.
 ---
 
+<amplify-code-block-switcher tab-heading-list="JavaScript, TypeScript">
```
jordanranz, line 6: Can we infer the `tab-heading-list` from the order and type of the code block?

client/src/amplify-ui/code-block-switcher/code-block-switcher.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  createActiveChildISetter = (i: number) => (_event: Event): void => {
+    this.activeChildI = i;
+    localStorage.setItem(this.tabLocalStorageKey, String(i));
+  };
+
+  render() {
+    // + 2 because CSS child indices start at 1, and the first child in the host is the div of tabs
+    const revealStyle = css`
+      > pre:nth-child(${this.activeChildI + 2}) {
```
jordanranz, line 49: How does the "first child in the host is the div of tabs" in a selector for `pre` tags?

docs/lib/auth/overview.md
harrysolovay: Not without letting the first render happen, which means a flicker & prerender pains. Considering refactoring TOC as well so that we get a single clean paint.

client/src/amplify-ui/code-block-switcher/code-block-switcher.tsx
harrysolovay: Inspect the DOM. The elements passed in via slot are rendered adjacent to the tab-containing div. As such, the `nth-child` selector needs to be passed the tab index plus 1. We add another because `nth-child` starts with 1, whereas arrays with 0 (ofc). Didn't mean to devsplain there haha.

client/src/amplify-ui/code-block-switcher/code-block.spec.ts
harrysolovay: Good catch

client/src/amplify-ui/code-block-switcher/code-block-switcher.style.ts
harrysolovay: Fixed

client/src/amplify-ui/code-block-switcher/code-block-switcher.style.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  &:first-child {
+    margin-left: 0.5rem;
+  }
+`;
+
+export const tabContainerStyle = css`
+  border-top-right-radius: 0.25rem;
+  border-top-left-radius: 0.25rem;
+  background-color: var(--color-ink-md);
```
jordanranz, line 40: this color needs an update

client/src/amplify-ui/code-block-switcher/code-block-switcher.tsx
jordanranz: Ok, we should use `pre:nth-of-type(${this.activeChildI})` and set the initial value to 1. Unless there is a good reason to not follow the CSS child indices order

client/src/amplify-ui/code-block-switcher/code-block-switcher.tsx
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }
+
+  createActiveChildISetter = (i: number) => (_event: Event): void => {
+    this.activeChildI = i;
+    localStorage.setItem(this.tabLocalStorageKey, String(i));
+  };
+
+  render() {
+    // + 2 because CSS child indices start at 1, and the first child in the host is the div of tabs
```
jordanranz, line 47: This comment can be updated

PR #1670, opened 2020-04-30
Suppress line numbers for console blocks

jpignata:
*Issue #, if available:* n/a

*Description of changes:* Suppress line numbers for console blocks

Line numbers aren't particularly meaningful for console output blocks, so this change suppresses them. As a result, it renames `oneLineStyle` to `noLineNumbersStyle`.

- Extract "console" to constant
- Make lineNumbers property a number (instead of parsing it later), and
  not optional
- Make language not optional

**Before**

**After**

### DIFF
--- a/client/src/amplify-ui/code-block/__snapshots__/code-block.spec.ts.snap
+++ b/client/src/amplify-ui/code-block/__snapshots__/code-block.spec.ts.snap
@@ -8,8 +8,38 @@ exports[`amplify-code-block Copy button is not visible if the language is set to
 `;
 
 exports[`amplify-code-block Copy button is visible for languages other than console 1`] = `
-<amplify-code-block class="css-1lj8awz css-75wbz6" language="javascript">
+<amplify-code-block class="css-1lj8awz" language="javascript">
   <!---->
+  const lol = 'i dunno'; alert(lol);
+  <div class="css-khgmir"></div>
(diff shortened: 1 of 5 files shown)


capi/src/init-node/index.ts
```diff
@@ -60,9 +60,9 @@ const highlight = (code: string, language: string): string => {
     languageIsSet ? `-${language}` : ""
   }">${highlighted}</div>`;
 
-  return `<amplify-code-block language="${language}" line-count="${String(
```
harrysolovay, line 63: Although JS is famous for implicit type conversions, strings+numbers concatenation doesn't feel very deliberate. Is there a reason that you removed the explicit casting?

jpignata: Hmm, I'm not sure I have a great answer for you! I wanted to pass in a number, and not a String, to save us from the `parseInt()` business, but I see what you mean in that we're returning a string here.

jpignata: I put the explicit casting back, and uh, it still works. It's.. a string template, with a number in it, that gets coerced to a string... that... is a React tag.. that.. gets replace with... a component.. that... and then... uhhh.. takes a number... that was a string... uhh..

Help I'm on the tippy top of a teetering tower of complexity and I have no idea what I'm doing.

Let me know if this is better?

PR #1708, opened 2020-05-05
Update escapehatch.md

wooj2:
Per discussion during sprint planning

### DIFF
--- a/docs/lib/predictions/fragments/ios/escapehatch.md
+++ b/docs/lib/predictions/fragments/ios/escapehatch.md
@@ -1,4 +1,4 @@
-If your application requires calling any of the AWS Services behind the Amplify Predictions category in a manner that we aren't doing on your behalf, we provide an escape hatch so you can get a reference to that service.  For example, to get a reference to `AWSRekognition`:
+If you need functionality in the AWS services used by the Amplify Predictions category that isn't available, we provide an escape hatch so you can get a reference to that service.  For example, to get a reference to `AWSRekognition`:
 
 ```swift
 guard let predictionsPlugin = try Amplify.Predictions.getPlugin(for: "awsPredictionsPlugin") as? AWSPredictionsPlugin else {
@@ -44,3 +44,11 @@ guard let textractService = predictionsPlugin.getEscapeHatch(key: .textract) as?
     return
 }
 ```
+
+**API Documentation Resources**
+* [Amazon Rekognition API Reference](https://docs.aws.amazon.com/rekognition/latest/dg/API_Reference.html)
+* [Amazon Translate API Reference](https://docs.aws.amazon.com/translate/latest/dg/API_Reference.html)
+* [Amazon Polly API Reference](https://docs.aws.amazon.com/polly/latest/dg/API_Reference.html)
+* [Amazon Transcribe API Reference](https://docs.aws.amazon.com/transcribe/latest/dg/API_Reference.html)
+* [Amazon Comprehend API Reference](https://docs.aws.amazon.com/comprehend/latest/dg/API_Reference.html)
+* [Amazon Textract API Reference](https://docs.aws.amazon.com/textract/latest/dg/API_Reference.html)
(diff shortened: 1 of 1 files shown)


docs/lib/predictions/fragments/ios/escapehatch.md
```diff
@@ -1,12 +1,46 @@
-For any of the AWS services behind predictions, you can use the SDK object to get access to any methods we are not calling on your behalf by using the Escape Hatch like so:
+If your application requires calling any of the AWS Services behind the Amplify Predictions category in a manner that we aren't doing on your behalf, we provide an escape hatch so you can get a reference to that service.  For example, to get a reference to `AWSRekognition`:
```
jpignata, line 1: Some nits:

- Services doesn't need to be capitalized, right?
- "in a manner that we aren't doing on your behalf" is a little confusing. Can we say something more direct?

docs/lib/predictions/fragments/ios/escapehatch.md
```diff
@@ -1,12 +1,46 @@
...
+
+guard let rekognitionService = predictionsPlugin.getEscapeHatch(key: .rekognition) as? AWSRekognition else {
+    print("Unable to get AWSRekognition")
+    return
+}
 
 let request = AWSRekognitionCreateCollectionRequest()
 if let request = request {
     rekognitionService.createCollection(request)
 }
```
jpignata, line 17: I think this is a useful example of why one would want to use this escape hatch! Any chance we can link to a place a developer could read up on `AWSRekognition` and friends?

docs/lib/predictions/fragments/ios/escapehatch.md
jpignata: I'm trying to come up with an alternate phrasing for the above and having an awful time. Here's a try:

> If you need functionality in the AWS services used by the Amplify Prediction category that isn't available, we provide...

Not sure if better.

docs/lib/predictions/fragments/ios/escapehatch.md
```diff
@@ -1,12 +1,46 @@
...
 ```
+
+In addition to `AWSRekognition`, this same pattern can be used to get access to `AWSTranslate`, `AWSPolly`, `AWSTranscribeStreaming`, `AWSComprehend`, and `AWSTextract`.  For example:
+```swift
+guard let translateService = predictionsPlugin.getEscapeHatch(key: .translate) as? AWSTranslate else {
+    print("Unable to get AWSTranslate")
+    return
+}
```
jpignata, line 25: Given these are identical, do you think a table would be a better bet here? e.g.:

| Escape Hatch Key |  Class |
| --- | --- |
| `textract` | `AWSTextract` |

docs/lib/predictions/fragments/ios/escapehatch.md
jpignata: *narrator voice* It wasn't.

docs/lib/predictions/fragments/ios/escapehatch.md
wooj2: I wrestled with this myself while updating the documentation.  On one hand, I think executable, non abstract code is clearer, and more obvious for developers.  They can even copy and paste the code, and just "have it work".  On the other hand, wading through duplicate code might be annoying.

I landed on just writing out the code because it might not be obvious where to use "Escape Hatch Key" in code, and explaining to substitute the escape hatch key and class in a code example like:
```
guard let xxxService = predictionsPlugin.getEscapeHatch(key: <ESCAPE_HATCH_KEY>) as? <AWS CLASS> else {
    print("Unable to get <AWS CLASS>")
    return
}
```
Seems harder to follow because it's abstract.  Maybe there's a better way to explain it?  thoughts?

docs/lib/predictions/fragments/ios/escapehatch.md
wooj2: Added a section at the bottom of the page.

docs/lib/predictions/fragments/ios/escapehatch.md
jpignata: What you think is best. It was more of an idle thought.

wooj2: Lets leave it for now.

PR #1712, opened 2020-05-05
Update customui.md

Arc-dev:
*Issue #, if available:*
Lack of documentation about customising attributes.

*Description of changes:*
Added the signUpConfig that I used in the old docs + snippet of code with an example of removing the default Phone Number attribute and adding address.

Tested locally with aws-amplify-react version 4.1.9 +signed up new Cognito user.

```
npm show aws-amplify-react version
4.1.9
```


jordanranz:
We will soon be swapping out this content for the new UI components and want to recommend using the `AmplifyAuthenticator` for customizing the Authenticator experience. There is also documentation about this in the old library [here](https://docs.amplify.aws/ui-legacy/auth/authenticator/q/framework/react#sign-upin-with-emailphone-number). We do appreciate this PR and the contribution.

PR #1721, opened 2020-05-07
Removed extraneous 'a'

danaugrs:
Removed an extraneous 'a'.


brene:
@danaugrs - thanks for the contribution! Good catch on the typo!

PR #1726, opened 2020-05-07
Added styling info for amplify-authenticator to show how to update la…

dabit3:
Updated the `Theming` section of docs to show how to apply layout to the `amplify-authenticator` component.

### DIFF
--- a/docs/ui/customization/fragments/web/theming.md
+++ b/docs/ui/customization/fragments/web/theming.md
@@ -1,3 +1,18 @@
+## Styling the amplify-authenticator
+
+You can directly style the `amplify-authenticator` component using CSS. Let's say that you wanted to update the layout by centering the component. You could use the following styling to achieve this.
+
+```css
+amplify-authenticator {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  min-height: 100vh;
+}
+```
+
+## Theming the UI Components
+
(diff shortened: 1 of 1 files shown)


jordanranz:
Awesome, this is great. I think I'll move this to the `amplify-authenticator` docs once I consolidate the web frameworks into a single file.

docs/ui/customization/fragments/web/theming.md
```diff
@@ -1,3 +1,18 @@
+## Styling the amplify-authenticator
+
+You can directly style the `amplify-authenticator` component using CSS. Let's say that you wanted to update the layout by centering the component. You could use the following styling to achieve this.
+
+```css
+amplify-authenticator {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  min-height: 100vh;
```
jordanranz, line 10: This line may not be the correct thing to do as it will set the entire component's height to 100vh. What we want is the containing app to be `100vh`.

Can we just add a note to make sure the app container in `100vh`?

dabit3:
This actually isn't going to work after testing.

docs/ui/customization/fragments/web/theming.md
dabit3: This works, but also ends up applying all styling to children so is not a good solution.

jordanranz:
We should probably move this to the `amplify-authenticator` specific docs

PR #1728, opened 2020-05-07
fix: include filter in lvl0, not lvl1

harrysolovay:
lvl1 is no longer accounted for in search results, which means we add the given filter label to the lvl0.


client/src/utils/transform-search-data.ts
```diff
@@ -51,12 +51,14 @@ export function transformData(items: Item[]): Item[] {
...
+          const newHeading = `${item.hierarchy.lvl0} (${label})`;
+          item._highlightResult.hierarchy.lvl0.value = newHeading;
         }
       }
     }
     return item;
   });
 }
+
+// Getting Started
```
jordanranz, line 64: You can remove this

PR #1732, opened 2020-05-08
Discord Links

mauerbac:
Screen Shot 2020-05-08 at 3 25 46 PM

To dos
- use lower case 
- clean-up Discord icon


harrysolovay:
I prefer uppercase text and no icon. Besides swapping out the link, I'm not sure this is the right path forward. That being said, I'm often in my own world when it comes to UI sensibilities haha.

@jordanranz what do you think?

brene:
@harrysolovay  - I actually prefer the normal case font. It's consistent with the rest of the page. Only uppercase for chat seems to be off :D

Issue #1740, opened 2020-05-10
[Syntax issues] REST API docs


ffxsam:
https://docs.amplify.aws/lib/restapi/fetch/q/platform/js

Two issues I wanted to point out:

```js
async getData() { 
  const apiName = 'MyApiName';
  const path = '/path';
  const myInit = { // OPTIONAL
    headers: {}, // OPTIONAL
  };

  return await API.get(apiName, path, myInit);
}

getData();
```

Besides the fact it's missing `function` on line 1, and `await` is missing on the actual function call, this should code example instead be:

```js
function getData() { 
  const apiName = 'MyApiName';
  const path = '/path';
  const myInit = { // OPTIONAL
    headers: {}, // OPTIONAL
  };

  return API.get(apiName, path, myInit);
}

await getData();
```

`return await` is redundant inside an `async` function, in fact, it's an [ESLint error/warning](https://eslint.org/docs/rules/no-return-await). The function is already returning a Promise implicitly when it's declared with `async`, so it's sufficient to just return the function that returns a Promise.

Same issue with the second code snippet on the page:

```js
async function head() { 
    const apiName = 'MyApiName';
    const path = '/path';
    const myInit = { // OPTIONAL
        headers: {}, // OPTIONAL
    };

    return await API.head(apiName, path, myInit);
}

head();
```

brene:
cc @sammartinez

mtliendo:
#1776

sammartinez:
Resolving as the Pull Request was merged

PR #1741, opened 2020-05-10
Syntax fixes and code improvements

ffxsam:
`return await` is redundant inside an `async` function, in fact, it's an [ESLint error/warning](https://eslint.org/docs/rules/no-return-await). The function is already returning a Promise implicitly when it's declared with `async`, so it's sufficient to just return the function that returns a Promise.

*Issue #, if available:* #1740 

*Description of changes:*


docs/lib/restapi/fragments/js/fetch.md
```diff
@@ -70,17 +70,17 @@ API
...
     const myInit = { // OPTIONAL
         headers: {}, // OPTIONAL
     };
 
-    return await API.head(apiName, path, myInit);
+    return API.head(apiName, path, myInit);
 }
 
-head();
+const response = await head();
```
harrysolovay, line 83: Top-level await is unsupported in most environments. Otherwise, good fixes! Surprised these were in here to begin with :)

ffxsam: Sure, I was just assuming that call would be inside another function. (the docs could be modified to show that, maybe)

docs/lib/restapi/fragments/js/fetch.md
```diff
@@ -26,17 +26,17 @@ API
...
-async getData() { 
+function getData() { 
   const apiName = 'MyApiName';
   const path = '/path';
   const myInit = { // OPTIONAL
     headers: {}, // OPTIONAL
   };
 
-  return await API.get(apiName, path, myInit);
+  return API.get(apiName, path, myInit);
```
sammartinez, line 36: @ffxsam What do you think of the suggestion? Part of it is that this is an example of async/await so I don't want to remove the code of the async/await callouts if we want to show an example

```suggestion
  await API.get(apiName, path, myInit);
  return;
```

docs/lib/restapi/fragments/js/fetch.md
```diff
@@ -26,17 +26,17 @@ API
 Example with async/await
 
 ```javascript
-async getData() { 
+function getData() {
```
sammartinez, line 29:
```suggestion
async function getData() { 
```
callout below

docs/lib/restapi/fragments/js/fetch.md
```diff
@@ -26,17 +26,17 @@ API
...
   const myInit = { // OPTIONAL
     headers: {}, // OPTIONAL
   };
 
-  return await API.get(apiName, path, myInit);
+  return API.get(apiName, path, myInit);
 }
 
-getData();
+const response = await getData();
```
sammartinez, line 39:
```suggestion
getData();
```

docs/lib/restapi/fragments/js/fetch.md
```diff
@@ -70,17 +70,17 @@ API
 Example with async/await:
 
 ```javascript
-async function head() { 
+function head() {
```
sammartinez, line 73:
```suggestion
async function head() { 
```

docs/lib/restapi/fragments/js/fetch.md
```diff
@@ -70,17 +70,17 @@ API
...
-async function head() { 
+function head() { 
     const apiName = 'MyApiName';
     const path = '/path';
     const myInit = { // OPTIONAL
         headers: {}, // OPTIONAL
     };
 
-    return await API.head(apiName, path, myInit);
+    return API.head(apiName, path, myInit);
```
sammartinez, line 80:
```suggestion
    await API.head(apiName, path, myInit);
    return;
```

docs/lib/restapi/fragments/js/fetch.md
sammartinez, line 83:
```suggestion
head();
```
base on above feedback

docs/lib/restapi/fragments/js/fetch.md
ffxsam: In this suggested change, you're just making a call to `API.get()` and discarding the results, returning nothing. Shouldn't a `getData()` function return the data itself?

docs/lib/restapi/fragments/js/fetch.md
ffxsam: I like this better:

```js
(async function() {
  await getData();
})();
```

Some developers who are not as advanced may be misled by the code example and just call `getData()`, when it really needs an `await` in front of it since it's returning a Promise. If the IIFE above is potentially confusing, then maybe just:

```js
async function main() {
  await getData();
}
```

These comments apply to the rest of your comments in this PR. I'm still of the opinion that the `head()` and `getData()` functions should return the result, per my initial suggestion.

sammartinez: Im good with this as well!

docs/lib/restapi/fragments/js/fetch.md
harrysolovay: Definitely better than wrapping in an anonymous function. Less is more. The writing surrounding this block can describe `head` as returning a response promise. The code sample should stay light.

sammartinez:
> hey @sammartinez - where did we leave off with this?

Was waiting for @ffxsam to come back with the IIFE change and once that was done, we will be good to get it in. @ffxsam Is this something you can get in this week?

ffxsam:
@sammartinez IIFEs added.

Any other changes that I personally think are correct are in. Just as a reminder to anyone else reading this and wondering about it: functions that return a Promise [don't need to be declared as async](https://eslint.org/docs/rules/no-return-await). That's implied by the return of a Promise.

Issue #1379, opened 2020-03-30
[v2-feedback] Incorrect Angular hosting guide


nhanna-suplari:
**Page**: [`https://docs.amplify.aws/start/getting-started/hosting?integration=angular`](https://docs.amplify.aws/start/getting-started/hosting?integration=angular)

**Feedback**: Should probably change the default for angular projects for the DistributionDir config. The default is incorrect for an angular 8 app, which namespaces the build output by the app name. Also, the error you get when you make that mistake (a 403 after publishing the app) is hard to trace back to the actual problem if you are using the amplify hosting solution, so would probably be good to add a warning during the publish step in the cli as well.

jnathanh:
@sammartinez looks like this just got marked stale. It also looks like @nachbar 's fix would work.

sammartinez:
resolving as the above PR has been merged

PR #1744, opened 2020-05-11
Fix Angular hosting DistributionDir

nachbar:
*Issue #, if available:*
[v2-feedback] Incorrect Angular hosting guide #1379

*Description of changes:*
Newer versions of Angular namespace the build output by the app name, but the default selected by amplify init does not.  As a result, hosting fails with an AccessDenied XML error document shown in the browser.

This PR provides instructions  to prevent the problem from the beginning, and also adds instructions at the point where the error will manifest (on a page that is not unique to Angular) to lead the frustrated student to the solution.

Also, the project name in the sample angular init output referenced the wrong project name, so that was fixed as well.


brene:
cc @dabit3 @jordanranz

sammartinez:
Thanks for contributing!

PR #1762, opened 2020-05-13
Generate Filter-specific Page Assets + Filters Without Fragments

harrysolovay:
* make content filterable without fragments
* generate filtered-page specific assets (so we need to load less data for every filtered page + friendlier to the prerendering process)
* update `capi` usage in client
* update version of `@stencil/core`


jordanranz:
Just add some more comments like we discussed, but otherwise LGTM

PR #1767, opened 2020-05-14
Android predictions doc

raphkim:
*Issue #, if available:*

*Description of changes:*
Add documentation for Android Predictions category


docs/lib/predictions/fragments/android/escapehatch.md
```diff
@@ -0,0 +1,47 @@
+If you need functionality in the AWS services used by the Amplify Predictions category that isn't available, we provide an escape hatch so you can get a reference to that service. For example, to get a reference to `AmazonRekognitionClient`:
+
+```java
+// Obtain reference to the plugin
+final AWSPredictionsPlugin predictionsPlugin;
+try {
+    predictionsPlugin = (AWSPredictionsPlugin) Amplify.Predictions.getPlugin("awsPredictionsPlugin");
+} catch (ClassCastException error) {
+    Log.e("PredictionsQuickstart", "Unable to cast to AWSPredictionsPlugin");
```
jpignata, line 9: Does this expect the error, too?

docs/lib/predictions/fragments/android/escapehatch.md
```diff
@@ -0,0 +1,47 @@
+If you need functionality in the AWS services used by the Amplify Predictions category that isn't available, we provide an escape hatch so you can get a reference to that service. For example, to get a reference to `AmazonRekognitionClient`:
+
+```java
+// Obtain reference to the plugin
+final AWSPredictionsPlugin predictionsPlugin;
```
jpignata, line 5: Hmm, instead of setting the `AWSPredictionsPlugin` externally, why not declare the `AWSPredictionsEscapeHatch` and set it in the try/catch?

docs/lib/predictions/fragments/android/escapehatch.md
```diff
@@ -0,0 +1,47 @@
...
+// Send a new request to the Rekognition endpoint directly using the client
+AmazonRekognitionClient client = escapeHatch.getRekognitionClient();
+CreateCollectionRequest request = new CreateCollectionRequest()
+        .withCollectionId("<new-collection-id-here>");
+client.createCollection(request);
+```
+
+In addition to `AmazonRekognitionClient`, this same pattern can be used to get access to `AmazonTranslateClient`, `AmazonPollyClient`, `AmazonComprehendClient`, and `AmazonTextractClient`. For example:
+
+```java
```
jpignata, line 23: This is largely just what's above, right?

docs/lib/predictions/fragments/android/escapehatch.md
```diff
@@ -0,0 +1,47 @@
...
+}
+AWSPredictionsEscapeHatch escapeHatch = plugin.getEscapeHatch();
+
+// Obtain a reference to each of the supported Amazon service clients
+AmazonRekognitionClient rekognitionClient = escapeHatch.getRekognitionClient();
+AmazonTranslateClient translateClient = escapeHatch.getTranslateClient();
+AmazonPollyClient pollyClient = escapeHatch.getPollyClient();
+AmazonComprehendClient comprehendClient = escapeHatch.getComprehendClient();
+AmazonTextractClient textractClient = escapeHatch.getTextractClient();
+```
```
jpignata, line 40: Can we tie this to the below? With this client, you can then... some bridging sentence.

docs/lib/predictions/fragments/android/getting-started.md
```diff
@@ -0,0 +1,171 @@
+## Prerequisites
+
+* [Install and configure the Amplify CLI](~/cli/start/install.md)
+
+```bash
+npm install -g @aws-amplify/cli
+amplify configure
+```
+
+## Automated Configuration
```
jpignata, line 10: We'll have a Project Setup guide that will make this redundant shortly.

docs/lib/predictions/fragments/android/getting-started.md
```diff
@@ -0,0 +1,171 @@
...
+
+When your backend is successfully updated, verify that `amplifyconfiguration.json` and `awsconfiguration.json` are properly updated inside the `res/raw` directory. Your new configuration file `amplifyconfiguration.json` will contain your default project `region` value as well as any necessary configuration options for each predictions method (i.e. target language and source language for translate, etc).
+
+
+## Manual Configuration
+
+If you have already created the resources in the cloud and would like to take advantage of those existing resources but still use the Amplify library in Android Studio, please follow the directions below:
+
+Create the file `amplifyconfiguration.json`:
+```
```
jpignata, line 49: `bash`

docs/lib/predictions/fragments/android/getting-started.md
```diff
@@ -0,0 +1,171 @@
...
+touch amplifyconfiguration.json
+```
+
+Copy the contents over and update the values for the specific predictions method you are looking to use:
```
jpignata, line 53: Hmm, instead of touch and then a blob, why don't you just tell them explicit to open the file in Android Studio, and copy and paste the whole thing?

docs/lib/predictions/fragments/android/getting-started.md
```diff
@@ -0,0 +1,171 @@
...
+                        "region": "us-west-2",
+                        "defaultNetworkPolicy": "auto"
+                    }
+                }
+            }
+        }
+    }
+}
+```
+Add both the `amplifyconfiguration.json` and the `awsconfiguration.json` to your project inside the `res/raw` directory.
```
jpignata, line 105: I don't see `awsconfiguration.json` referenced anywhere but here.

docs/lib/predictions/fragments/android/getting-started.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Configure your application
+
+Add the following dependencies to your **app**'s' `build.gradle` file and click "Sync Now" when prompted:
+
+```gradle
+dependencies {
+    implementation 'com.amplifyframework:core:0.10.0'
```
jpignata, line 113: `1.0.0`

docs/lib/predictions/fragments/android/getting-started.md
```diff
@@ -0,0 +1,171 @@
...
+    implementation 'com.amplifyframework:aws-predictions:0.10.0'
+    implementation 'com.amazonaws:aws-android-sdk-mobile-client:2.16.+'
+}
+```
+
+Add the following compile options in the same file to support the Java 8 features that Amplify uses:
+
+```gradle
```
jpignata, line 121: `groovy`

docs/lib/predictions/fragments/android/getting-started.md
```diff
@@ -0,0 +1,171 @@
...
+
+If you have already created the resources in the cloud and would like to take advantage of those existing resources but still use the Amplify library in Android Studio, please follow the directions below:
+
+Create the file `amplifyconfiguration.json`:
+```
+touch amplifyconfiguration.json
+```
+
+Copy the contents over and update the values for the specific predictions method you are looking to use:
+```
```
jpignata, line 54: `console`

docs/lib/predictions/fragments/android/getting-started.md
```diff
@@ -0,0 +1,171 @@
...
+  compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
+}
+```
+
+## Initialize Amplify
+
+Add the following code to your `MainActivity#onCreate` method (ideally this would go in your Application class but this works for getting started quickly):
```
jpignata, line 132: Our project setup guide will have them create an `Application` class, so we can change this to alter that. Since Auth is in master, can we use that instead of `AWSMobileClient`? cc @TrekSoft

docs/lib/predictions/fragments/android/getting-started.md
```diff
@@ -0,0 +1,171 @@
...
+});
+```
+
+## Working with the API
+
+Here is an example of translating text. In order to override any choices you made in regards to target or source languages while adding this resource using the Amplify CLI, you can pass in them in directly as parameters as shown below.
+
+```java
+Amplify.Predictions.translateText(
+        "I like to eat spaghetti",
```
jpignata, line 160: ¡me gusta!

docs/lib/predictions/fragments/android/getting-started.md
```diff
@@ -0,0 +1,171 @@
...
+        LanguageType.ENGLISH,
+        LanguageType.SPANISH,
+        result -> Log.i("PredictionsQuickstart", result.getTranslatedText()),
+        error -> Log.e("PredictionsQuickstart", error.toString())
+);
+```
+
+As a result of executing this code, this you should see the following printed to your console:
```
jpignata, line 168: Should or will?

docs/lib/predictions/fragments/android/getting-started.md
```diff
@@ -0,0 +1,171 @@
...
+## Working with the API
+
+Here is an example of translating text. In order to override any choices you made in regards to target or source languages while adding this resource using the Amplify CLI, you can pass in them in directly as parameters as shown below.
+
+```java
+Amplify.Predictions.translateText(
+        "I like to eat spaghetti",
+        LanguageType.ENGLISH,
+        LanguageType.SPANISH,
+        result -> Log.i("PredictionsQuickstart", result.getTranslatedText()),
```
jpignata, line 163: For horizontal space concerns, can we use a shorter tag value throughout?

docs/lib/predictions/fragments/android/identify-celebrity.md
```diff
@@ -0,0 +1,50 @@
...
+            result -> {
+                IdentifyCelebritiesResult identifyResult = (IdentifyCelebritiesResult) result;
+                CelebrityDetails metadata = identifyResult.getCelebrities().get(0);
+                Log.i("PredictionsQuickstart", metadata.getCelebrity().getName());
+            },
+            error -> Log.e("PredictionsQuickstart", error.toString(), error)
+    );
+}
+```
+As a result of passing in an image of a celebrity, you should see the corresponding celebrity name printed to the console.
```
jpignata, line 50: Should or will?

docs/lib/predictions/fragments/android/identify-celebrity.md
```diff
@@ -0,0 +1,50 @@
...
+? Who should have access? (Use arrow keys)
+  Auth users only
+❯ Auth and Guest users
+
+```
+Run `amplify push` to create the resources in the cloud
+
+## Working with the API
+
+```java
```
jpignata, line 36: Can you add the Kotlin example here?

docs/lib/predictions/fragments/android/identify-entity.md
```diff
@@ -0,0 +1,123 @@
...
+    );
+}
+```
+As a result of passing in an image, the bounding box ([`android.graphics.RectF`](https://developer.android.com/reference/android/graphics/RectF)) that captures detected entity will be printed to the console.
+
+**Note**: Bounding boxes for entities are returned as ratios, so make sure to properly scale it before using it.
+
+### Detect pre-determined entities in an image
+
+In order to match entities from a pre-created [Amazon Rekognition Collection](https://docs.aws.amazon.com/rekognition/latest/dg/collections.html), ensure that both `collectionId` and `maxEntities` are set in your `amplifyconfiguration.json` file. The value of `collectionId` should be the name of your collection that you created either with the CLI or the SDK. The value of `maxEntities` should be a number greater than `0` or less than `51` (50 is the max number of entities Rekognition can detect from a collection). If both `collectionId` and `maxEntities` do not have valid values in the `amplifyconfiguration.json` file, then this call will just detect entities in general with facial features, landmarks, etc.
```
jpignata, line 61: Hmm. Can I see an example snippet?

docs/lib/predictions/fragments/android/identify-entity.md
```diff
@@ -0,0 +1,123 @@
...
+❯ Yes
+  No
+
+? Who should have access? (Use arrow keys)
+  Auth users only
+❯ Auth and Guest users
+```
+Run `amplify push` to create the resources in the cloud
+
+**Note**: Only one mode of entity detection can be used per app. If entity detection was already configured to default settings, then CLI will forbid you from adding another resource for identifying entities. In that case, run `amplify update predictions`, select the corresponding resource, and then re-configure it as necessary.
```
jpignata, line 106: This language is kind of intimidating (ES IST VERBOTEN!!! 😡). Can we cut to the point and rephrase? i.e., if they've already configured this, they can update their settings via the CLI (es ist erlaubt 😄).

docs/lib/predictions/fragments/android/identify-text.md
```diff
@@ -0,0 +1,65 @@
...
+? Who should have access? (Use arrow keys)
+  Auth users only
+❯ Auth and Guest users
+```
+
+## Working with the API
+
+### Detect text in an image
+
+Amplify will make calls to both Amazon Textract and Rekognition depending on the type of text you are looking to identify (i.e. image or document).
```
jpignata, line 35: Amazon Rekognition

docs/lib/predictions/fragments/android/identify-celebrity.md
```diff
@@ -0,0 +1,50 @@
+## Set up your backend
+
+If you haven't already done so, run `amplify init` inside your project and then `amplify add auth` (we recommend selecting the *default configuration*).
```
jpignata, line 3: Doesn't the CLI prompt you to do this? Shouldn't the setup guide walk through this fundamental step? @wooj2 @TrekSoft

docs/lib/predictions/fragments/android/identify-text.md
```diff
@@ -0,0 +1,65 @@
...
+    Amplify.Predictions.identify(
+            TextFormatType.PLAIN,
+            image,
+            result -> Log.i("PredictionsQuickstart", ((IdentifyTextResult) result).getFullText()),
+            error -> Log.e("PredictionsQuickstart", error.toString(), error)
+    );
+}
+```
+
+**Note**: Do *NOT* pass `IdentifyActionType.DETECT_TEXT` as identify action. Pass in an instance of `TextFormatType` instead.
```
jpignata, line 50: This is confusing to me. Why the screaming **NOT**? Is this just a helpful warning?

docs/lib/predictions/fragments/android/translate.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Override configured language
+
+In order to override any choices you made in regards to target or source languages while adding this resource using the Amplify CLI, you can pass in them in directly as parameters as shown below.
+
+```java
+public void translate(String text) {
+    Amplify.Predictions.translateText(
+            text,
+            LanguageType.ENGLISH,
```
jpignata, line 62: Is there an `AUTO` option in our 1.0?

docs/lib/predictions/fragments/android/text-speech.md
```diff
@@ -0,0 +1,72 @@
...
+    Amplify.Predictions.convertTextToSpeech(
+            text,
+            result -> playAudio(result.getAudioData()),
+            error -> Log.e("PredictionsQuickstart", error.toString(), error)
+    );
+}
+```
+As a result of running this code, you will hear audio of the text being emitted from your device.
+
+**Note**: Android API 23 added support for [`MediaDataSource`](https://developer.android.com/reference/android/media/MediaDataSource), which allows for `InputStream` from Amplify to be read directly without writing to a file.
```
jpignata, line 72: Do you have a code sample for that? You could present that in a tab.

docs/lib/predictions/fragments/android/identify-celebrity.md
wooj2: Not sure if i fully understand your question.  Are you saying if you try to run `amplify add auth` before running `amplify init`, doesn't it tell you to run amplify init?  If that is your question, then yes.  i see:
```bash
$ amplify add auth
You are not working inside a valid amplify project.
Use 'amplify init' in the root of your app directory to initialize your project with Amplify
Error: You are not working inside a valid amplify project.
Use 'amplify init' in the root of your app directory to initialize your project with Amplify
...
...
```

docs/lib/predictions/fragments/android/identify-celebrity.md
wooj2: I think should works nicely here, mostly because if they pass in a celebrity we are unaware of then they might not see a celebrity name.

docs/lib/predictions/fragments/android/identify-celebrity.md
jpignata: Ah, sorry no, I meant:

Doesn't the CLI prompt you to add auth when you add predictions if it is not yet configured?

docs/lib/predictions/fragments/android/identify-celebrity.md
jpignata: That can be phrased much more precisely:

> As a result of passing in an image, `identify` will return the name of a celebrity if detected.

docs/lib/predictions/fragments/android/identify-celebrity.md
raphkim: you are right! If `auth` isn't yet enabled while adding predictions, then CLI will prompt you to configure it also. However, I directly used the documentation from iOS page for simplicity. Should I change it?

docs/lib/predictions/fragments/android/identify-text.md
raphkim: Yes. Based on design, `IdentifyActionType` enum is normally what you pass here... However, it returns an error when you pass `IdentifyActionType#DETECT_TEXT`.

In iOS, `IdentifyAction` uses [enum within enum](https://github.com/aws-amplify/amplify-ios/blob/master/Amplify/Categories/Predictions/Models/IdentifyAction.swift) to allow customer to specify a specific type of detection (sub-sub-category of "identify" if you will). Unfortunately, Android can't really do that the way iOS is doing, so I wrote an interface to simulate it. The way I did it was:

[`IdentifyActionType`](https://github.com/aws-amplify/amplify-android/blob/master/core/src/main/java/com/amplifyframework/predictions/models/IdentifyActionType.java), [`TextFormatType`](https://github.com/aws-amplify/amplify-android/blob/master/core/src/main/java/com/amplifyframework/predictions/models/TextFormatType.java), and [`LabelType`](https://github.com/aws-amplify/amplify-android/blob/master/core/src/main/java/com/amplifyframework/predictions/models/LabelType.java) are enums that implement [`IdentifyAction`](https://github.com/aws-amplify/amplify-android/blob/master/core/src/main/java/com/amplifyframework/predictions/models/IdentifyAction.java):

Where `IdentifyAction` is required to be specified, customer must supply it with an instance of `IdentifyActionType` UNLESS that sub-category has sub-sub-categories, in which case the other type must be specified. In hindsight, I think it makes sense for it to also accept `IdentifyActionType` here and then just assume a default value for `LabelType` or `TextFormatType`... I think I'll do that actually...

docs/lib/predictions/fragments/android/translate.md
raphkim: nope not yet... It's an effort that would involve every platform to adapt (minor change though if any), and I have not yet brought up the suggestion to other teams yet.

docs/lib/predictions/fragments/android/text-speech.md
raphkim: Good idea.

docs/lib/predictions/fragments/android/escapehatch.md
raphkim: Yea I guess I got carried away while trying to follow the iOS doc. It seems intuitive enough in Android side to not have to specify this section I guess. I will remove it.

docs/lib/predictions/fragments/android/escapehatch.md
raphkim: `try-catch` should not even be necessary, but I only had it to match the error message on iOS side. This will only break if the predictions plugin was not configured yet.

docs/lib/predictions/fragments/android/getting-started.md
raphkim: Ah! you are right. It was a miss on my part while copying the instruction from iOS, but it would certainly be more helpful to instruct them to use Android Studio to directly create the files where they are required.

docs/lib/predictions/fragments/android/getting-started.md
raphkim: I didn't see it in iOS doc either, which I found somewhat confusing. I was hoping to add some template, but it seems that there are a lot of things being added to `awsconfiguration.json` for auth and I don't know the current state of auth category. Are more things going to be added to `awsconfiguration.json`? Or less? Are things going to be added to `amplifyconfiguration.json` too on top of what we already have on `awsconfiguration.json`? 

Thinking about ^ gave me a headache, so I just went with keeping silent like iOS did haha

docs/lib/predictions/fragments/android/getting-started.md
raphkim: oof you are right. It's not added in this version yet haha

docs/lib/predictions/fragments/android/getting-started.md
raphkim: Yea this portion of documentation will require an update after Auth category is finished.

docs/lib/predictions/fragments/android/getting-started.md
raphkim: "will" sounds better haha. "This code is tested and approved"

docs/lib/predictions/fragments/android/getting-started.md
raphkim: Yea I got the tag value from `"StorageQuickstart"`, which looks better than `"PredictionsQuickstart"` haha. I can remove the prefix and juts call it `"QuickStart"` maybe?

docs/lib/predictions/fragments/android/identify-celebrity.md
raphkim: I want to add kotlin example everywhere ideally haha. I'll work on that when I get time

docs/lib/predictions/fragments/android/identify-entity.md
raphkim: CLI workflow is where these values are inputed. There is an example flow right after :)

docs/lib/predictions/fragments/android/escapehatch.md
jpignata: Ah! I think that would be self-explanatory. I'd remove this for clarity.

docs/lib/predictions/fragments/android/getting-started.md
jpignata: We shouldn't need `awsconfiguration.json` at all, AFAIK.

docs/lib/predictions/fragments/android/getting-started.md
jpignata: No opinion, but we should be consistent throughout. Up to you.

docs/lib/predictions/fragments/android/identify-entity.md
jpignata: Hmm, that's not super clear from the above. It says "ensure that both" -- as a reader I'm wondering what I need to do.

docs/lib/predictions/fragments/android/identify-celebrity.md
jpignata: Up to you and @wooj2

docs/lib/predictions/fragments/android/identify-text.md
jpignata: Okay, but I guess as a reader.. tell me what I should do instead of leading with what I shouldn't do.

docs/lib/predictions/fragments/android/getting-started.md
raphkim: will keep until it IS redundant

docs/lib/predictions/fragments/android/getting-started.md
raphkim: Renamed to `"AmplifyQuickstart"` to save 4 characters

docs/lib/predictions/fragments/android/identify-entity.md
raphkim: fixed!

docs/lib/predictions/fragments/android/identify-text.md
raphkim: fixed

docs/lib/predictions/fragments/android/identify-text.md
raphkim: removed warning since the code snippet already tells the user to use the correct enum. In the future I'll add the code changes to make it not breaking to begin with.

docs/lib/predictions/fragments/android/text-speech.md
raphkim: On second thought, it actually turned out to be non-trivial (or at least not trivial enough to write the code for it in the docs). I think the right way to do it is include our own implementation of `MediaDataSource` in the source code, and point them to it in the doc when it's ready.

docs/lib/predictions/fragments/android/getting-started.md
raphkim: removed manual configuration portion

docs/lib/predictions/fragments/android/getting-started.md
raphkim: removed reference to awsconfiguration

docs/lib/predictions/fragments/android/getting-started.md
raphkim: removed manual configuration portion

docs/lib/predictions/fragments/android/identify-entity.md
raphkim: re-worded a lot of these to not make reference to amplifyconfiguration

docs/lib/predictions/fragments/android/identify-celebrity.md
raphkim: leaving it as is for now, since both are technically valid ways to add auth. I think it makes sense to add it at the same time as predictions, but this "Set up your backend" portion is present in every page...

raphkim:
Closing due to wrong target branch

PR #1798, opened 2020-05-21
feat: Added auth document structure

royjit:
*Issue #, if available:*

*Description of changes:*


client/src/sitemap.xml
```diff
@@ -301,187 +301,187 @@
     <priority>.5</priority>
   </url>
   <url>
-    <loc>https://docs.amplify.aws/lib/auth/custom-auth-flow</loc>
```
wooj2, line 304: In the most recent version of the `next` branch, we have deleted this file, and it is no longer being tracked by git.. before pushing in we should just remove this from the commit.

docs/lib/auth/working-with-api.md
```diff
@@ -1,7 +1,14 @@
 ---
-title: Working with API
+title: Working with other Category
```
TrekSoft, line 2: This should be other Categories but I also don't think this matches the content you have below. Working with other categories is just as simple as making sure Auth is added and the user is signed in. Maybe this section should be removed and this topic covered under the Escape Hatch section. Either that or it should be renamed to something like, "Getting credentials" where we explain that this is not something the user should need to worry about or do at all within Amplify but if they really want to get them, they can use the escape hatch method by casting fetchAuthSession.

docs/lib/auth/signOut.md
```diff
@@ -0,0 +1,6 @@
+---
+title: SignOut
```
TrekSoft, line 2: I think "Sign Out" is more proper - same for below

docs/lib/auth/user-attributes.md
```diff
@@ -0,0 +1,7 @@
+---
+title: User Attributes
```
TrekSoft, line 2: We should be consistent in capitalization - either every significant word starts with a capital as in here or only the first word does as in "Register a user"

docs/lib/auth/menu.json
```diff
@@ -11,14 +11,14 @@
...
-		"hosted-ui",
-		"custom-auth-flow",
-		"device-features"
+    
+    "register-user",
+    "authenticate-user",
+    "auth-events",
+    "working-with-api",
+    "user-attributes",
+    "device-features",
```
TrekSoft, line 20: I think "Remember device" would be clearer than "Device features"

docs/lib/auth/user-attributes.md
```diff
@@ -0,0 +1,7 @@
...
+description: Update user attribtues
+---
+
+<inline-fragment platform="ios" src="~/sdk/auth/fragments/ios/device-features.md"></inline-fragment>
```
TrekSoft, line 6: Device features shouldn't go under User Attributes :D

docs/lib/auth/menu.json
TrekSoft: Hmm, would it make sense to have a section for "Password Management" where we cover updating it, resetting it, confirming the reset?

docs/lib/auth/auth-events.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Auth Events
+description: Listen to various auth events
+---
+
+AWS Cognito Auth Plugin sends important events through the Amplify Hub.
+
```
wooj2, line 7: This needs to be removed, otherwise, JS will show a section called "Auth events"

docs/lib/auth/working-with-api.md
```diff
@@ -1,7 +1,14 @@
...
\ No newline at end of file
+How to use auth category with other categories
+
+## UnAuth access
+
+## Get aws credentials
+
+## Get identity id
+
+## Get cognito tokens
```
wooj2: This needs to be removed, otherwise, JS will show a section called "Working with other category"

docs/lib/auth/signOut.md
```diff
@@ -0,0 +1,6 @@
+---
+title: SignOut 
+description: SignOut a user
+---
+
+How to signOut a user and clear tokens
```
wooj2, line 6: Needs to be removed otherwise will show up as "SignOut" under JS

docs/lib/auth/escapehatch.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Escape Hatch
+description: Underlying service
+---
+
+Access the escape hatch for Auth category
```
wooj2, line 6: This needs to be removed, otherwise, JS will show a section called "Escape Hatch"

PR #1800, opened 2020-05-21
Unify Predictions category for both iOS and Android

wooj2:
* Updated structure to start with getting started (unified between ios and android)
* Moved detect celebrities under entities (this seems to be what JS did)
* Removed overview page, but moved its contents to pages where it made sense.
  * First paragraph from the overview page was moved to the getting started section
  * Descriptions of individual use cases were moved to those specific pages as a lead-in to the page.
* I could have unified each page, but refrained from doing so because i think that would have been a lot more work
* Note that each page says something like "if you haven't run amplify init yet..." -- I think we should change this, but this PR has become too big already, so I'm leaving that for now.


docs/lib/predictions/fragments/android/overview.md
```diff
@@ -0,0 +1 @@
+this section will have some concepts
```
wooj2, line 1: need to delete this file.

docs/lib/predictions/fragments/ios/getting-started/10_coreml.md
```diff
@@ -0,0 +1 @@
+Specifically on iOS, we leverage Apple’s Core ML [Vision Framework](https://developer.apple.com/documentation/vision) to support offline use cases, and augment results returned from Amazon Services for a couple of our use cases.  For more information, see each individual use case.
```
wooj2, line 1: Need to update this to include Natural langauge.

docs/lib/predictions/fragments/ios/getting-started/20_preReq.md
```diff
@@ -0,0 +1,2 @@
+* An iOS application targeting at least iOS 13.0 with Amplify libraries integrated
```
jpignata, line 1: Should we tell people what happens on iOS <13? Say, what happens?

docs/lib/predictions/fragments/ios/identify-entity.md
```diff
@@ -70,3 +74,27 @@ func detectEntities(_ image: URL) {
...
+      print(result)
+    case .failed(let error):
+      print(error)
+    default:
+      print("")
+    }
+  })
+}
+```
+As a result of passing in a URL of an image of a celebrity, you should see the corresponding celebrity name printed to the screen along with additional metadata.
```
jpignata, line 100: should or will? *audience canned laughter* I kind of hate seeing "should" as a reader. It's a sign of weakness on the part of the writer. We can be precise and say "if a celebrity is detected, you will see their name in the result"

docs/lib/predictions/fragments/ios/identify-text.md
```diff
@@ -1,3 +1,7 @@
+The following APIs will allow you to identify text (words, tables, pages from a book) from an image.
```
jpignata, line 1: See my above niggle.

docs/lib/predictions/fragments/android/identify-text.md
```diff
@@ -1,3 +1,5 @@
+The following APIs will allow you to identify text (words, tables, pages from a book) from an image.
```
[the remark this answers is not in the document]
wooj2: I like calling out examples of specific cases that we support because text in an image is really vague.

Words I interpret as words that may not appear in a linear form.. perhaps you can imagine an advertisement with words spaced out in various places with various fonts

Tables I interpret to be words an numbers that often line up on a X or Y coordinate system

Pages from a book i would imagine would be mostly words that line up on vertical points on a page.

I think calling out shows flexibility/versatility of the API, but if you don't think it's adding much value we can remove it.

docs/lib/predictions/fragments/ios/getting-started/20_preReq.md
wooj2: The cocoapod spec basically says iOS 13, if they try to run it, they won't even be able to include the pod at the moment... i think it's fine to leave as is.

docs/lib/predictions/fragments/ios/identify-entity.md
wooj2: okay, i can change this to "well known celebrity, you will see" the corresponding celebrity name

PR #1802, opened 2020-05-22
 Add docs for cancel API request 

Amplifiyer:
*Issue #, if available:* aws-amplify/amplify-js#5753

*Description of changes:*
Add docs for cancel API request . See aws-amplify/amplify-js#5753 for more details

### DIFF
--- a/docs/lib/graphqlapi/fragments/js/cancel-request.md
+++ b/docs/lib/graphqlapi/fragments/js/cancel-request.md
@@ -28,7 +28,8 @@ You need to ensure that the promise returned from `API.graphql()` has not been m
 async function makeAPICall() {
     return API.graphql(graphqlOperation(...));
 }
-const promise = makeAPICall()
+const promise = makeAPICall();
+
 
 // The following will NOT cancel the request.
 API.cancel(promise, "my error message");
--- a/docs/lib/graphqlapi/fragments/js/cancel-request.md
+++ b/docs/lib/graphqlapi/fragments/js/cancel-request.md
@@ -1,0 +1,35 @@
+## CANCEL API requests
+
+You may cancel any query or mutation requests made through API category by keeping a reference to the promise returned.
+
+```javascript
+const promise = API.graphql(graphqlOperation(...));
+
+try {
+    await promise;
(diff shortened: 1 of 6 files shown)


docs/lib/graphqlapi/fragments/js/cancel-request.md
```diff
@@ -0,0 +1,35 @@
...
+API.cancel(promise, "my message for cancellation");
+```
+
+You need to ensure that the promise returned from `API.graphql()` has not been modified. Typically async functions wrap the promise being returned into another promise. For example, the following will not work
+
+```javascript
+async function makeAPICall() {
+    return API.graphql(graphqlOperation(...));
+}
+const promise = makeAPICall()
```
sammartinez, line 31:
```suggestion
const promise = makeAPICall();

```

docs/lib/restapi/fragments/js/cancel.md
```diff
@@ -0,0 +1,35 @@
...
+API.cancel(promise, "my message for cancellation");
+```
+
+You need to ensure that the promise returned from `API.get()` or any other API calls has not been modified. Typically async functions wrap the promise being returned into another promise. For example, the following will not work
+
+```javascript
+async function makeAPICall() {
+    return API.get(myApiName, myPath, myInit);
+}
+const promise = makeAPICall()
```
sammartinez, line 31:
```suggestion
const promise = makeAPICall();

```

docs/lib/restapi/fragments/js/cancel.md
```diff
@@ -0,0 +1,35 @@
+## CANCEL API requests
```
sammartinez, line 1: Do we need two of the same files to be created? We can reuse the `cancel-request` fragment within this similar documentation. Let me know what you think

Amplifiyer: I was thinking about it, right now I have two different examples `API.get` and `API.graphql` in the two files. Also was wondering if cross referencing fragments was a good idea. If you feel it's better to reuse, I can make the change.

PR #1803, opened 2020-05-22
Add missing property value

manu2504:
Missing paramater value in this.amplifyService.setAuthState() example


jordanranz:
Hey @manu2504, the documentation being changed is using object property shorthand which is valid if the key and variable names are the same: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Syntax

PR #1804, opened 2020-05-22
Changes from the latest round of feedback

jpignata:
*Issue #, if available:* n/a

*Description of changes:* Changes from the latest round of feedback

By submitting this pull request, I confirm that you can use, modify, copy, and
redistribute this contribution, under the terms of your choice.

### DIFF
--- a/docs/fragments/lib/android.md
+++ b/docs/fragments/lib/android.md
@@ -11,4 +11,3 @@ This guide shows how to build an app using our Amplify Libraries for Android and
 ### How are the Amplify Libraries for Android different from the AWS Mobile SDK for Android?
 
 The Amplify Android client libraries are use-case centric whereas the AWS Mobile SDK for Android is service-centric. This enables you to focus on your use-case more rather than figuring out the AWS service nuances. The Amplify libraries provide a highly abstracted category based programming model. You can also use the Mobile SDK with the Amplify libraries using escape hatches if the use case you are trying to build is not currently available in Amplify libraries.
-
--- a/docs/fragments/lib/android.md
+++ b/docs/fragments/lib/android.md
@@ -1,3 +1,9 @@
+<amplify-callout warning>
+
+**Amplify libraries should be used for all new cloud connected applications.** If you are currently using the AWS Mobile SDK for Android, you can access the documentation [here](~/sdk/sdk.md).
+
+</amplify-callout>
+
 ## Amplify Android
 
 This guide shows how to build an app using our Amplify Libraries for Android and the Amplify CLI toolchain.
--- a/docs/fragments/lib/ios.md
+++ b/docs/fragments/lib/ios.md
@@ -1,3 +1,10 @@
+<amplify-callout warning>
+
+**Amplify libraries should be used for all new cloud connected applications.** If you are currently using the AWS Mobile SDK for iOS, you can access the documentation [here](~/sdk/sdk.md).
+
+</amplify-callout>
+
+
 ## Amplify iOS
 
 This guide shows how to build an app using our Amplify Libraries for iOS and the Amplify CLI toolchain.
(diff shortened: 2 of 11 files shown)


docs/fragments/lib/android.md
```diff
@@ -10,5 +10,4 @@ This guide shows how to build an app using our Amplify Libraries for Android and
 
 ### How are the Amplify Libraries for Android different from the AWS Mobile SDK for Android?
 
-The Amplify Android client libraries are use-case centric whereas the AWS Mobile SDK for Android is service-centric. This enables you to focus on your use-case more rather than figuring out the AWS service nuances. The Amplify libraries provide a highly abstracted category based programming model. You can also use the Mobile SDK with the Amplify libraries using escape hatches if the use case you are trying to build is not currently available in Amplify libraries.
-
+The Amplify Android client libraries are use-case centric whereas the AWS Mobile SDK for Android is service-centric. This enables you to focus on your use-case more rather than figuring out the AWS service nuances. The Amplify libraries provide a highly abstracted category based programming model. You can also use the Mobile SDK with the Amplify libraries using escape hatches if the use case you are trying to build is not currently available in Amplify libraries.
```
wooj2, line 13: Suggestions: 
"use-case more rather" -- remove the word "more"
"rather than figuring out the AWS service nuances." -- rather than figuring out low level AWS service nuances
"You can also use the Mobile SDK" - We may choose not to rely on the mobile sdk right now -- perhaps it would be better put as: "You can access lower level service clients by using escape hatches.." (this isn't technically true for auth since we are currently using AWSMobileClient.. but... hmm i think calling it lower level service clients is more accurate)

docs/fragments/lib/ios.md
```diff
@@ -9,7 +9,5 @@ This guide shows how to build an app using our Amplify Libraries for iOS and the
 ## FAQs
 
 ### How are these different from the AWS Mobile SDK for iOS?
-The Amplify iOS client libraries are use-case centric whereas the AWS Mobile SDK for iOS is service-centric. This enables you to focus on your use-case more rather than figuring out the AWS service nuances. The Amplify libraries provide a highly abstracted category based programming model. You can also use the Mobile SDK with the Amplify libraries using escape hatches if the use case you are trying to build is not currently available in Amplify libraries.
 
-### Do I have to use the AWS Mobile SDK when using the Amplify Libraries for iOS?
-No. The AWS Mobile SDK for iOS is not a dependency to use Amplify Libraries for iOS. 
\ No newline at end of file
+The Amplify iOS client libraries are use-case centric whereas the AWS Mobile SDK for iOS is service-centric. This enables you to focus on your use-case more rather than figuring out the AWS service nuances. The Amplify libraries provide a highly abstracted category based programming model. You can also use the Mobile SDK with the Amplify libraries using escape hatches if the use case you are trying to build is not currently available in Amplify libraries.
```
wooj2, line 13: Same feedback as commented on android.

docs/lib/lib.md
```diff
@@ -7,6 +7,13 @@ filterKey: platform
 
 The Amplify open-source client libraries provide use-case centric, opinionated, declarative, and easy-to-use interfaces across different categories of cloud powered operations enabling mobile and web developers to easily interact with their backends. These libraries are powered by the AWS cloud and offer a pluggable model which can be extended to use other providers. The libraries can be used with both new backends created using the Amplify CLI and existing backend resources.
 
+<amplify-callout warning>
+
+**Amplify libraries should be used for all new cloud connected applications.** If you are currently using the AWS Mobile SDK, you can access the documentation [here](~/sdk/sdk.md).
+
+</amplify-callout>
+
+
```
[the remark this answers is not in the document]
jpignata: Fixed!

docs/fragments/lib/android.md
wooj2: per discussion offline, i suspect we will update this at a later time.

PR #1807, opened 2020-05-22
Consolidate Authenticator UI into a single file

jordanranz:
*Description of changes:*
- Consolidate Authenticator into single MD file with `docs-filter` components
- Updated the custom form field example with `usernameAlias`
- Small breakpoint change to TOC for small screens


docs/ui/auth/fragments/web/authenticator.md
```diff
@@ -0,0 +1,1251 @@
+<amplify-authenticator></amplify-authenticator>
+
+## Installation
+
+<docs-filter framework="react">
```
sammartinez, line 5: NIT/Non-Blocking: I believe as a follow up to this we can see about splitting the code snippets up into separate fragments. I feel it'll make it cleaner in the long run for maintainability

PR #1815, opened 2020-05-23
Updated @auth docs for GraphQL transform

dabit3:
- Updated truth tables & main @auth examples to be more real-world based
- Fixed a couple of mistakes in truth tables
- Fixed some basic formatting
- Added examples for multi-auth
- A few other small enhancements / changes


docs/cli/graphql-transformer/directives.md
```diff
@@ -476,60 +476,66 @@ type Post
...
 - **read**: If the record's owner is not the same as the logged in user (via `$ctx.identity.username`), throw `$util.unauthorized()` in any resolver that returns an object of this type.
-- **create**: Inject the logged in user's `$ctx.identity.username` as the *ownerField* automatically.
-- **update**: Add conditional update that checks the stored *ownerField* is the same as `$ctx.identity.username`.
-- **delete**: Add conditional update that checks the stored *ownerField* is the same as `$ctx.identity.username`.
+- **create**: Inject the logged in user's `$ctx.identity.username` as the `ownerField` automatically.
+- **update**: Add conditional update that checks the stored `ownerField` is the same as `$ctx.identity.username`.
+- **delete**: Add conditional update that checks the stored `ownerField` is the same as `$ctx.identity.username`.
 
-**Note**: When specifying operations as a part of the @auth rule, the operations not included in the list are not protected by default. For example, let's say you have the following schema:
+**Note**: When specifying operations as a part of the `@auth` rule, the operations not included in the list are not protected by default.
```
jpignata, line 488: This is an important note. Should we use `amplify-callout`?

client/src/sitemap.xml
```diff
@@ -780,6 +780,16 @@
     <changefreq>hourly</changefreq>
```
jpignata, line 780: I believe @jordanranz is fixing this, but can you nuke this file from the pull for now? It's just noise.

docs/cli/graphql-transformer/directives.md
```diff
@@ -476,60 +476,66 @@ type Post
...
-  @auth(rules: [{ allow: owner, operations: [read] }]) {
+  @auth(rules: [{ allow: owner }]) {
   id: ID!
   updatedAt: AWSDateTime!
   content: String!
 }
 ```
 
-In this schema, only the owner of the object has the authorization to perform read (getTodo and listTodos) operations on the owner created object. But this does not prevent any other owner (any user other than the creator or owner of the object) to update/delete some other owner's object.
+In this schema, only the owner of the object has the authorization to perform read (`getTodo` and `listTodos`) and update operations (`updateTodo` and `deleteTodo`) on the owner created object. This prevents any other owner (any user other than the creator of the object) to update/delete some other owner's object.
```
jpignata, line 501: Suggestion: in the second sentence, ditch the parenthetical for clarity, i.e.:

> This prevents the object from being updated or deleted by users other than the creator of the object.

docs/cli/graphql-transformer/directives.md
```diff
@@ -476,60 +476,66 @@ type Post
...
+
 Here's a truth table for the above-mentioned schema. In the table below `other` refers to any user other than the creator or owner of the object.
```
jpignata, line 503: Idle Thought: "or" is kind of a weird word here. I get it's a truth table, but I don't think that fact is pulling much weight in this sentence in terms of helping me understand. Maybe...

> Here's a table outlining which users are permitted to execute which operations. `Other` means all users, `owner` refers to the user who created the object.

Say, why don't we just use "All Users" instead of "other?"

docs/cli/graphql-transformer/directives.md
```diff
@@ -476,60 +476,66 @@ type Post
...
 
 |       | getTodo | listTodos | createTodo | updateTodo | deleteTodo |
 |:------|:-------:|:---------:|:----------:|:----------:|:----------:|
 | owner |    ✅   |     ✅    |     ✅     |      ✅   |     ✅     |
-| other |    ❌   |     ❌    |     ✅     |      ✅   |     ✅     |
+| other |    ❌   |     ❌    |     ✅     |      ❌   |     ❌     |
 
-If you want to prevent updates and deletes operations, you would need to modify the @auth rule to explicitly include the `update` and `delete` operation and your schema should look like the following:
+Next, let's say that we wanted wanted to modify the schema to allow anyone to create, read, and update, but only the owner to be able to delete.
```
jpignata, line 510: Some typos. And I think you should use the second person to be consistent:

> Next, here's a schema that allows all users to create, read, and update, but only the owner to delete:

docs/cli/graphql-transformer/directives.md
```diff
@@ -476,60 +476,66 @@ type Post
...
 
 ```graphql
 type Todo @model
-  @auth(rules: [{ allow: owner, operations: [read, update, delete] }]) {
+  @auth(rules: [{ allow: owner, operations: [create, delete] }]) {
```
jpignata, line 514: This took me three re-reads to confirm you were correct. Perhaps more prose is necessary, or maybe I'm a dummy. 🤷‍♂️

docs/cli/graphql-transformer/directives.md
```diff
@@ -476,60 +476,66 @@ type Post
...
   id: ID!
   updatedAt: AWSDateTime!
   content: String!
 }
 ```
 
+In this schema, only the owner of the object has the authorization to perform delete operations on the owner created object, but anyone can read or update them.
```
jpignata, line 521: "This is because read and update aren't specified as owner-only actions, so all users are able to perform them. Since delete is specified as an owner only action, only the object's creator can delete the object." Something, something?

docs/cli/graphql-transformer/directives.md
```diff
@@ -476,60 +476,66 @@ type Post
...
+
 Here's a truth table for the above-mentioned schema. In the table below `other` refers to any user other than the creator or owner of the object.
```
jpignata, line 523: Same comment as above.

docs/cli/graphql-transformer/directives.md
```diff
@@ -476,60 +476,66 @@ type Post
...
-|:------|:-------:|:--------:|:----------:|:----------:|:----------:|
-| owner |    ✅    |     ✅    |      ✅     |      ✅     |      ✅     |
-| other |    ❌    |     ❌    |      ✅     |      ❌     |      ❌     |
+|:------|:-------:|:---------:|:----------:|:----------:|:----------:|
+| owner |    ✅   |     ✅    |     ✅     |      ✅   |     ✅     |
+| other |    ✅   |     ✅    |     ✅     |      ✅   |     ❌     |
 
 
-> **Note:** Specifying `@auth(rules: [{ allow: owner, operations: [create]}])` still allows anyone who has access to your API to create records (as shown in the above truth table). However, including this is necessary when specifying other owner auth rules to ensure that the owner is stored with the record so it can be verified on subsequent requests.
+> **Note:** Specifying `@auth(rules: [{ allow: owner, operations: [create]}])` allows any signed in user to create records (as shown in the above truth table). Including this is necessary when specifying other owner auth rules to ensure that the owner is stored with the record so it can be verified on subsequent requests.
```
jpignata, line 531: Hmmm, I'm unclear on what this means. This particular incantation feels very important to me and maybe should be spelled out completely and explicitly.  I'm starting to think a schema and a table for each common combination would be helpful here for comprehension.

docs/cli/graphql-transformer/directives.md
```diff
@@ -550,11 +556,9 @@ type Draft @model
 }
 ```
 
-
 ### Ownership with create mutations
 
-The ownership authorization rule tries to make itself as easy as possible to use. One
-feature that helps with this is that it will automatically fill ownership fields unless
+The ownership authorization rule will automatically fill ownership fields unless
```
jpignata, line 561: 👍 to this edit.

docs/cli/graphql-transformer/directives.md
```diff
@@ -476,60 +476,66 @@ type Post
...
 
 
-> **Note:** Specifying `@auth(rules: [{ allow: owner, operations: [create]}])` still allows anyone who has access to your API to create records (as shown in the above truth table). However, including this is necessary when specifying other owner auth rules to ensure that the owner is stored with the record so it can be verified on subsequent requests.
+> **Note:** Specifying `@auth(rules: [{ allow: owner, operations: [create]}])` allows any signed in user to create records (as shown in the above truth table). Including this is necessary when specifying other owner auth rules to ensure that the owner is stored with the record so it can be verified on subsequent requests.
+
+### Multiple authorization rules
+
+You may also apply multiple ownership rules on a single `@model` type. 
+
+For example, imagine you have a type **Draft** that stores unfinished posts for a blog. You might want to allow the **Draft's owner** to create, update, delete, and read **Draft** objects. However, you might also want the **Draft's editors** to be able to update and read **Draft** objects.
```
jpignata, line 537: I'd bold just the name we use in the schema and pull the possessive out.

docs/cli/graphql-transformer/directives.md
```diff
@@ -620,6 +625,39 @@ This would return:
...
+            "id": "...",
+            "title": "A new draft",
+            "owner": "someuser@my-domain.com",
+            "editors": ["editor1@my-domain.com", "editor2@my-domain.com"]
+        }
+    }
+}
+```
+
 You can try to do the same to **owner** but this will throw an **Unauthorized** exception because you are no longer the owner of the object you are trying to create
```
jpignata, line 661: The same what? Another create?

docs/cli/graphql-transformer/directives.md
```diff
@@ -681,7 +719,10 @@ to a known set of groups. For example, you can allow all **Admin** users to crea
 update, delete, get, and list Salary objects.
 
 ```graphql
-type Salary @model @auth(rules: [{allow: groups, groups: ["Admin"]}]) {
+type Salary @model
+  @auth(rules: [
```
jpignata, line 723: This seems inconsistent with the formatting of these rules in the doc thus far.

docs/cli/graphql-transformer/directives.md
```diff
@@ -868,9 +907,37 @@ type Profile @model @auth(rules: [{allow: owner, provider: oidc, identityClaim:
 By using a configured `oidc` provider for the API, it is possible to authenticate the users against it to perform operations on the `Post` type, and `owner` authorization is also possible.
 
 
-### Combining Authorization Rules
+### Combining Multiple Authorization Rules
```
jpignata, line 910: For consistency, don't title case these (as you didn't above):

```
### Combining multiple authorization rules
```

docs/cli/graphql-transformer/directives.md
```diff
@@ -868,9 +907,37 @@ type Profile @model @auth(rules: [{allow: owner, provider: oidc, identityClaim:
...
+
+Amplify GraphQL APIs have a primary **default** Authentication type and, optionally, additional secondary authentication types. The objects and fields in the GraphQL schema can have rules with different authorization providers assigned based on the authentication types configured in your app.
+
+One of the most common scenarios for multiple authorization rules is for combining public and private access. Take for example a blog. Blogs usually offer public access for viewing posts (guest users as well as signed in users can view them) and private access for updating or deleting posts (signed in users that created the post).
```
jpignata, line 914: Suggestions for clarity:

> For example, blogs typically allow public access for viewing a post but only allow a post's creator to update or delete that post.

docs/cli/graphql-transformer/directives.md
```diff
@@ -868,9 +907,37 @@ type Profile @model @auth(rules: [{allow: owner, provider: oidc, identityClaim:
...
+
+Let's take a look at how we can combine public and private access to achive this:
```
jpignata, line 916: Typo: achieve

Suggestion:

> You can combine public and private authorization directives to achieve this:

docs/cli/graphql-transformer/directives.md
```diff
@@ -868,9 +907,37 @@ type Profile @model @auth(rules: [{allow: owner, provider: oidc, identityClaim:
...
+    ]
+  ) {
+  id: ID!
+  title: String
+  owner: String
+}
+```
 
-The objects and fields in the GraphQL schema can have rules with different authorization providers assigned.
+> The above schema assumes a default authentication type of `Amazon Cognito User Pools` and a secondary authentication type of `API key`.
```
jpignata, line 938: Probably don't need backticks for the Cognito bit or API Key, unless these tokens are inputs somewhere.

docs/cli/graphql-transformer/directives.md
```diff
@@ -886,7 +953,9 @@ type Post @model
 }
 ```
 
-In the example above the model is protected by Cognito User Pools by default and the `owner` can perform any operation on the `Post` type, but a Lambda function through the configured IAM policies can only call the ```getPost``` and ```listPosts``` query.
+> The above schema assumes a default authentication type of **Amazon Cognito User Pools** and a secondary authentication type of **IAM**.
```
jpignata, line 956: This is duplicated information.

docs/cli/graphql-transformer/directives.md
```diff
@@ -886,7 +953,9 @@ type Post @model
...
+
+In the example above the `Post` model is protected by Cognito User Pools by default and the `owner` can perform any operation on the `Post` type. A Lambda function through the configured IAM policies can also call the `getPost` and `listPosts` query.
```
jpignata, line 958: "A Lambda function through the configured IAM policies?"

I'm not sure, but I think this means:

> You can also call `getPosts` and `listPosts` from an AWS Lambda function if it is configured with the appropriate IAM policies in its execution role.

docs/cli/graphql-transformer/directives.md
dabit3: Good point, updated.

docs/cli/graphql-transformer/directives.md
dabit3: Agreed, I've made some updates here. I think "All users" is misleading because owners would be included here, instead I think we need to explain what "other" means or have another better word. I used this: 

 **owner** refers to the user who created the object,  **other** refers to all other users.

docs/cli/graphql-transformer/directives.md
dabit3: Updated description slightly.

docs/cli/graphql-transformer/directives.md
dabit3: 🔥🔥🔥🔥

docs/cli/graphql-transformer/directives.md
dabit3: I've made a few more updates, let me know what you think.

dabit3:
Ok @jpignata, that was some really useful feedback. I've pushed up a bunch of new updates, please have a look and let me know what you think.

docs/cli/graphql-transformer/directives.md
```diff
@@ -453,11 +453,13 @@ enum ModelMutation { create update delete }
 
 > Note: The operations argument was added to replace the 'queries' and 'mutations' arguments. The 'queries' and 'mutations' arguments will continue to work but it is encouraged to move to 'operations'. If both are provided, the 'operations' argument takes precedence over 'queries'.
 
-### Owner Authorization
+### Owner authorization
+
+Enabling `owner` authorization allows any signed in user to create records.
```
undefobj, line 458: Maybe put "by default" in here.

docs/cli/graphql-transformer/directives.md
```diff
@@ -476,70 +478,126 @@ type Post
...
+Owner authorization specifies whether a user can access or operate against an object. To do so, each object will get an `ownerField` field (by default `owner` will be added to the object if not specified) that stores ownership information and is verified in various ways during resolver execution.
 
-You can use the *operations* argument to specify which operations are augmented as follows:
+You can use the `operations` argument to specify which operations are enabled as follows:
 
 - **read**: If the record's owner is not the same as the logged in user (via `$ctx.identity.username`), throw `$util.unauthorized()` in any resolver that returns an object of this type.
-- **create**: Inject the logged in user's `$ctx.identity.username` as the *ownerField* automatically.
-- **update**: Add conditional update that checks the stored *ownerField* is the same as `$ctx.identity.username`.
-- **delete**: Add conditional update that checks the stored *ownerField* is the same as `$ctx.identity.username`.
+- **create**: Inject the logged in user's `$ctx.identity.username` as the `ownerField` automatically.
```
undefobj, line 492: I wonder if this is too much details. Do we need to have all the VTL info?

docs/cli/graphql-transformer/directives.md
```diff
@@ -476,70 +478,126 @@ type Post
...
+- **update**: Add conditional update that checks the stored `ownerField` is the same as `$ctx.identity.username`.
+- **delete**: Add conditional update that checks the stored `ownerField` is the same as `$ctx.identity.username`.
 
-**Note**: When specifying operations as a part of the @auth rule, the operations not included in the list are not protected by default. For example, let's say you have the following schema:
+You must ensure that the `create` operations rule is either specified directly or indirectly to ensure that the owner's identity is stored with the record so it can be verified on subsequent requests.
```
undefobj, line 496: I would change "direct/indirect" to "is specified explicitly or inferred from defaults" and then say which is the case in the comments above the examples below. It's not clear why case #3 below is bad, and if it is bad why would you do it?

docs/cli/graphql-transformer/directives.md
```diff
@@ -476,70 +478,126 @@ type Post
...
+  @auth(rules: [{ allow: owner }]) {
   id: ID!
   updatedAt: AWSDateTime!
   content: String!
 }
 ```
 
-In this schema, only the owner of the object has the authorization to perform read (getTodo and listTodos) operations on the owner created object. But this does not prevent any other owner (any user other than the creator or owner of the object) to update/delete some other owner's object.
-Here's a truth table for the above-mentioned schema. In the table below `other` refers to any user other than the creator or owner of the object.
+In this schema, only the owner of the object has the authorization to perform read (`getTodo` and `listTodos`), update (`updateTodo`), and delete (`deleteTodo`) operations on the owner created object. This prevents the object from being updated or deleted by users other than the creator of the object.
```
undefobj, line 536: I wonder if the spirit of this example is lost. Sounds like the original intention was to show an example where users could create their own objects and other users could just do get/list against it.

undefobj: Ok reading further, I think you address this below.

docs/cli/graphql-transformer/directives.md
dabit3: I was wondering the same thing. I've updated to be clear without mentioning anything about $ctx or resolver language.

docs/cli/graphql-transformer/directives.md
dabit3: Oh I was looking for a good way to word this and this nails it. Thank you!

PR #1831, opened 2020-05-26
Fix compile error in Java code sample

richardmcclellan:

### DIFF
--- a/docs/start/getting-started/fragments/android/native_integrate.md
+++ b/docs/start/getting-started/fragments/android/native_integrate.md
@@ -102,7 +102,7 @@ Next, you'll create a Todo and save it to DataStore.
   ```kotlin
   Amplify.DataStore.save(
           item,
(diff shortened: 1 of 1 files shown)


docs/start/getting-started/fragments/android/native_integrate.md
```diff
@@ -90,7 +90,7 @@ Next, you'll create a Todo and save it to DataStore.
   ```java
     Amplify.DataStore.save(
             item,
-            success -> Log.i("Tutorial", "Saved item: " + success.item.getName()),
+            success -> Log.i("Tutorial", "Saved item: " + success.item().getName()),
```
jpignata, line 93:
```suggestion
            success -> Log.i("Tutorial", "Saved item: " + item.getName()),
```

richardmcclellan: I think it's better to keep referencing `success.item()`.  Theoretically it should be the same as `item`, but good to confirm.

PR #1832, opened 2020-05-26
Adds Android & iOS Auth Docs

TrekSoft:
Default proguard settings have allowAccessModification enabled which can turn private/protected members into public.

From the proguard manual:

"you probably shouldn't use this option when processing code that is to be used as a library, since classes and class members that weren't designed to be public in the API may become public."

The only way to remove that is to get rid of the defaultProguardFile and move all settings to a .pro file minus the allowAccessModification setting

https://issuetracker.google.com/issues/147447502

Apparently Google is working to fix this as it is definitely a bug in default behavior for a library: https://issuetracker.google.com/issues/144300603


docs/lib/auth/fragments/native_common/getting_started/common.md
```diff
@@ -56,4 +56,4 @@ The isSignedIn property of the authSession will be false since we havent signed
 Congratulations! You've successfully setup AWS Cognito Auth plugin.  Check out the following links to see other Amplify Auth use cases:
 
 * [Authenticate a user with username password]
-* [Authenticate a user with another auth provider]
\ No newline at end of file
+* [Authenticate a user with another auth provider]
```
wooj2: Seems like we need to add links here?

docs/lib/auth/fragments/android/access_credentials/10_fetchAuthSession.md
```diff
@@ -0,0 +1,15 @@
+```java
+Amplify.Auth.fetchAuthSession(
+        result -> {
+            AWSCognitoAuthSession cognitoAuthSession = (AWSCognitoAuthSession) result;
+
+            switch(cognitoAuthSession.getAWSCredentials().getType()) {
+                case SUCCESS:
+                    Log.i("AuthQuickStart", "AWSCredentials" + cognitoAuthSession.getAWSCredentials().getValue().toString());
```
rjuliano, line 8:
```suggestion
                case SUCCESS:
                    Log.i("AuthQuickStart", "AWSCredentials" + cognitoAuthSession.getAWSCredentials().getValue().toString());
                    break;
```

palpatim:
iOS: the `fetchUserAttributes` example should use a trailing closure syntax instead of the named closure

royjit:
> fetchUserAttributes



> iOS: the `fetchUserAttributes` example should use a trailing closure syntax instead of the named closure

Updated

PR #1833, opened 2020-05-26
[Android][GettingStarted] Remove Amplify Tools

lawmicha:
*Issue #, if available:*

*Description of changes:*
- Should we remove Amplify Tools dependency? iOS getting started does not have it and is a specific DataStore use case. 

*Misc*
- "Project: Todo" was confusing, it was actually "Project `MyAmplifyApp`"


jpignata:
We should not remove the Android Studio IDE integration from the directions, is that the question?

lawmicha:
> We should not remove the Android Studio IDE integration from the directions, is that the question?

is `com.amplifyframework.amplifytools` the same amplify tools that you can modify the schema, build and automatically run `amplify codegen models` ?

jpignata:
Yes - it's the IDE integration with Android Studio that allows you to codegen models and push.

PR #1834, opened 2020-05-26
Update create your application with logging and remove error

wooj2:
@royjit found a difference in behavior for an empty config.. updating the configuration to:
* Remove the error that was coming up before
* Described how to turn on logging.

### DIFF
--- a/docs/lib/project-setup/fragments/ios/create-application/40_verifyAmplifyLibraries.md
+++ b/docs/lib/project-setup/fragments/ios/create-application/40_verifyAmplifyLibraries.md
@@ -15,7 +15,15 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/lib/project-setup/fragments/ios/create-application/40_verifyAmplifyLibraries.md
```diff
@@ -15,7 +15,15 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
 }
 ```
 
-Build your project (`CMD + b`).  If your build succeeeds, then you have successfully added the Amplify library to your project.  In order to run the application (without it erroring), you will need to provision resources in the backend.  If you attempt run your application at this point, you will see the following error:
+Build your project (`CMD + b`).  If your build succeeeds, then you have successfully added the Amplify library to your project.  If you run the application, you should not see any errors being printed to the console.
```
jpignata, line 18: If I shouldn't see them, maybe we don't need to explicitly call this out.

PR #1843, opened 2020-05-26
Updates from dogfooding graphql api

wooj2:
I think the concepts page needs to be re-worked for iOS... also, did not dogfood configure auth modes yet.

### DIFF
--- a/docs/lib/graphqlapi/fragments/ios/getting-started/20_installLib.md
+++ b/docs/lib/graphqlapi/fragments/ios/getting-started/20_installLib.md
@@ -3,6 +3,7 @@ To install the Amplify API to your application, **add `AmplifyPlugins/AWSAPIPlug
 ```ruby
 target 'MyAmplifyApp' do
   use_frameworks!
+  pod 'Amplify'
   pod 'AmplifyPlugins/AWSAPIPlugin'
 end
 ```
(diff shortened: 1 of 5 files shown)


docs/lib/graphqlapi/fragments/ios/getting-started/20_installLib.md
```diff
@@ -3,6 +3,7 @@ To install the Amplify API to your application, **add `AmplifyPlugins/AWSAPIPlug
 ```ruby
 target 'MyAmplifyApp' do
   use_frameworks!
+  pod 'Amplify'
```
lawmicha, line 6: plugins should pull Amplify in already, any reason to add this explicitly?

wooj2: Technically you are right,, it is not needed, but project setup has it, and we say that your {odfile should look something like "this".  So by removing it, you might leave readers a bit confused on whether or not they were suppose to remove it.

lawmicha: okay sounds good, i'm running through the tutorial at the moment and have draft PR for removing unnecessary Pods here https://github.com/aws-amplify/docs/pull/1844 if we decide to remove it, then i'll go and update all places

PR #1846, opened 2020-05-26
Fix navigation issues

harrysolovay:
Uses RAF, as there are no event-listening capabilities for certain URL change events, oddly. This solution is light enough IMO. & most importantly, it works consistently!

### DIFF
--- a/client/src/docs-ui/page/page.tsx
+++ b/client/src/docs-ui/page/page.tsx
@@ -1,14 +1,4 @@
-import {
-  Component,
-  Host,
-  h,
-  State,
-  Listen,
-  Element,
-  Prop,
-  Build,
-} from "@stencil/core";
-import {MatchResults} from "@stencil/router";
+import {Component, Host, h, State, Listen, Element, Build} from "@stencil/core";
 import {
   sidebarLayoutStyle,
   pageStyle,
@@ -47,9 +37,6 @@ const SELECTED_TABS_LOCAL_STORAGE_KEY = `amplify-docs::selected-tabs`;
 @Component({tag: "docs-page", shadow: false})
 export class DocsPage {
   @Element() el: HTMLElement;
-
-  /** match path */
-  @Prop() readonly match: MatchResults;
 
   @State() pageData?: Page;
   @State() blendUniversalNav?: boolean;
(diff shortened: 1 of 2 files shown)


client/src/docs-ui/page/page.tsx
```diff
@@ -107,14 +96,39 @@ export class DocsPage {
     }
   }
 
-  @Watch("match")
-  onRouteChange() {
-    this.getPageData();
+  componentDidLoad() {
+    let {pathname} = location;
```
jordanranz, line 100: Is there a reason for this to be outside of `updatePageData`?

client/src/docs-ui/page/page.tsx
```diff
@@ -107,14 +96,39 @@ export class DocsPage {
...
+    const updatePageData = () => {
+      if (location.pathname !== pathname) {
+        this.isFirstRenderOfCurrentPage = true;
+        pathname = location.pathname;
+        this.getPageData();
+      }
+    };
+    this.rafId = (function watchForRouteChange() {
```
jordanranz, line 108: This needs comments for explanation, it is unreadable

jordanranz: Also does the IIFE need to be nested into an assignment here?

client/src/docs-ui/page/page.tsx
```diff
@@ -107,14 +96,39 @@ export class DocsPage {
...
+    }
+    this.isFirstRenderOfCurrentPage = false;
   }
 
-  @Listen("popstate", {target: "window"})
-  onPopState() {
-    this.getPageData();
+  componentWillUnload() {
+    if (this.rafId) {
+      cancelAnimationFrame(this.rafId);
```
jordanranz, line 130: Add some comments here.

jordanranz:
Overall, I don't really understand why we have to use request animation frame to solve this. This seems like a hack.

client/src/docs-ui/page/page.tsx
```diff
@@ -130,72 +148,65 @@ export class DocsPage {
...
+              this.validFilterValue = allFilters.includes(filterValue);
+              if (this.validFilterValue) {
+                this.setSelectedFilters({[this.filterKey]: this.filterValue});
               }
-            } else {
-              this.filterValue = undefined;
             }
           } else {
-            this.filterKey = undefined;
+            this.filterValue = undefined;
```
jordanranz, line 197: What is this change for?

PR #1859, opened 2020-05-27
[iOS] [datastore] run-through sections updates

lawmicha:
*Issue #, if available:*

*Description of changes:*

1.  perhaps this is an xcode issue? i'm on 11.3.1
image

2.
image

**Removed this change from this PR due merge conflict**
3. Following the Non-relational sections have a Post model with optional rating, status. moving to relational then uses required fields, this breaks some of the existing functions i had added as i was following along. updating the schema to use optionals to match previous Post model in the relational example
```
type Post @model {
  id: ID!
  title: String!
  rating: Int // keep optional
  status: PostStatus // keep  optional
  # New field with @connection
  comments: [Comment] @connection(keyName: "byPost", fields: ["id"])
}
```

4. The lazy loading of comments is amazing! This is a minor change here to replace the comment about fetching with DataStore.query just to make it easier too run through
```
Amplify.DataStore.query(Post.self, byId: "111") {
    switch $0 {
    case .success(let post):
        if let postWithComments = post {
            if let comments = postWithComments.comments {
                for comment in comments {
                    print(comment.content)
                }
            }
        } else {
            print("Post not found")
        }
    case .failure(let error):
        print("Post not found - \(error.localizedDescription)")
    }
}
```
i think this is a good additional since even the next example has a DataStore.Query involved

5. when following the relational to many-to-many example, since it's on the same page, i already have code the uses rating and status. keeping this in the model will make it easier to add the code example

6. conflict resolution section - added more steps in the console section

7. confliction resolution `conflictHandler` fix

### DIFF
--- a/docs/lib/datastore/fragments/ios/data-access/query-predicate-multiple-snippet.md
+++ b/docs/lib/datastore/fragments/ios/data-access/query-predicate-multiple-snippet.md
@@ -1,6 +1,6 @@
 ```swift
 let p = Post.keys
-Amplify.DataStore.query(Post.self, where: { p.rating > 4 && p.status == .active }) {
+Amplify.DataStore.query(Post.self, where: p.rating > 4 && p.status == .active) {
     switch $0 {
     case .success(let result):
         // result if of type [Post]
--- a/docs/lib/datastore/fragments/ios/data-access/query-predicate-multiple-snippet.md
+++ b/docs/lib/datastore/fragments/ios/data-access/query-predicate-multiple-snippet.md
@@ -1,6 +1,6 @@
 ```swift
 let p = Post.keys
-let predicate = p.rating > 4 && p.status == .active
+let predicate = p.rating > 4 && p.status == PostStatus.active
 Amplify.DataStore.query(Post.self, where: predicate) {
     switch $0 {
     case .success(let result):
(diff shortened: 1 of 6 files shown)


docs/lib/datastore/fragments/ios/data-access/query-predicate-multiple-snippet.md
```diff
@@ -1,6 +1,7 @@
 ```swift
 let p = Post.keys
-Amplify.DataStore.query(Post.self, where: { p.rating > 4 && p.status == .active }) {
+let predicate = p.rating > 4 && p.status == .active
+Amplify.DataStore.query(Post.self, where: predicate) {
```
drochetti, line 4: can you keep the condition inline?

docs/lib/datastore/fragments/ios/data-access/query-predicate-multiple-snippet.md
```diff
@@ -15,7 +16,8 @@ You can also write this in a compositional function manner by replacing the oper
 
 ```swift
 let p = Post.keys
-Amplify.DataStore.query(Post.self, where: { p.rating.gt(4).and(p.status.eq(.active)) }) {
+let predicate = p.rating.gt(4).and(p.status.eq(.active))
+Amplify.DataStore.query(Post.self, where: predicate) {
```
drochetti, line 20: leave the predicate inline

docs/lib/datastore/fragments/ios/data-access/query-predicate-multiple-snippet.md
lawmicha: woops, some of the errors got me mixed up what was valid syntax. added back inline

PR #1909, opened 2020-06-03
correct errors in the API(GraphQL)/Directives section

UnleashedMind:
*Issue #, if available:*

*Description of changes:*
Correct errors found when testing the schemas in the document
https://github.com/aws-amplify/amplify-cli/pull/4092


docs/cli/graphql-transformer/directives.md
```diff
@@ -940,7 +940,7 @@ The @auth directive allows the override of the default provider for a given auth
 
 ```graphql
 # private authorization with provider override
-type Post @model @auth(rules: [{ allow: private, provider: oidc }]) {
+type Post @model @auth(rules: [{ allow: private, provider: iam }]) {
```
undefobj, line 943: Is this correct? The section is for "Authorization using an OIDC provider"

UnleashedMind: oidc is not supported with private auth: 
This is the error message by the CLI:  
`InvalidDirectiveError: @auth directive with 'private' strategy only supports 'userPools' (default) and 'iam' providers, but found 'oidc' assigned.`

docs/cli/graphql-transformer/directives.md
```diff
@@ -1367,8 +1374,8 @@ The @function directive allows you to quickly connect lambda resolvers to an App
 Let's assume you have deployed an *echo* function with the following contents:
 
 ```javascript
-exports.handler = function (event, context) {
-  context.done(null, event.arguments.msg);
+exports.handler = async event => {
```
undefobj, line 1377: Can we use the idiomatic `async function(event, context){...}` style from the Lambda docs? https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html

docs/cli/graphql-transformer/directives.md
undefobj: If that's the case then you need to change more than just this code sample, the text says it's for OIDC section

UnleashedMind: What would you suggest? 
We can remove the part that does not use the OIDC provider.

docs/cli/graphql-transformer/directives.md
UnleashedMind: changed

PR #1914, opened 2020-06-03
Removed erronous page for JS docs and "native_" prefix for native

brene:
*Issue #, if available:*

*Description of changes:*


jpignata:
This will change the URLs that have been out there for a week now. Can we create 301 redirects?

PR #1933, opened 2020-06-08
chore: enable prerendering of landing page

harrysolovay:
Throw content errors relating to prerender process memory leak. Fix related content errors. Enable prerendering of landing page.


capi/src/init-node/valid-tags.json
```diff
@@ -0,0 +1,56 @@
+{
```
jordanranz, line 1: I don't love the idea of maintaining this list manually.

docs/cli/start/workflows.md
```diff
@@ -39,8 +39,8 @@ During the init process, the root stack is created with three resources:
 The provider logs the information of the root stack and the resources into the project metadata file (amplify/backend/amplify-meta.json).
 The root stack's template can be found in `amplify/backend/awscloudformation`.
 
-### amplify <category> add
-Once init is complete, run the command `amplify <category> add` to add resources of a category to the cloud. This will place a CloudFormation template for the resources of this category in the category's subdirectory `amplify/backend/<category>` and insert its reference into the above-mentioned root stack as the nested child stack. When working in teams, it is good practice to run an `amplify pull` before modifying the backend categories.
+### amplify \<category\> add
```
jordanranz, line 42: If we do maintain a valid tags set, do we still need to escape or can we just infer that this is text at build time.

harrysolovay: We need to escape it, otherwise it'll produce an invalid tag error at build time.

capi/src/init-node/html-to-hyperscript.ts
```diff
@@ -40,10 +41,28 @@ const Hyperscript = (
...
   }
+
+  if (node.nodeName === "#comment") {
+    return null;
+  }
+
+  if (node.tagName) {
+    if (!validTags[node.tagName] && !node.tagName.startsWith("amplify-")) {
+      throw new Error(
+        `Invalid tag "${node.tagName}" encountered in "${srcPath}"`,
```
jordanranz, line 52: Can we add how a user could resolve this issue. Maybe just a note about escaping it.

docs/cli/start/workflows.md
jordanranz: ok

PR #1949, opened 2020-06-11
lighthouse fixes

harrysolovay:
* minify stencil-output JS
* ensure alt tags in all imgs (with the exception of doc-writer-define imgs)
* include Algolia styles in app.css bundle
* update viewport meta tag to meet accessibility standard (zoomable)
* preload critical js & css
* 512x512 icon in manifest
* font-display swap for non-blocking on text
* enable Stencil-built service workers


client/src/docs-ui/chat-button/__snapshots__/chat-button.spec.ts.snap
```diff
@@ -3,7 +3,7 @@
 exports[`docs-chat-button Render logic should render 1`] = `
 <docs-chat-button>
   <amplify-external-link class="css-swyh7p" graphic="white" href="https://discord.gg/jWVbPfC">
-    <img class="css-dp2ecj" src="/assets/discord-white.svg">
+    <img alt="discord logo" class="css-dp2ecj" src="/assets/discord-white.svg">
```
jordanranz, line 6: Capitalize this

client/src/docs-ui/footer/footer.tsx
```diff
@@ -8,6 +8,7 @@ import {
   socialLinkContainerStyle,
 } from "./footer.style";
 import * as links from "../../constants/links";
+import * as img from "../../constants/img";
```
jordanranz, line 11: Looks like we do individual imports in most places. Let's keep it consistent and do it here as well

PR #1956, opened 2020-06-12
chore: begin offloading computation to workers

harrysolovay:
[lighthouse recommendation](https://web.dev/mainthread-work-breakdown/?utm_source=lighthouse&utm_medium=devtools)


client/src/cache.worker.ts
```diff
@@ -0,0 +1,13 @@
+import {getPage as capiGetPage} from "./api/get-page";
+import {Page} from "./api/types/page";
```
sammartinez, line 2: Is this file still needed since we are removing the `cache.ts` file?

client/src/docs-ui/internal-link/internal-link.tsx
```diff
@@ -25,9 +25,9 @@ export class DocsInternalLink {
   @State() isChildActive?: boolean;
 
   @Watch("selectedFilters")
-  componentWillLoad() {
+  async componentWillLoad() {
```
sammartinez, line 28: Does this affect anything on Stencil side since it is there lifecycle method?

client/src/docs-ui/select-anchor/reroute-if-necessary.worker.ts
```diff
@@ -0,0 +1,21 @@
...
+export const rerouteIfNecessary = async (path: string) => {
+  if (rereouteCache.has(path)) {
+    return rereouteCache.get(path);
+  }
+  const rerouted = await (async () => {
+    if (
+      path.includes("/lib") &&
+      (path.includes("/q/platform/ios") || path.includes("/q/platform/android"))
+    ) {
+      return `/lib/q/platform/${(await parseURL(path)).params
```
sammartinez, line 14: What happens with the path param fails here? Does it break us?

package.json
```diff
@@ -36,6 +37,7 @@
     "eslint-config-prettier": "^6.5.0",
     "eslint-plugin-prettier": "^3.1.1",
     "eslint-plugin-react": "^7.16.0",
+    "fs-extra": "^9.0.1",
```
sammartinez, line 40: New dependency?

client/src/amplify-ui/code-block/code-block.tsx
```diff
@@ -59,9 +59,9 @@ export class AmplifyCodeBlock {
       return (
         <div class={lineCountStyle}>
           <div>
-              {new Array(this.lineCount)
-                .fill(null)
-                .map((_, i) => <span>{String(i + 1)}</span>)}
+            {new Array(this.lineCount).fill(null).map((_, i) => (
```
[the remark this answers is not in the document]
harrysolovay: An update to prettier must have resulted in different formatting. I didn't write this change. Must have been the dirty work of command + s. Nbd.

client/src/docs-ui/internal-link/internal-link.tsx
harrysolovay: Stencil's `componentWillLoad` lifecycle method can be async. It does not affect anything.

client/src/docs-ui/select-anchor/reroute-if-necessary.worker.ts
harrysolovay: What do you mean?

package.json
harrysolovay: Yea. It's effectively the same module as the core fs module. Just promise-based.

client/src/docs-ui/select-anchor/reroute-if-necessary.worker.ts
```diff
@@ -0,0 +1,21 @@
...
+const rereouteCache = new Map<string, string>();
+
+export const rerouteIfNecessary = async (path: string) => {
+  if (rereouteCache.has(path)) {
+    return rereouteCache.get(path);
+  }
+  const rerouted = await (async () => {
+    if (
+      path.includes("/lib") &&
+      (path.includes("/q/platform/ios") || path.includes("/q/platform/android"))
```
jordanranz, line 12: What exactly is this function doing? It seems strange to hardcode these values here when they are constants in a FILTER enum that we maintain.

client/src/docs-ui/select-anchor/reroute-if-necessary.worker.ts
```diff
@@ -0,0 +1,21 @@
+import {parseURL} from "../../utils/url/url.worker";
+
+const rereouteCache = new Map<string, string>();
+
+export const rerouteIfNecessary = async (path: string) => {
```
jordanranz, line 5: This name is a bit ambiguous. Can we reword this?

client/src/docs-ui/select-anchor/reroute-if-necessary.worker.ts
jordanranz: I think he is asking what happens when the underlying `serializeURL` fails

client/src/docs-ui/select-anchor/select-anchor.tsx
```diff
@@ -62,11 +57,50 @@ export class DocsSelectAnchor {
...
+    this.selectedOption = filterKey && this.selectedFilters?.[filterKey];
+    this.selectedOptionMetadata =
+      filterKey &&
+      this.selectedOption &&
+      (filterMetadataByOptionByName[filterKey][
+        this.selectedOption
+      ] as FilterOptionMetadata);
+
+    this.options =
+      this.sortedVersions &&
```
jordanranz, line 72: Let's break this out into another function like `getOptions`

client/src/docs-ui/select-anchor/select-anchor.tsx
```diff
@@ -62,11 +57,50 @@ export class DocsSelectAnchor {
...
+      ((await Promise.all(
+        this.sortedVersions.map(async ([filterValue, filterRoute]) => {
+          if (filterValue !== this.selectedOption) {
+            const meta =
```
jordanranz, line 76: Let's change this variable name to describe what the boolean represents

client/src/docs-ui/select-anchor/reroute-if-necessary.worker.ts
harrysolovay: It won't (see how it ensures the path includes the virtual query params)

client/src/docs-ui/select-anchor/reroute-if-necessary.worker.ts
harrysolovay: Interesting. Not sure if this is needed anymore. I believe it was originally responsible for rerouting to sdk as the default. Then someone must have swapped out the "/sdk" portions of these strings with "/lib" when it went GA. Unless you can think of a reason not to, I'll get rid of this dud code.

client/src/docs-ui/select-anchor/select-anchor.tsx
harrysolovay: Be more specific. What variable name?

jordanranz: `meta`

harrysolovay: Renamed

client/src/docs-ui/select-anchor/reroute-if-necessary.worker.ts
harrysolovay: Got rid of it altogether

PR #1968, opened 2020-06-16
Enabling autotrack of sessions & pageView

brene:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/client/src/utils/track.ts
+++ b/client/src/utils/track.ts
@@ -8,6 +8,14 @@ if (!configured) {
   Auth.configure(awsexports);
   Analytics.configure(awsexports);
   configured = true;
+
+  Analytics.autoTrack("session", {
+    enable: true,
+  });
+  Analytics.autoTrack("pageView", {
+    enable: true,
+    type: "SPA",
+  });
 }
 
 export enum AnalyticsEventType {


harrysolovay:
Here's the issue: AmplifyJS page view tracking utilizes the [`pushstate`](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState), [`popstate`](https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event) and [`replacestate`](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState) events. The docs site's routing does not work via these events. Instead, we have a light [raf](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) loop, which observes URL changes (all of which those events do not detect) and swaps out page data accordingly. Also, `replaceState` is called as users scroll between sections (to reflect the ID in the URL). This would result in duplicate tracking for any given page. I don't believe this is the right solution. Sadly, I think we need to implement additional analytics requirements in a way that's more tailored to this project's routing mechanism. Let's discuss offline.

harrysolovay:
@brene is it alright if we close this PR out for now?

PR #1979, opened 2020-06-18
fix(storage/react-native): Update docs for react native storage components

### LINKED ISSUE #1913: The docs for react native storage component is incorrect
Docs site reference - https://docs.amplify.aws/ui/storage/s3-album/q/framework/react-native

The doc for storage component has same content as of react and is no where related to react native misleading our customers.

cedricgrothues:
Fixes #1913 

*Description of changes:* The current documentation for the react-native storage components has the same content as the react component, even though they are not related at all. This PR adapts the documentation for the react-native components.


docs/ui/storage/fragments/react-native/s3-image.md
```diff
@@ -1,82 +1,40 @@
...
-<S3Image path={path} picker fileToKey={fileToKey} />
+return (
+    <S3Image imgKey={key} body={this.state.image_body} />
+);
 ```
 
-<amplify-callout>
-
-`S3Image` will escape all spaces in key values to underscore. For example, 'a b' will be converted to 'a_b'.
+Set the `resizeMode` property to determine how resize the image if the frame doesn't match the image dimensions:
```
mauerbac, line 38: `...to determine how to resize the image...`

docs/ui/storage/fragments/react-native/s3-album.md
```diff
@@ -1,27 +1,16 @@
...
 
-![Image](~/images/S3Album_and_code.png)
-
-```jsx
-import { S3Album } from 'aws-amplify-react-native';
-
-render() {
-    return <S3Album path={path} />
-```
+![S3Album Example](~/images/s3album-react-native.png)
```
ashika01, line 3: Lets remove this image.Lets create a code block with the sample code in the image. Also, add an image on top of this code block with how S3Album look on ios and android simulators.

docs/ui/storage/fragments/react-native/s3-album.md
```diff
@@ -1,27 +1,16 @@
...
```
ashika01, line 4: Lets add a line on what `path` prop does over here.

docs/ui/storage/fragments/react-native/s3-image.md
```diff
@@ -1,82 +1,40 @@
 `S3Image` component renders an *Amazon S3 object key* as an image:
 
 ```jsx
```
ashika01, line 3: Lets try to render an image on both ios and android simulator and update it here.

ashika01:
@cedricgrothues This PR seems to address gap in components docs. Could you have a look at this http://localhost:3333/ui-legacy/storage/customization/q/framework/react-native and address any gaps here as well?

cedricgrothues:
> @cedricgrothues This PR seems to address gap in components docs. Could you have a look at this http://localhost:3333/ui-legacy/storage/customization/q/framework/react-native and address any gaps here as well?

Judging from the comments in the [`Storage` implementation](https://github.com/aws-amplify/amplify-js/blob/master/packages/storage/src/providers/AWSS3Provider.ts#L218-L219),  this doesn't seem to be possible at all. According to that, the `config` object may only contain the `level`, `contentType`, or `progressCallback` parameters.

PR #1984, opened 2020-06-18
Include warning about build image

ajonp:
Update to docs warning about build setting versioning (not something that is covered in official docs).


ajonp:
Hoping this doesn't let this item go stale. @renebrandel is there someone specifically focused on docs?

sw-yx (approved):

mauerbac:
@ajonp - so sorry for the delay on this. I just reviewed this two things 

1) The language is a bit too causal for the docs. Any chance you can make it more formal
2) Non angular devs might get confused. Can we make it more generic and remove the example? Thank you

Issue #1971, opened 2020-06-17
AWSPinpointAnalyticsPlugin requires Application in constructor


itirbit:
**Page**: [`/lib/analytics/getting-started/q/platform/android`](https://docs.amplify.aws/lib/analytics/getting-started/q/platform/android)

**Feedback**:

If you follow the instruction 

        try {
            Amplify.addPlugin(new AWSCognitoAuthPlugin());
            Amplify.addPlugin(new AWSPinpointAnalyticsPlugin());
            Amplify.configure(getApplicationContext());
            Log.d(TAG, "Initialize Amplify");
        } catch (AmplifyException error){
            Log.e(TAG, "Couldn't initialize Amplify", error);
        }

it will get to an error

    C:\Projects\tm\build\android\MyActivity.java:397: error: constructor AWSPinpointAnalyticsPlugin in class AWSPinpointAnalyticsPlugin cannot be applied to given types;
            Amplify.addPlugin(new AWSPinpointAnalyticsPlugin());
                              ^
      required: Application
      found: no arguments

I guess it should be

     Amplify.addPlugin(new AWSPinpointAnalyticsPlugin(getApplication()));

PR #1998, opened 2020-06-23
Fix analytics add documentation

jpignata:
*Issue #, if available:* #1971

*Description of changes:* `AWSPinpointAnalyticsPlugin` expects the application to be passed in during initialization.


richardmcclellan:
This fix works if the sample code is used inside of an `Activity`.  However, since these docs tell the user to put this sample code inside your `Application`, I think `this` should be used as the parameter.

jpignata:
> This fix works if the sample code is used inside of an `Activity`. However, since these docs tell the user to put this sample code inside your `Application`, I think `this` should be used as the parameter.

You're right! I'll run through a test on both and confirm.

jpignata:
@richardmcclellan Thanks! Tested on both Java and Kotlin. Fix pushed.

PR #2005, opened 2020-06-24
Getting Started fixes for Android

jpignata:
*Issue #, if available:* n/a

*Description of changes:* Fixes for the Getting Started guide and updates for Android Studio 4.0:

- Replace the call to Hub with DataStore.observe as this is what that
  facility is for and Hub is moreso plumbing.
- Change the guide to create a standalone application class in the
  integration section to avoid a possible race condition exposed by
  Android Studio 4.0.


docs/start/getting-started/fragments/android/integrate.md
```diff
@@ -2,44 +2,98 @@ Next you'll use the generated model to read and write data. In this section you'
...
   </amplify-block-switcher>
 
+1. Open **manifests** > **AndroidManifest.xml** to configure your application.
+
+1. Add the `android:name` and `tools:replace` attributes to the `application` node:
+
+  ```xml
+  <application
+      android:name=".MyAmplifyApplication"
+      tools:replace="android:name"
```
jamesonwilliams, line 66: This probably breaks Multidex on devices with API < 21.

The Google docs on this topic propose [three ways to embed multidex into the application class](https://developer.android.com/studio/build/multidex). Our library currently uses the first of these options. More resilient for consumers would probably be either the second or third options. The forth (and even better option) would be to raise the `minSdk` on our product tin to 21, so that the on-boarding were as smooth as possible.

docs/start/getting-started/fragments/android/add-api.md
```diff
@@ -20,24 +20,11 @@ Now that your have DataStore persisting data locally, in the next step you'll co
...
-                  Log.i("Tutorial", "Priority: " + todo.getPriority().toString());
-              }
-
-              if (todo.getDescription() != null) {
-                  Log.i("Tutorial", "Description: " + todo.getDescription());
-              }
-          }
+  Amplify.DataStore.observe(Todo.class,
+          started -> Log.i("Tutorial", "Observation began."),
+          change -> Log.i("Tutorial", change.item().toString()),
```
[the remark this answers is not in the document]
jpignata: A TODO!

docs/start/getting-started/fragments/android/integrate.md
jpignata: Gotcha, yeah, skimmed there earlier. I think we need to huddle up on this Multidex stuff and think about the right approach.

PR #2016, opened 2020-06-25
Bump ui version for updated prop docs

jordanranz:
*Description of changes:*
- Bump UI module version
- Gain new prop docs for the authenticator

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "dependencies": {
     "@aws-amplify/analytics": "^3.1.7",
     "@aws-amplify/auth": "^3.2.4",
-    "@aws-amplify/ui-components": "^0.5.2",
+    "@aws-amplify/ui-components": "latest",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
   "dependencies": {
     "@aws-amplify/analytics": "^3.1.7",
     "@aws-amplify/auth": "^3.2.4",
-    "@aws-amplify/ui-components": "^0.5.0",
+    "@aws-amplify/ui-components": "^0.5.2",
```
ericclemmons, line 17: What are your thoughts on using `latest`?  One problem is that `yarn.lock` won't respect it, so it'll require regenerating the lock file.

I feel that when docs are deployed, these should always be updated (such that daily deployments of the doc would always reflect `latest`).

jordanranz: yea super good point

PR #2048, opened 2020-07-01
feat: android snippets for next datastore guide

jamesonwilliams:


docs/lib/datastore/fragments/android/getting-started/30_platformIntegration.md
```diff
@@ -1,2 +1,34 @@
...
+    repositories {
+        mavenLocal()
+        google()
+        jcenter()
+    }
+
+    dependencies {
+        classpath 'com.android.tools.build:gradle:4.0.0'
+        // Add this next line:
+        classpath 'com.amplifyframework:amplify-tools-gradle-plugin:1.0.1'
```
drochetti, line 18: are you guys updating the version on every release? On iOS we decided to drop the version on documentation, it makes our lives way easier.

ps: no need to address it now, I'm just curious.

jamesonwilliams: Hm. Yea, we can use `:1.+` or something. It's not a best practice, but it would keep our documentation correct, more-of-the-time.

PR #2053, opened 2020-07-01
[Guides] warn up front that dynamic Nextjs is not yet supported

sw-yx:
warn up front that dynamic Nextjs is not supported - moving this important info from the bottom to the top of the Guide.


sw-yx:
cc @dabit3

dabit3:
Thanks @swyx!

PR #2086, opened 2020-07-09
[S3] documentation update on complex object

DongQuanRui:
*Description of changes:*
The previous version of complex object documentation contains some confusion. I figured out myself on uploading complex object using `AWSS3TransferUtility`. So I am also updating the documentation.


docs/sdk/storage/fragments/ios/graphql-api.md
```diff
@@ -1,126 +1,36 @@
...
-        }
-    }
-  }
-```
-**Response Mapping Template**
-```
-  $util.toJson($util.dynamodb.fromS3ObjectJson($context.source.file))
-```
-The AWS AppSync SDK doesn't take a direct dependency on the AWS SDK for S3, but takes in `AWSS3TransferUtility` and `AWSS3PresignedURLClient` clients as part of `AWSAppSyncClientConfiguration`. The code generator used above for generating the API generates the S3 wrappers required to use the previous clients in code. To generate the wrappers, pass the `--add-s3-wrapper` flag while running the code generator tool. You will also need to take a dependency on the `AWSS3` SDK. You can do that by updating your Podfile:
+and run `amplify push` to update the AppSync backend resource, remember to choose `udpate API.swift` when `Amplify CLI` prompts the choice
```
lawmicha, line 31: `udpate`

PR #2092, opened 2020-07-10
chore: clarify ios s3 escape hatch purpose

jamesonwilliams:
The current escape hatch wording de-emphasizes the utility of the AWSS3 client.

This is resolved by offering a more balanced wording as to why one would use the escape hatch.


jamesonwilliams:
@rromanchuk

PR #2106, opened 2020-07-13
fix(getting-started/data-model): Update incorrect path

### LINKED ISSUE #2060: [Feedback]Incorrect Path In Example
**Page**: [`/start/getting-started/data-model/q/integration/js`](https://docs.amplify.aws/start/getting-started/data-model/q/integration/js)

**Feedback**:

In the example, after we run:

`amplify add api`

for the `? Provide API name:`, we enter: `amplifyjsapp`

which generates the following directory:

amplify/backend/api/**amplifyjsapp**/schema.graphql

but the example shows: **myapi** instead.

So either the API name should be: myapi in which case the path will be correct, or the path should be changed to amplifyjsapp

cedricgrothues:
Fixes #2060 

*Description of changes:* Update an incorrect path from `myapp` to `amplifyjsapp`

### DIFF
--- a/docs/start/getting-started/fragments/vanillajs/data-model.md
+++ b/docs/start/getting-started/fragments/vanillajs/data-model.md
@@ -39,7 +39,7 @@ Accept the **default values** which are highlighted below:
 
 The CLI should open this GraphQL schema in your text editor.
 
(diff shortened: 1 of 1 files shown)


mauerbac:
Thanks @cedricgrothues !

PR #2112, opened 2020-07-15
add missing fields to Creating a Form API with DynamoDB and Lambda guide

sw-yx:
*Issue #, if available:*

*Description of changes:* added missing fields which are hard to add later due to related bug https://github.com/aws-amplify/amplify-cli/issues/4840

cc @dabit3


dabit3:
Thanks!!

sw-yx:
note that the guide still doesnt actually work as is because of the bug mentioned above

PR #2114, opened 2020-07-15
Update pinpoint SDK notifications handling for iOS 10 and above

lawmicha:
*Issue #, if available:*

*Description of changes:*
This adds the code snippet for handling push notifications in iOS 10 and up, passes userInfo to pinpoint SDK for creating back metric events for pinpoint campaign/journey

related:
https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/1649518-usernotificationcenter support is iOS 10+


docs/sdk/push-notifications/fragments/ios/getting-started.md
```diff
@@ -162,6 +163,22 @@ Use the following steps to connect push notification backend services to your ap
         )
     }
     ```
+    For iOS 10 and above, pass the notification event to pinpoint SDK in `userNotificationCenter(_:willPresent:withCompletionHandler:)` and `userNotificationCenter(_:didReceive:withCompletionHandler:)`
+    ```swift
+    extension AppDelegate: UNUserNotificationCenterDelegate {
+    func userNotificationCenter(_ center: UNUserNotificationCenter, willPresent notification: UNNotification, withCompletionHandler completionHandler: (UNNotificationPresentationOptions) -> Void) {
+        // Handle foreground push notifications
+        pinpoint!.notificationManager.interceptDidReceiveRemoteNotification(notification.request.content.userInfo, fetchCompletionHandler: { _ in })
```
palpatim, line 171: Do we make it a habit of force unwrapping the object in these documentation snippets?￼

docs/sdk/push-notifications/fragments/ios/getting-started.md
```diff
@@ -162,6 +163,22 @@ Use the following steps to connect push notification backend services to your ap
...
+        completionHandler(.badge)
```
palpatim, line 172: Should the completion handler be invoked inside pinpoints fetch completion handler, or afterwards as you have here? Basically, I don't know what's happening in the fetch completion handler, so I don't know whether there's going to be a race condition from invoking it immediately after.￼

lawmicha: This brings up a good point, the pinpoint SDK actually doesn't do anything with the fetch completion handler. we should improve the documentation to mention that the developer always needs to call it and reference the apple docs

- there's a comment in the pinpoint SDK that says 
```
//We must rely on the user calling the completion handler because if we call it ourselves as well as the user it would cause a crash due to calling it twice.
```
- fetchCompletionHandler is never used inside [pinpoint SDK](https://github.com/aws-amplify/aws-sdk-ios/blob/0f985322625566a566726e8ca2c24d5a5bc678b8/AWSPinpoint/AWSPinpointNotificationManager.m#L136)

- the developer is still responsible for invoking it. apple docs indicate that developer always needs to invoke it (https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/1649518-usernotificationcenter). If the older docs is any indication, it talks about the impact of not calling the completion handler [didReceiveRemoteNotification:fetchCompletionHandler](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623013-application) 

> As soon as you finish processing the notification, you must call the block in the handler parameter or your app will be terminated. Your app has up to 30 seconds of wall-clock time to process the notification and call the specified completion handler block. In practice, you should call the handler block as soon as you are done processing the notification. The system tracks the elapsed time, power usage, and data costs for your app’s background downloads. Apps that use significant amounts of power when processing remote notifications may not always be woken up early to process future notifications.

docs/sdk/push-notifications/fragments/ios/getting-started.md
lawmicha: updated

docs/sdk/push-notifications/fragments/ios/getting-started.md
palpatim: OK, then we should deprecate this API and provide a replacement that doesn't take a `fetchCompletionHandler` argument.

lawmicha: makes sense, PR here https://github.com/aws-amplify/aws-sdk-ios/pull/2910

lawmicha:
Once https://github.com/aws-amplify/aws-sdk-ios/pull/2910 is completed, the documentation can be updated to use the latest API methods (the ones that do not take in the extra `fetchCompletionHandler`)

palpatim (approved):

PR #2141, opened 2020-07-23
[iOS] AWSAPIPlugin GraphQL Advanced workflow docs with custom GraphQL requests

lawmicha:
*Issue #, if available:*

*Description of changes:*
- this adds some info about creating custom graphql requests
- had decided to put this in advanced workflows but open to moving this somewhere else
- we should also create code snippets for android/ios and there may be some API differences, ie. `decodePath` in iOS

Some use cases that this will address:
 - https://github.com/aws-amplify/amplify-ios/issues/663 (configurable max depth, only retrieve subset of fields)
 - https://github.com/aws-amplify/amplify-ios/issues/661 (custom inputs)
 - https://github.com/aws-amplify/amplify-ios/issues/599 (configurable max depth)
 - https://github.com/aws-amplify/amplify-ios/issues/666 (configurable max depth)


richardmcclellan:
On Android, we have had [one similar request](https://github.com/aws-amplify/amplify-android/issues/577) to make selection set depth configurable.    Currently, we have this hardcoded to 2 for API and 1 for DataStore, but since the CLI asks for this as an input, perhaps we could use that value instead.

docs/lib/graphqlapi/fragments/ios/advanced-workflows.md
```diff
@@ -0,0 +1,225 @@
...
+        }
+        """
+        return GraphQLRequest<JSONValue>(document: document,
+                                         variables: ["id": id],
+                                         responseType: JSONValue.self)
+    }
+}
+
+```
+Query with `Amplify.API.query(request: .getCommentWithPost(byId: "[COMMENT_ID]"))`. Notice here that `JSONValue` is used as the `responseType`. `JSONValue` is utility type that can be used to represent an arbitrary JSON response.
```
richardmcclellan, line 145: Is there not a way for the customer to get a strongly typed response (i.e. a Post object and a list of Comment objects)?  For Android, the deserializer supports this, with an unlimited depth.

docs/lib/graphqlapi/fragments/ios/advanced-workflows.md
```diff
@@ -0,0 +1,225 @@
...
+        }
+        """
+        return GraphQLRequest<JSONValue>(document: document,
+                                         variables: ["postId": postId,
+                                                     "todoId": todoId],
+                                         responseType: JSONValue.self)
+    }
+}
+```
+Once you have the response data in a `JSONValue`, you can access each object in the JSON structure by encoding it back to Data and decoding it to the expected Model.
```
richardmcclellan, line 173: For Android, there is not a way to do this currently.   The deserializer will throw an exception if there is more than one element in the `data` object.    To add support for this, we could either:
1. If `data` has one element, deserialize into a `responseType` object (like we already do), but if more than one, deserialize into a `Map<{operationName}, {responseType}>`.
2. Add support for `decodePath` similar to iOS.

I'm inclined towards #1 because I'd prefer to avoid requiring customers to specify a `decodePath`.  I'm open to thoughts from you, because I think we should have platform parity.

docs/lib/graphqlapi/fragments/ios/advanced-workflows.md
```diff
@@ -0,0 +1,225 @@
...
+        case .failure(let errorResponse):
+            print("Response contained errors: \(errorResponse)")
+        }
+    case .failure(let apiError):
+        print("Failed with error: \(apiError)")
+    }
+}
+```
+
+If you have custom models or when Model have required fields which you have decided not to include in the response, you can create a `Codable` that conforms to the structure of the response data that you expect. From the previous example, the `Codable` can look like this
```
richardmcclellan, line 200:
```suggestion
If you have custom models or your Model has required fields that you have decided not to include in the response, you can create a `Codable` that conforms to the structure of the response data that you expect. From the previous example, the `Codable` would look like this
```

docs/lib/graphqlapi/fragments/ios/advanced-workflows.md
```diff
@@ -0,0 +1,225 @@
+This section describes different use cases for constructing your own custom GraphQL requests and how to approach it. You may want to construct your own GraphQL request if you want to
+- retrieve only a subset of the data to reduce data transfer
+- retrieve nested objects at a depth that you choose
+- combine multiple operations into a single request
+- decode the GraphQL response to a format of your choice
```
richardmcclellan, line 5: Since Amplify is an opinionated framework, we should avoid giving customers a "choice" of response format.    I do get that you need to explain what `responseType` and `decodePath` should be, but I would just remove this bullet point.

lawmicha:
> On Android, we have had [one similar request](https://github.com/aws-amplify/amplify-android/issues/577) to make selection set depth configurable. Currently, we have this hardcoded to 2 for API and 1 for DataStore, but since the CLI asks for this as an input, perhaps we could use that value instead.

Thanks, I've opened an issue to track this difference https://github.com/aws-amplify/amplify-ios/issues/681 I think introducing a walk-depth parameter from the API would achieve something like this and have a default walk depth tha is the same across both platforms. I'm not sure how we would tie in the number from the CLI question to the code at runtime since that number is currently used for the old code generated file for AppSync SDK

lawmicha:
> Sounds like there are 3 use cases this documentation is looking to address:
> 
> 1. Specific a subset of a selection set to reduce network response size.
> 2. Specify max depth for selection set to request.
> 3. Multiple operations in one request.
> 
> Customers can do all 3 of these today on Android by creating a `GraphQLRequest` object from a document `String`, a `Map` of `variables`, the `responseType`, and a `variablesSerializer`.
> 
> #3 will currently fail to deserialize the response though because we have an assumption that `data` will only contain a single element, and throw an exception if not. I suggested two solutions in another comment.
> 
> What's the reasoning for suggesting to use `JSONValue` for the `responseType`? On Android, we do support using a `String` as the `responseType`, though I'd prefer to avoid pushing customers in that direction, and haven't identified any use cases where we can't be smart and just deserialize to some sensible types.

In this doc, the suggestion for JSONValue was to help with debugging, and the final code should just deserialize to the type so that it's efficient. For example, to create a custom type, i have to visually see what the response structure looks like, so starting off with JSONValue or String and printing that out helps with that. If using String or JSONValue does the same job, then we can simplfy the example and just deserialize to String and remove the example of turning it back to Data and back to a Model

docs/lib/graphqlapi/fragments/ios/advanced-workflows.md
lawmicha: The issue is first, the Model-to-GraphQLRequest builder does not generate the selection set of the comments so data is not returned. Even if we default to a depth of 2 and do add the selection set for comments, then the second issue is when deserializing it doesn't match the response data structure. The response data contains the post fields, followed by the "items" key, and inside "items" is the array of Comments. For a successfull deserialization, the type should be 
````
Post {
  var items: [Comment]
}
``` 
instead of 
Post {
  var comments: [Comment]
}
```
Does Android check for "items" and do some custom deserialization into the Post's list of comments model? Even if it does do a default selection set of depth 2, the list of comments should be paginated, how does the developer get the next x number of comments for that post?

Comment (richardmcclellan) [reply docs/lib/graphqlapi/fragments/ios/advanced-workflows.md:145]: > The issue is first, the Model-to-GraphQLRequest builder does not generate the selection set of the comments so data is not returned. 

On Android, this is not an issue since we default to a depth of 2, for the API category.    To change the depth, all we need to do is expose a way to provide that constant.

> 
> Does Android check for "items" and do some custom deserialization into the Post's list of comments model?

Yes.  The expected type is `List<Comment>`, which is processed by the [`IterableDeserializer`](  https://github.com/aws-amplify/amplify-android/blob/27503d7a64c014e39eea4062d1962fb5b5128a27/aws-api/src/main/java/com/amplifyframework/api/aws/IterableDeserializer.java).  If the `JsonElement` is a `JsonObject`, rather than a `JsonArray`, (it is, and it contains `items` and `nextToken`), then the deserializer parses the value of `items` as a `List<Comment>`.

> Even if it does do a default selection set of depth 2, the list of comments should be paginated, how does the developer get the next x number of comments for that post?

Great question - Android doesn't support pagination on nested models, since we discard the pagination details during deserialization.   It is a limitation, but I don't think it's a huge deal because customers could still obtain a page of Comments by making a new request with Comments as the top level item, which does support pagination.

File: docs/lib/graphqlapi/fragments/ios/advanced-workflows.md (hunk 2 above)
Comment (lawmicha) [reply docs/lib/graphqlapi/fragments/ios/advanced-workflows.md:173]: For 1.  I agree that when data only has one element, then the developer should be able to just specify the response type without having to set the decodePath to the operationName. the `decodePath` parameter isn't very useful when the data contains arrays, then you won't be able to go further than the objects before that, for example,
```
data {
  getTodo {
    id: 111
    items [
       {  id: 222 }, { id: 333 }
    ]
  }
  getPost {
    id: 444
  }
}
```
decodePath of "getTodo.items" is the furthest you can go since the value at "items" is an array of objects


when there is more than one, can we take no opinion on deserialization and just deserialize to whatever the response is at "data"? this will be an object like
```
ResponseType {
  operationName1: ResponseType1
  operationName2: ResponseType2
}
```
and would map to the same response structure that the developer can see if they use the console or tried deserialization to String and printed it out

File: docs/lib/graphqlapi/fragments/ios/advanced-workflows.md [hunk 5]
```diff
@@ -0,0 +1,225 @@
...
+
+A GraphQL request is automatically generated for you when using AWSAPIPlugin with the existing workflow. For example, if you have a Todo model, a mutation request to save the Todo will look like this:
+```swift
+let todo = Todo(name: "my first todo", description: "todo description")
+Amplify.API.mutate(request: .create(todo))
+```
+Underneath the covers, a request is generated with a GraphQL document and variables and sent to the AppSync service. 
+
+```json
+// GraphQL document
```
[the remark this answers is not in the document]
lawmicha: sure, updating

docs/lib/graphqlapi/fragments/ios/advanced-workflows.md
lawmicha: removing

docs/lib/graphqlapi/fragments/ios/advanced-workflows.md
richardmcclellan: Ahh, that's right - #1 isn't actually possible since we would need the customer to provide the expected `responseType` for each operation, and currently, the GraphQLRequest constructor only asks for one `responseType`.

So, there's no way to support multiple operations with Android, currently. 

I know there has been at least one customer request for it, so I will keep thinking about the best way to support this, but for now, I think we should just say it is not supported in the docs.

lawmicha: on iOS, we can first get parity in behavior: "if there is only one element at data, deserialize to it" as API behavior change, this will allow the documentation to be updated to omit the decodePath. 

for elements which is under "data.items", we'll need custom deserialization logic in the plugin code to support this like Android has. 

until both of these changes happen, APIPlugin on iOS requires the decodePath to be set for both scenarios
1. getting element under "data"
2. getting list of elements under "data.items" for list queries

lawmicha: This PR on iOS should address the decodePath parity and then I can update the docs here to remove the `decodePath` examples https://github.com/aws-amplify/amplify-ios/pull/714

docs/lib/graphqlapi/fragments/ios/advanced-workflows.md
```diff
@@ -0,0 +1,225 @@
...
+        }
+        """
+        return GraphQLRequest<Todo>(document: document,
+                                    variables: ["id": id],
+                                    responseType: Todo.self,
+                                    decodePath: operationName)
+    }
+}
+```
+The decode path specifies which part of the response to deserialize to the `responseType`. You'll need to specify the operation name to deserialize the object at "data.getTodo" successfully into a Todo model.
```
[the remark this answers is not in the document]
lawmicha: thanks for the insight, just got something working on iOS for a PaginatedList: https://github.com/aws-amplify/amplify-ios/pull/715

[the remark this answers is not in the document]
lawmicha: In terms of the decodePath (https://github.com/aws-amplify/amplify-ios/pull/714) we do want the change to have a default behavior of assuming data in an object with single child, however this does break the API behavior for existing customers (see 3rd point in customer impact in PR), so we decided to hold off on releasing this in 1.x.x versions. What i've done in this docs PR is to consolidate the two sections, and remove decodePath from the Android sections `20_custom.md` in iOS contains decodePath and not in Android

docs/lib/graphqlapi/fragments/ios/advanced-workflows.md
lawmicha: So it is clear that Android and IOS contains different walk-depths by default, and it will take some major verions of the API to change this behavior. What i've done here is for the "nested data" section, instead of showcasing a walk-depth of two (this would be odd for Android since the default is already 2), it's showcasing the scenario where there's two nested objects and developer only wants to retrieve one of them. Post contains list of comments and and list of notes and only wants to get post with comments, omitting the notes

docs/lib/graphqlapi/fragments/ios/advanced-workflows.md
lawmicha: getting list of elements under "data.items" has been merged in, however decodePath will go in later

PR #2166, opened 2020-07-30
In Getting Stared guide, clarify Android project integration

KristinaPeterson:
@mohitsriv Minor updates to clarify instructions for a beginner/first-time customer.

*Issue #, if available:*

*Description of changes:*
In Configure Amplify and DataStore:
Clarification on step #1 
Added new step between #5 and #6 with information to add to the manifest node
Added more detail to step #7

In Query Todos:
Step #4 small language edit


docs/start/getting-started/fragments/android/integrate.md
```diff
@@ -3,6 +3,7 @@ Next you'll use the generated model to read and write data. In this section you'
 ## Configure Amplify and DataStore
 
 First, we'll add the DataStore plugin and configure Amplify by creating an Application class and overriding the `onCreate()` method.
+1. Navigate to the **example.todo** folder located at **ToDo** > **app** > **src** > **main** > **java/kotlin** > **com** > **example.todo**
```
jamesonwilliams, line 6: The `.` in `example.todo`:

This _probably is_ a valid instruction, if the user is navigating a UI pane in Android Studio. Android Studio will break from listing _directory structure_, and instead list _Java packages_ in some cases. In Android projects, directory structure and java package identifiers match, except for `/` being written as `.`, instead.

Not sure if this is worth clarifying, or how best to do it (without getting into trivialities.)

TLDR: `example.todo` is actually a directory path itself, `**example** > **todo**`.

PR #2170, opened 2020-07-31
Set deugaring flag to true for API and DataStore

raphkim:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-android/issues/687

*Description of changes:*
Add instructions to set desugaring flag to `true` for GraphQL API and DataStore categories, which are dependent on `java.time` package.

Note: `java.time` package is actually used in the `core` library, but only API and DataStore categories use the `Temporal` class extensively.


docs/lib/graphqlapi/fragments/android/getting-started.md
```diff
@@ -112,6 +112,9 @@ Also in your **app** `build.gradle`, add this piece of code to support the Java
 ```groovy
 android {
   compileOptions {
+    // Enable support for the new language APIs
+    coreLibraryDesugaringEnabled true
+    // Enable Java 8 features
     sourceCompatibility JavaVersion.VERSION_1_8
     targetCompatibility JavaVersion.VERSION_1_8
   }
```
Diederikjh, line 120: This block also needs to be added, otherwise the build will fail (after end of android block):
```
project.dependencies {
    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.0.10'
}

```

jamesonwilliams: +1, if we do need de-sugaring in the downstream project, we should also document this required dependency.

Diederikjh:
Maybe I added confusion by saying the build fails without the `project.dependencies` block.  If both  `project.dependencies` and `coreLibraryDesugaringEnabled true` are commented out, then my application (that contains generated amplify model files) builds fine.  The application then just crashes at runtime when an instance of AWSDateTime is being instantiated on an Android device < 26. 

Partial stack trace:
```
java.lang.NoClassDefFoundError: Failed resolution of: Ljava/time/ZoneOffset;
        at com.amplifyframework.core.model.temporal.Temporal$DateTime.<init>(Temporal.java:199)
```

If the application gradle file only has the `coreLibraryDesugaringEnabled true` line set, but not the `project.dependencies` block, then the build fails with 

> Cause: coreLibraryDesugaring configuration contains no dependencies. If you intend to enable core library desugaring, please add dependencies to coreLibraryDesugaring configuration.

Or am I missing something else?

jamesonwilliams:
> Let's validate the contents of the `Temporal.class`, as it exists in our 1.1.1 release `.aar`s, by unzipping the aar, the jar inside of it, and then running `javap` on the classfile.

I performed the following exercise:
```sh
mkdir explode
cd explode
curl https://repo1.maven.org/maven2/com/amplifyframework/aws-api-appsync/1.1.1/aws-api-appsync-1.1.1.aar -o aws-api-appsync-1.1.1.aar
unzip *.aar
unzip *.jar
temps=`find . -name "*Tempor*"`
for t in $temps; do javap -v $t > ${t}.javap; done 
find . -name "*.javap" -exec grep java\/time -l {} + 
```

In the 1.1.1 release of `aws-api-appsync`, the following to contain references to `java.time`:
1. `Temporal.Time` 
2. `Temporal.DateTime`
3. `Temporal.Date`

So, we either need to run r8 over the library artifacts, for the next release, or document a requirement that the _consumer_ run de-sugaring, as proposed in this PR.

PR #2176, opened 2020-08-02
Better naming of PostEditor

tobiasfeil:
I think it makes more sense to use a noun that describes the actual relationship. Currently, a noun is used that clearly describes the user and not the post, which is asymmetric as opposed to the symmetric nature of the relationship.

*Issue #, if available:*

*Description of changes:*


sw-yx:
appreciate the thoughts but my 2c is this is superficial enough to not matter 🤷‍♂️

tobiasfeil:
Haha yep I guess it's pretty minor, can be closed if you want

PR #2186, opened 2020-08-03
Add Angular and Vue examples

gsans:
*Issue #, if available:*
NA

*Description of changes:*
Add Angular and Vue examples to DataStore section.

### DIFF
--- a/docs/lib/datastore/fragments/js/examples.md
+++ b/docs/lib/datastore/fragments/js/examples.md
@@ -1,6 +1,7 @@
-## Example React application
+## Example application
 
-`App.js`:
+<amplify-block-switcher>
+<amplify-block name="React">
 
 ```jsx
 import React, { useEffect } from "react";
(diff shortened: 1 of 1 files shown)


docs/lib/datastore/fragments/js/examples.md
```diff
@@ -1,6 +1,7 @@
-## Example React application
```
sammartinez, line 1: I guess im confused, why is this being removed ? I do still see this fits the React application below that follows.

docs/lib/datastore/fragments/js/examples.md
```diff
@@ -75,10 +76,126 @@ function App() {
 
 export default App;
 ```
+</amplify-block>
+<amplify-block name="Angular">
 
-## Example React Native application
```
sammartinez, line 79: Since the below is an Angular example, can you look to add a title from the above here?

docs/lib/datastore/fragments/js/examples.md
```diff
@@ -75,10 +76,126 @@ function App() {
...
+      DataStore.delete(Post, Predicates.ALL);
+    }
+  }
+};
+</script>
+```
+</amplify-block>
 
-`App.js`:
+<amplify-block name="React Native">
```
sammartinez, line 198: Same callout as above, unless im missing something we should have the callouts of which example is what for our customers to not confuse them.

gsans:
This is because I'm using the amplify switcher control which lets you pick the language for the example. 

After adding this control some texts have to be removed as they don't apply now.

sammartinez:
> This is because I'm using the amplify switcher control which lets you pick the language for the example.
> 
> After adding this control some texts have to be removed as they don't apply now.

I figured as much, still waking up with coffee on my side so thank you for answering my questions :) Do you have screenshots of the switcher locally?

gsans:
There's not much to this change really. Just added the switcher

Screenshot 2020-08-05 at 17 05 25
Screenshot 2020-08-05 at 17 05 31
Screenshot 2020-08-05 at 17 05 37

PR #2218, opened 2020-08-10
Add Auth Overview for iOS & Android

mauerbac:
Exposing the auth overview used for JavaScript to be viewable for iOS and Android. The current auth overview is library agnostic and should be viewable for iOS and Android devs.


jpignata:
> The current auth overview is library agnostic and should be viewable for iOS and Android devs.

Is it? Doesn't it include [this](https://docs.amplify.aws/lib/auth/overview/q/platform/js#accessing-aws-services), which is very much not?

mauerbac:
Updated a version for iOS/Android with corresponding function signatures

docs/lib/auth/fragments/common/overview.md
```diff
@@ -0,0 +1,39 @@
...
+1. Clients pass the tokens to the backend that perform custom logic to allow or deny actions
+1. Clients sign the requests and the backend validates the signature, allowing or denying actions depending on predefined policy. The predefined rules are known as [IAM](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies.html) policies and automatically configured by the Amplify CLI.
+	
+The first mode is a common authorization method for REST or GraphQL APIs, while the second mode is necessary for interfacing with AWS services such as S3, Pinpoint, Sumerian, and others. 
+
+## Sign-up and sign-in
+
+For many apps, user sign-up and sign-in is all that is required. Once authenticated the app can talk to an API to access and mutate data. In this case, you can simply create a User Pool by running `amplify add auth` using the Amplify CLI and selecting the default setup. In your application you can use `Amplify.Auth.signUp` and `Amplify.Auth.signIn` (or an Amplify UI component) to complete this process and retrieve tokens. The Amplify client will refresh the tokens calling `Amplify.Auth.fetchAuthSession` if they are no longer valid. 
+	
+![Image](~/images/SimpleAuthZ.png)
```
jpignata, line 24: Can you add some copy here (and everywhere the 'image' placeholder is) for accessibility purposes?

docs/lib/auth/fragments/common/overview.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Social Provider Federation
+	
+Many apps also support login with social providers such as Facebook, Google Sign-In, or Login With Amazon. The preferred way to do this is via an OAuth redirect which lets users login using their social media account and a corresponding user is created in User Pools. With this design you do not need to include an SDK for the social provider in your app. Set this up by running `amplify add auth` and selecting the social provider  option. Upon completion you can use `Amplify.Auth.signInWithWebUI()` in your application to show a pre-built "Hosted UI". Or pass in a social provider name (e.g. `Amplify.Auth.signInWithSocialWebUI(AuthProvider.facebook()`) to interface directly and build out your own UI.
+	
+![Image](~/images/SocialAuthZ.png)
+	
+You can also get credentials directly from Identity Pools by passing tokens from a provider directly to `Amplify.Auth.fetchAuthSession`. However you will have to use that provider's SDK directly in your app and manage token refresh and auth flows manually.
```
royjit, line 32: This statement is wrong. We do not currently support passing a token from an identity provider to Cognito in Amplify. It is only available through escape hatch.

docs/lib/auth/fragments/common/overview.md
royjit, line 24: We had a discussion with @TrekSoft regarding this. This image does not mean anything or add any value to the doc, we should provide some good illustration which is useful. I prefer to remove this image.

docs/lib/auth/fragments/common/overview.md
```diff
@@ -0,0 +1,39 @@
...
+	
+
+## Accessing AWS services
+
+Some apps need to use AWS services which require [signing requests](https://docs.aws.amazon.com/general/latest/gr/signing_aws_api_requests.html). Examples of this would be storing images or videos on S3, or sending analytics to Pinpoint or Kinesis. Amplify automatically signs requests with short term credentials from a Cognito Identity Pool which automatically expire, rotate, and refresh by the Amplify client libraries. Setting up your backend with `amplify add auth` and calling `Amplify.Auth.signIn` will automatically do this for you as well after the client authenticates. The diagram below shows how JWT tokens are returned from User Pools and AWS credentials from Identity Pools. You can access these at any time with `Amplify.Auth.fetchAuthSession`.
+	
+![Image](~/images/AWSAuthZ.png)
```
royjit, line 39: Same comment here.. "The diagram below shows how JWT tokens are returned from User Pools and AWS credentials from Identity Pools" -> but sadly looking at the image, we do not get an idea on what is happening.

PR #2221, opened 2020-08-10
Updated Authenticator docs with methods & enums section, layout info …

dabit3:
…for customizing css


docs/ui/auth/fragments/web/authenticator.md
```diff
@@ -868,6 +869,67 @@ export withAuthenticator(App, {initialAuthState: 'signup'});
...
 </docs-filter>
 
+## Methods & Enums
+
+### AuthState
+
+`AuthState` is an enum with the following values:
+
+```js
+enum AuthState {
```
jamesonwilliams, line 879: Would it be beneficial to document _when_ each one of these is achieved, and how each one transitions to the others?

docs/ui/customization/layout.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Layout
+description: Managing the layout of UI Components
+filterKey: framework
+---
+
+### Managing Layout with CSS
```
jordanranz, line 7: The top-level customization section is meant for component-agnostic customization documentation. Instead of using "layout" here, maybe we can just highlight how you can select the top level tag to customize css. Similar to Ionic docs: https://ionicframework.com/docs/theming/css-variables#component-variables

jordanranz:
Awesome addition, small comment. thank you @dabit3

docs/ui/customization/layout.md
dabit3: Ok cool, so I can remove the `filter` field here?

dabit3: Ah actually I see, instead of calling it "layout" it can be something like "Component Variables"?

Issue #2220, opened 2020-08-10
[Todo] Document Hub listener for DataStore to know if an item is synced or not


sw-yx:
**Is your feature request related to a problem? Please describe.**

document this https://github.com/aws-amplify/amplify-js/issues/4808#issuecomment-640574885

related issue https://github.com/aws-amplify/amplify-js/issues/6534#issuecomment-671385854

**Describe the solution you'd like**

the actual solution i would like is to be able to check this inline without setting up the hub listener.

**Describe alternatives you've considered**
n/a

PR #2223, opened 2020-08-10
Documented datastore events and methods

dabit3:
Documented DataStore events and methods that were previously not documented.

A couple of other minor fixes.

closes https://github.com/aws-amplify/docs/issues/2220

### DIFF
--- a/docs/lib/datastore/fragments/js/other-methods.md
+++ b/docs/lib/datastore/fragments/js/other-methods.md
@@ -8,6 +8,12 @@ import { DataStore } from '@aws-amplify/datastore';
 await DataStore.clear();
 ```
 
+<amplify-callout>
+
+If your app is has authentication implemented, it is recommended to call `DataStore.clear()` on signin/signout to remove any user-specific data. This method is often important to use for shared device scenarios or where you need to purge the local on-device storage of records for security/privacy concerns.
+
+</amplify-callout>
+
 ## Start
 
 To manually start the sync process, use the `start` method:
@@ -17,3 +23,5 @@ import { DataStore } from '@aws-amplify/datastore';
 
 await DataStore.start();
 ```
+
+Synchronization starts automatically whenever you run `DataStore.query()` however you can explicitly begin the process with `DataStore.start()`.
--- a/docs/lib/datastore/fragments/js/other-methods.md
+++ b/docs/lib/datastore/fragments/js/other-methods.md
@@ -1,0 +1,19 @@
+## Clear
+
+To clear local data from DataStore, use the `clear` method:
+
+```js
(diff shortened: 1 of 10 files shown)


sw-yx:
in this commit i added a couple of payload illustrations from here https://github.com/aws-amplify/amplify-js/issues/4808#issuecomment-640574885, what do you think?

dabit3:
@sw-yx nice!! Yes that's really helpful.

undefobj:
> in this commit i added a couple of payload illustrations from here [aws-amplify/amplify-js#4808 (comment)](https://github.com/aws-amplify/amplify-js/issues/4808#issuecomment-640574885), what do you think?
> 
> what i'd really, REALLY like to write up is sample code to mark clientside data tentative, UNTIL it is synced. i guess since i want it that badly, it's on me to write. i'll assign that to myself.

Should we also put these on the Hub page?

docs/lib/datastore/fragments/js/other-methods.md
```diff
@@ -0,0 +1,19 @@
+## Clear
+
+To clear local data from DataStore, use the `clear` method:
+
+```js
```
undefobj, line 5: I would put a note about `.clear()` being important to use for shared device scenarios or where you need to purge the local on-device storage of records for security/privacy concerns.

docs/lib/datastore/fragments/js/other-methods.md
```diff
@@ -0,0 +1,19 @@
...
+import { DataStore } from '@aws-amplify/datastore';
+
+await DataStore.clear();
+```
+
+## Start
+
+To manually start the sync process, use the `start` method:
```
undefobj, line 13: We might want something like "_Synchronization starts automatically whenever you run `DataStore.query()` however you can explicitly begin the process with `DataStore.start()`_"

@manueliglesias can you confirm this behavior is still the case?

manueliglesias: Yes, that's right.

dabit3: cc @manueliglesias I'm adding this line into the PR, if this is no longer the case please let me know so I can update or remove it from the docs:

```md
Synchronization starts automatically whenever you run `DataStore.query()` however you can explicitly begin the process with `DataStore.start()`.
```

manueliglesias: @dabit3 Sounds good! At some point it would be good to show how to wait for the sync process to finish (using Hub) because if your first-ever operation is a query, you might get back an empty result if the sync hasn't fetched enough data t that point in time.

PR #2271, opened 2020-08-20
Add Flutter Warning to Guides

fjnoyp:
Flutter Preview has no support for API so we want to warn developers not to read the flutter docs contained in Guides.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/guides/fragments/flutter/flutterAPIWarning.md
+++ b/docs/guides/fragments/flutter/flutterAPIWarning.md
@@ -1,0 +1,8 @@
+
+<amplify-callout warning>
+
+### The API Category is not yet implemented in Flutter Preview.
+
+#### This section will be updated once the category becomes available.
+
+</amplify-callout>
--- a/docs/guides/fragments/flutter/flutterFuncWarning.md
+++ b/docs/guides/fragments/flutter/flutterFuncWarning.md
@@ -1,0 +1,8 @@
+
+<amplify-callout warning>
+
+### The Functions Category is not yet implemented in Flutter Preview.
+
+#### This section will be updated once the category becomes available.
+
+</amplify-callout>
--- a/docs/guides/functions/connecting-a-rest-api.md
+++ b/docs/guides/functions/connecting-a-rest-api.md
@@ -3,6 +3,8 @@ title: Connecting a REST API to a Lambda function
 description: How to connect a REST API to a Lambda function
 ---
 
+<inline-fragment platform="flutter" src="~/guides/fragments/flutter/flutterFuncWarning.md"></inline-fragment>
+
 In this guide you will learn how to connect a REST API to an existing Lambda function.
 
 To get started, create a new API:
--- a/docs/lib/storage/fragments/flutter/download.md
+++ b/docs/lib/storage/fragments/flutter/download.md
@@ -10,3 +10,20 @@ try {
   print(e.toString());
 }
 ```
+
+### Generate a download URL
+
+You can also retrieve a URL for the object in storage:
+
+```dart
+
+try {
+  GetUrlResult result =
+    await Amplify.Storage.getUrl(key: "myKey");
+  print(result.url); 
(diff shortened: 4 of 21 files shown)


docs/guides/functions/connecting-a-rest-api.md
```diff
@@ -3,6 +3,8 @@ title: Connecting a REST API to a Lambda function
 description: How to connect a REST API to a Lambda function
 ---
 
+<inline-fragment platform="flutter" src="~/guides/fragments/flutter/flutterWarning.md"></inline-fragment>
```
haverchuck, line 6: I believe that the content under guides/functions is NOT API category, but the functions category.

docs/guides/fragments/flutter/flutterWarning.md
```diff
@@ -0,0 +1,8 @@
+
+<amplify-callout warning>
+
+### The API Category is not yet implemented in Flutter Preview.
+
+#### Please ignore the Flutter Guides section for now.
```
haverchuck, line 6: I would suggest some other than 'Please ignore...'.  Maybe something like "This section will be updated once the category becomes available."

docs/lib/storage/fragments/flutter/download.md
```diff
@@ -10,3 +10,20 @@ try {
...
+
+### Generate a download URL
+
+You can also retrieve a URL for the object in storage:
+
+```dart
+
+try {
+  GetUrlResult result =
+    await Amplify.Storage.getUrl(key: "myKey", options: options);
```
Ashish-Nanda, line 22: We should not pass an options param in the API call if we didn't create the object itself.

docs/lib/storage/fragments/flutter/download.md
```diff
@@ -10,3 +10,20 @@ try {
   print(e.toString());
 }
 ```
+
+### Generate a download URL
```
jpignata, line 14: Does this belong in this pull?

mlabieniec: Yes one additional update to storage we discussed

fjnoyp: yes we wanted a minor update to our Storage docs to include GetURL

docs/lib/storage/fragments/flutter/download.md
```diff
@@ -10,3 +10,20 @@ try {
...
+
+You can also retrieve a URL for the object in storage:
+
+```dart
+
+try {
+  GetUrlResult result =
+    await Amplify.Storage.getUrl(key: "myKey");
+  let url = result.url;
```
Ashish-Nanda, line 23: Lets remove this line and instead print `result.url`?

fjnoyp: ok

docs/guides/fragments/flutter/flutterAPIWarning.md
```diff
@@ -0,0 +1,8 @@
+
+<amplify-callout warning>
+
+### The API Category is not yet implemented in Flutter Preview.
```
jpignata, line 4: A nit, but can we make this look like the other call outs? Here's how it renders vs. how others do. We shouldn't use headers which have a semantic meaning for this.

**Current**
CleanShot 2020-08-20 at 13 15 38

**Other**
CleanShot 2020-08-20 at 13 15 49

docs/guides/fragments/flutter/flutterFuncWarning.md
```diff
@@ -0,0 +1,8 @@
+
+<amplify-callout warning>
+
+### The Functions Category is not yet implemented in Flutter Preview.
```
jpignata, line 4: I believe all of these guides relate to the API and Auth categories, right? This copy seems wrong. I'd just remove the warning as these guides are still applicable.

Issue #2252, opened 2020-07-23
Amplify configure command overwrites existing default profile


gsans:
**Describe the bug**
After running amplify configure the default profile has been overwritten.

**Amplify CLI Version**
4.24.1

**To Reproduce**
`amplify configure`

**Expected behavior**
The default profile should be preserved for better interoperability with other libraries sharing profile information. See extract below from user.

**Screenshots**
NA

**Desktop (please complete the following information):**
OS: OS
Node Version: 14.5.0

**Additional context**
Issue reported during Amplify Days workshop

From Discord:
Guessing it would be better to maintain a separate profile in my ~/.aws/config file for Amplify. We're using us-west-1 for most things with our app so just noticed I overwrote my default config with us-west-2 :man_facepalming: should have been obvious to me

offlineprogrammer:
Closed by the PR 2329

PR #2329, opened 2020-08-28
Explain how named profile is stored by the Amplify CLI

UnleashedMind:
*Issue #, if available:*
#2252 

*Description of changes:*


docs/cli/start/install.md
```diff
@@ -70,7 +70,9 @@ This would update/create the AWS Profile in your local machine
 
 Successfully set up the new user.
 ```
+The Amplify CLI uses the shared AWS [configuration and credentials files](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html) to store the profile information. Once `amplify configure` finishes execution, the named profile will be update / created in the `~/.aws/config` and `~/.aws/credentials` files. 
 
+Please note that those two files are shared, other tools such as the AWS CLI use them as well. Updating a profile's credentials and region may cause the other tools to stop functioning properly. Use caution when specifying the profile name, if you already have the `default` profile set up in your local machine, enter a different profile name when prompted.
```
yuth, line 75:
```suggestion
<amplify-callout warning>
Please note that those two files are shared, other tools such as the AWS CLI use them as well. Updating a profile's credentials and region may cause the other tools to stop functioning properly. Use caution when specifying the profile name, if you already have the `default` profile set up in your local machine, enter a different profile name when prompted.
</amplify-callout>
```

PR #2339, opened 2020-08-31
feat(android-api) update docs for new pagination approach added in #amplify-android/pull/750

richardmcclellan:
This simplifies the approach for getting paginated results from `Amplify.API` on Android.  All query responses are now parsed as `PaginatedResult<M>` objects.  Previously some were parsed as `Iterable<M>`, which hid the pagination details.   This will not cause any breaking changes because `PaginatedResult` now implements `Iterable`.  Customers can still access the `Iterable<M>` via `response.getData().getItems()`, but can now also access it via `response.getData()`.


jpignata:
@richardmcclellan Does this need an Rx sample?

richardmcclellan:
> @richardmcclellan Does this need an Rx sample?

Yes, that would be very good to have, partly because I believe both the Java/Kotlin approaches are susceptible to stack overflows.  I need to put together a quick POC to prove that it works, and then I can add the example.

PR #2349, opened 2020-09-02
feat(api) Android RxJava code samples for paginated list queries

richardmcclellan:


docs/lib/graphqlapi/fragments/android/query-data.md
```diff
@@ -139,5 +154,31 @@ fun query(request: GraphQLRequest<PaginatedResult<Todo>>) {
 }
 ```
 
+</amplify-block>
+
+<amplify-block name="RxJava">
+
+```java
```
jamesonwilliams, line 161: This example is a little intense! Do you think we should add some syntactic fluff to simplify it, into the Rx bindings? Like a version of `RxAmplify.query` that is paging-aware?

jamesonwilliams (approved):

docs/lib/graphqlapi/fragments/android/query-data.md
richardmcclellan: Ooh, that is a really good idea.  Then, we could also simply [this logic](https://github.com/aws-amplify/amplify-android/blob/df70332ab7f6197fe5756a85196f8505fa05e95d/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java#L216-L231) in the DataStore SyncProcessor to use it as well.

richardmcclellan: The above idea will require some work in the library itself.  Until that's done, I'm going to merge this PR, since this is the best solution currently for API pagination with RxJava.  Later, we can update the docs to the simpler version.

PR #2367, opened 2020-09-08
Add missing appsync Tag/Untag Resource actions

rob3c:
These missing `appsync:TagResource` and `appsync:UntagResource` actions caused a stack creation failure when following the AWS Amplify Angular getting started tutorial [here](https://docs.amplify.aws/start/getting-started/data-model/q/integration/angular#creating-the-api-with-database) at the `amplify push` step after creating the backend. (Actually just the `appsync:TagResource` one was reported as `CREATE_FAILED` in the console with status code 403 and error code `AccessDeniedException`, but I added both after similar pairs in the existing policy.)

I was able to successfully deploy via `amplify push` after updating the policy with them.


jkervine (approved):
I propose that this change is approved. Was just about to submit a PR about this myself.

mauerbac:
@aws-amplify/amplify-cli , can we get a review?

docs/cli/usage/iam.md
```diff
@@ -64,6 +64,8 @@ The Amplify CLI requires the below IAM policies for performing actions across al
                 "appsync:ListResolvers",
                 "appsync:DeleteType",
                 "appsync:GetType",
+                "appsync:TagResource",
```
kaustavghosh06, line 67: I believe this action is already present in the doc here - https://docs.amplify.aws/cli/usage/iam
The untag one is not present. @ammarkarachi do we need it?

ammarkarachi: Yes we do, let's do a review for other resources too and update it.

rob3c: It looks like TagResource was added after I made this PR 6 weeks ago. Is there anything I need to do as part of this PR thread, or are you all going to take over from here?

PR #2375, opened 2020-09-09
Update screenshot of configure project

changxu0306:
Fix wrong project name in previous configure project screenshot.
Remove a blank line in "initializing Amplify in Java" to make it consistent with that in Kotlin.


jamesonwilliams (approved):
Thanks @changxu0306!

richardmcclellan (changes requested):
It looks like the same image is used for [Getting Started](https://docs.amplify.aws/start/getting-started/setup/q/integration/android#create-a-new-android-application) and [Project Setup](https://docs.amplify.aws/lib/project-setup/create-application/q/platform/android), so this fixes Project Setup, but breaks Getting Started.

changxu0306:
> It looks like the same image is used for [Getting Started](https://docs.amplify.aws/start/getting-started/setup/q/integration/android#create-a-new-android-application) and [Project Setup](https://docs.amplify.aws/lib/project-setup/create-application/q/platform/android), so this fixes Project Setup, but breaks Getting Started.

Good catch! Will fix it then.

richardmcclellan:
> > It looks like the same image is used for [Getting Started](https://docs.amplify.aws/start/getting-started/setup/q/integration/android#create-a-new-android-application) and [Project Setup](https://docs.amplify.aws/lib/project-setup/create-application/q/platform/android), so this fixes Project Setup, but breaks Getting Started.
> 
> Good catch! Will fix it then.

Great!  I think we should just have separate images for each flow, rather than using the same one, that's how iOS handles it.

changxu0306:
> > > It looks like the same image is used for [Getting Started](https://docs.amplify.aws/start/getting-started/setup/q/integration/android#create-a-new-android-application) and [Project Setup](https://docs.amplify.aws/lib/project-setup/create-application/q/platform/android), so this fixes Project Setup, but breaks Getting Started.
> > 
> > 
> > Good catch! Will fix it then.
> 
> Great! I think we should just have separate images for each flow, rather than using the same one, that's how iOS handles it.

Yes. That's what I'm doing :-)

richardmcclellan (approved):

PR #2378, opened 2020-09-10
fix(amplify-flutter): signout callout

haverchuck:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-flutter/issues/93

*Description of changes:*
This PR makes a note about signout behavior differences between iOS and Android platforms.


docs/lib/auth/fragments/flutter/signout/00_signout_callout.md
```diff
@@ -0,0 +1,3 @@
+<amplify-callout usage>
+Please note that the Android platform will allow a new user to sign in even if a different user has already signed in, while the iOS platform will force any currently authenticated user to signout before a different user can sign in.
```
Ashish-Nanda, line 2: Suggestion to change it to an action required on the developers part: "On iOS, if a user is signed in, you would have to explicitly call signOut before a new user can sign in."

TrekSoft: Also think it would be good to clarify that Android doesn't allow multiple users to be signed in at the same time but rather it replaces the previous sign in. Maybe something like this:

"Please note that to sign in a new user, you must first sign out of your existing account on iOS. On Android, this is not required and you can simply sign in the new user without first signing out."

Ashish-Nanda (approved):
Minor comment, otherwise LGTM.

docs/lib/auth/fragments/flutter/signout/00_signout_callout.md
haverchuck: @Ashish-Nanda I think @TrekSoft 's suggestion will satisfy yours as well.

mauerbac:
is this ready for merge?
@Ashish-Nanda

Ashish-Nanda (approved):
LGTM

mauerbac:
looking for final signoff @TrekSoft , @royjit , @drochetti or @mlabieniec

royjit:
I was just wondering why we cannot handle this in the Flutter side? Flutter could check if the user is signed in and then throw an error. This check can be removed once Android fixes the behavior. This way Flutter developer do not need to worry about the internals and will have the same behavior for iOS and Android. 

Also, I would like to see how this works in JS @elorzafe

TrekSoft:
Yeah I thought that was what we landed on as the decision.

PR #2393, opened 2020-09-15
hotfix: add missing next.svg

ericclemmons:
It was mistakenly placed in client/src/assets/assets/integrations/next.svg, which isn't versioned.

---


ashika01 (approved):
Lets get this in 👍

PR #2419, opened 2020-09-19
Fix angular example path

victorleungtw:
*Issue #, if available:*
The angular path is incorrect to import awsconfig

*Description of changes:*
For angular example, in app.module.ts, to reference awsconfig, the path would be '../aws-exports'; instead of './aws-exports';


wlee221 (approved):
Thanks for the catch, @victorleungtw 👍

PR #2436, opened 2020-09-23
Added TSC build instructions

DavidBrear:
I could not get the babel instructions working so I figured out how to get `tsc` to work then copied what I did here. I left the Babel instructions incase people still want that. I was working with a React Native typescript project so I already had `tsc` installed which I would imagine a lot of react / react native devs would as well.

*Issue #, if available:*

*Description of changes:*

I added instructions to get typescript code compiled using the `tsc` function as opposed to Babel.


DavidBrear:
Looks like this is failing because it doesn't know the word `esnext` is there a way to comment out those words so that they aren't going to break the spellchecker?

sw-yx (approved):

PR #2452, opened 2020-09-24
fix: Android DataStore ops that take predicate

jamesonwilliams:
The DataStore includes one query method which accepts a second parameter
of type `QueryOptions`. All other DataStore methods use only a
`QueryPredicate`.

The `Where.matches` utility method will convert a `QueryPredicate` into
a `QueryOptions`. This can be used as an adapter when dealing with the
`query(...)` method which accepts `QueryOptions`.

However, `save(...)` and `delete(...)` just take a predicate, itself.
So, we need to remove the `Where.matches(...)` bit from the
documentation. It doesn't compile.


TrekSoft (approved):
Cat8u3

changxu0306 (approved):

PR #2455, opened 2020-09-25
Fixed mutation example

TrekSoft:
Changed example from update to create since update requires an ID of an existing record to work...


jpignata (approved):
LGTM

PR #2486, opened 2020-10-02
Fix/vue auth state mgmt

jennapederson:
*Issue #, if available:* n/a

*Description of changes:*

These docs tripped me up a bit as I was trying to add copy for both authenticated and unauthenticated users.

- Removed the conditional from `amplify-authenticator` and nested the rest inside to be more consistent with other [doc examples](https://docs.amplify.aws/start/getting-started/auth/q/integration/vue#create-login-ui)
- Added a check for signed out to add additional context


mauerbac:
Thanks @jennapederson ! @aws-amplify/amplify-js , can we get a review?

amhinson (approved):
LGTM 👍

PR #2492, opened 2020-10-04
Use AWSDateTime as type for createdAt and updatedAt.

gotgenes:
*Issue #, if available:* N/A

*Description of changes:*

The documentation for the directive `@model` states:

>The automatically added `createdAt` and `updatedAt` fields can’t be set in create or update mutation. If these fields need to be controlled as part of the mutation, they should be in the input schema and should have `AWSDateTime` as their type.

This corrects examples where `createdAt` or `updatedAt` are specified as type `String!` to instead be of type `AWSDateTime!` per the statement above.

This also adds minor formatting improvements, and corrects an instance of `AWSDatetime` to `AWSDateTime` (note the capital T).

### DIFF
--- a/docs/cli/graphql-transformer/directives.md
+++ b/docs/cli/graphql-transformer/directives.md
@@ -74,7 +74,7 @@ type Post @model(queries: { get: "post" }, mutations: null, subscriptions: null)
 }
 ```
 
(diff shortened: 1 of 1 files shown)


mauerbac:
@aws-amplify/amplify-cli , can we get a review?

jamesonwilliams (approved):

PR #2520, opened 2020-10-09
chore: simplify android project setup guides

jamesonwilliams:
Amplify Android supports API level 16 and above. However, we can greatly
simplify the initial setup experience by assuming that the user will
target 21+. We keep a callout to guide the user towards setup for lower
API levels, if they do target them.

The user does not *really* need to add mavenCentral() as a dependency.
mavenCentral() is updated sooner when we relase, but the propagation
delay of jcenter() is probably suitable for most users. Particularly if
the tradeoff is more complexity in the getting started guide.

Lastly, the Amplify Core library is included as a transitive dependency
when including the API and DataStore plugins, so it doesn't need to be
spelled out explicitly in the dependencies block.

Resolved: https://github.com/aws-amplify/docs/issues/2555


TrekSoft (approved):
Cat14u4

raphkim (approved):

Issue #2512, opened 2020-10-07
[Feedback] Incorrect documentation. Storage.get does not directly return a Blob. 


ferdingler:
**Page**: [`/lib/storage/download/q/platform/js`](https://docs.amplify.aws/lib/storage/download/q/platform/js)

**Feedback**:

Documentation states that `Storage.get` returns a `Blob` of object data if download is true, which is incorrect. `Storage.get` returns an **object that contains a property `Body`** which is a `Blob`. That is different than saying that that the response of Storage.get is a Blob. 

Spent a few minutes debugging an issue because of this slight confusion.

sw-yx:
aye. that was my fault, I'm sorry about that. i currently have a PR pending review with a fix and even more info. https://github.com/aws-amplify/docs/pull/2521/files

sw-yx:
welp, that was fast, i think we fixed it

ferdingler:
Thanks for the quick fix!

PR #2521, opened 2020-10-09
(fix) add further instructions for downloading in Storage.get docs

sw-yx:
*Description of changes:* (fix) add further instructions for downloading in Storage.get docs

closes https://github.com/aws-amplify/docs/issues/2512


docs/lib/storage/fragments/js/download.md
```diff
@@ -23,12 +42,37 @@ Send object data for immediate file download:
...
+// data.Body is a Blob
 data.Body.text().then(string => { 
   // handle the String data return String 
 })
 ```
 
+You can programmatically download Blobs using JavaScript:
+
+```js
+// https://blog.logrocket.com/programmatic-file-downloads-in-the-browser-9a5186298d5c/
```
sammartinez, line 54: Is this ok to use ? Does the person that owns the blog understand we would use this? I don't want to recommend a blog link if it leads to the person will remove it. Can we look to have this remove?

sw-yx: done!

docs/lib/storage/fragments/js/download.md
```diff
@@ -23,12 +42,36 @@ Send object data for immediate file download:
...
+    }, 150);
+  };
+  a.addEventListener('click', clickHandler, false);
+  a.click();
+  return a;
+}
+
+// usage
+Storage.get(fileKey, { download: true })
+        .then(res => downloadBlob(res.Body, downloadFileName)) // can derive from fileKey if you wish
```
sw-yx, line 72:
```suggestion
        .then(res => downloadBlob(res.Body, downloadFileName)) // derive downloadFileName from fileKey if you wish
```

docs/lib/storage/fragments/js/download.md
```diff
@@ -14,7 +14,26 @@ await Storage.get(key: string, config: {
 
 `Storage.get` returns a signed URL `string` to your file if `download` is false, which is the default.
 
-`Storage.get` returns a [`Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob) of object data if `download` is true.
+`Storage.get` returns an object with a `Body` of [`Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob) of object data if `download` is true. The full return signature looks like this:
```
sw-yx, line 17:
```suggestion
If `download` is true, `Storage.get` returns an object with a `Body` field of type [`Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob). It is up to you to then save it to disk (we suggest a method below). 

The full return signature looks like this:
```

docs/lib/storage/fragments/js/download.md
```diff
@@ -12,9 +12,30 @@ await Storage.get(key: string, config: {
 })
 ```
 
-`Storage.get` returns a signed URL `string` to your file if `download` is false, which is the default.
-
-`Storage.get` returns a [`Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob) of object data if `download` is true.
+`Storage.get` merely returns a signed URL `string` to your file, if `download` is false, which is the default. You can use this to create a download link for people to click. This is usually the recommended option. Note that this method **does not check if the file actually exists** as that would involve an extra API call.
```
sammartinez, line 15:
```suggestion
`Storage.get` returns a signed URL `string` to your file, if `download` is false, which is the default. You can use this to create a download link for people to click on. This is our recommended option. Note that this method **does not check if the file actually exists** as that would involve an extra API call.
```

sammartinez (approved):
LGTM 🌮  Thanks for getting the updates done @sw-yx !

docs/lib/storage/fragments/js/download.md
```diff
@@ -12,25 +12,78 @@ await Storage.get(key: string, config: {
...
-Send object data for immediate file download:
+### File download option
+
+The `download` option send you the object data for download or programmatic manipulation:
 
 ```javascript
 const data = await Storage.get(`filename.txt`, { download: true })
 
-// data is a Blob
+// data.Body is a Blob
```
sw-yx, line 34: this was a mistake by me in my first PR to this doc, which led to https://github.com/aws-amplify/docs/issues/2512

docs/lib/storage/fragments/js/download.md
```diff
@@ -12,25 +12,78 @@ await Storage.get(key: string, config: {
 })
 ```
 
-`Storage.get` returns a signed URL `string` to your file if `download` is false, which is the default.
+`Storage.get` returns a signed URL `string` to your file, if `download` is false, which is the default. You can use this to create a download link for people to click on. This is our recommended option. Note that this method **does not check if the file actually exists** as that would involve an extra API call.
 
-`Storage.get` returns a [`Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob) of object data if `download` is true.
+```js
+// get the signed URL string
+const signedURL = await Storage.get(file.__data.key)
```
sw-yx, line 19:
```suggestion
const signedURL = await Storage.get(key) // get key from Storage.list
```

docs/lib/storage/fragments/js/download.md
```diff
@@ -12,25 +12,78 @@ await Storage.get(key: string, config: {
...
+  a.click();
+  return a;
+}
+
+// usage
+Storage.get(fileKey, { download: true })
+        .then(res => downloadBlob(res.Body, downloadFileName)) // derive downloadFileName from fileKey if you wish
+```
+
+The full return signature of `Storage.get(somekey, { download: true })` looks like this:
```
sw-yx, line 66:
```suggestion
The full return signature of `Storage.get(key, { download: true })` looks like this:
```

PR #2530, opened 2020-10-11
docs: remove unused CSS property in api.md

pobch:
*Description of changes:*

CSS property `flex: 1` do nothing on a flex container here. `flex` property should be used on a flex "item", not on the flex "container".


mauerbac:
@harrysolovay , is this a quick one you can take a look at?

harrysolovay (approved):

PR #2553, opened 2020-10-15
Update advanced.md

nicholaspuglia:
Federated types are in camel case rather than snake case. Found in FederatedConfig type:

`export interface FederatedConfig {
    auth0Config?: {
        audience?: string;
        clientID: string;
        domain: string;
        responseType: string;
        redirectUri: string;
        returnTo?: string;
        scope?: string;
    };
    amazonClientId?: string;
    facebookAppId?: string;
    googleClientId?: string;
    oauthConfig?: {
        [key: string]: any;
    };
}`


mauerbac:
@elorzafe @sammartinez , can we get a quick review?

mauerbac:
@elorzafe , can you review?

amhinson (commented):
It looks like snake case was correct with `aws-amplify-react` ([code](https://github.com/aws-amplify/amplify-js/blob/958f61ef2918728cc46b9b210d60e868edd87f39/packages/aws-amplify-react/src/Auth/FederatedSignIn.tsx#L150-L156)), but the new UI components use camel case ([code](https://github.com/aws-amplify/amplify-js/blob/958f61ef2918728cc46b9b210d60e868edd87f39/packages/amplify-ui-components/src/common/types/auth-types.ts#L24-L40)). 

However, I think this example in the docs would be better served to provide an example for Angular, Vue & React. Perhaps using the `<amplify-block-switcher>` component, i.e.:

```
<amplify-block-switcher>
  <amplify-block name="Angular">
    ...
  </amplify-block>
  <amplify-block name="Vue">
    ...
  </amplify-block>
  <amplify-block name="React">
    ...
  </amplify-block>
</amplify-block-switcher>
```

amhinson (commented):
At the least though, we could just update the import and usage:
```
import { AmplifyAuthenticator } from '@aws-amplify/ui-react';

...

<AmplifyAuthenticator federated={federated}>
```

PR #2599, opened 2020-10-28
Don't run tests on docs changes

jpignata:
*Issue #, if available:* n/a

*Description of changes:* Currently the entire test suite must run whenever docs change. This splits the workflow into two parts: spellcheck which runs on every pull request and test which ignores changes made in the docs directory.

### DIFF
--- a/.github/workflows/spellcheck.yml
+++ b/.github/workflows/spellcheck.yml
@@ -1,11 +1,8 @@
 name: Test
-on:
-  pull_request:
-    paths-ignore:
-      - 'docs/**'
+on: pull_request
 jobs:
-  test:
-    name: Test
+  spellcheck:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -1,5 +1,8 @@
 name: Test
```
sammartinez, line 1: Do we want to look to remove this? It seems like CircleCI is already handling the build step and this `test.yml` seems to be doing the same

sammartinez: Checking CircleCI, seems like its not doing anything, we can look to remove the CircleCi config then as we aren't doing anything with it at a later time.

sammartinez (approved):
LGTM 🌮

PR #2640, opened 2020-11-07
Fix issue with large response from graphql query

franrios:
*Description of changes:*
When Graphql response is large, resolving the promise on the first data chunk will cause an issue.  We have to wait for all the data chunks and then concatenate.


franrios:
@mauerbac

dabit3:
Hey @franrios , can you give some more insight into this update? I.e. do you think this will apply to the majority of use cases or is this an edge case?

franrios:
This applies to all the use cases. When listening to data through `result.on('data', ...` you will get the first chunk. This is fine if the response is small but it won't work for big ones. You need to save all the data chunks and the only way to know when all the chunks are received is by listening to the `end` event.

PR #2664, opened 2020-11-16
Updating docs with more descriptive information for 3rd party plugins

renebrandel:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/cli/plugins/plugins.md
+++ b/docs/cli/plugins/plugins.md
@@ -24,9 +24,10 @@ Plugins are explicitly managed in the Amplify CLI pluggable platform. Plugins en
 
 ## Third-party plugins
 
-* [amplify-category-video](https://www.npmjs.com/package/amplify-category-video)
-* [amplify-category-docs](https://www.npmjs.com/package/amplify-category-docs)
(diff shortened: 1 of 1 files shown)


mauerbac (approved):
lgtm

kaustavghosh06 (approved):

r0zar:
:+1: LGTM as well.

_On a side note..._

It would be nice to get some plugins listed that are not of the 'category' type.

I've been wondering if they could help solve for some of the asks from customers/internally, without having to make core code changes.

That being said, my understanding of their use cases will likely remain fuzzy until I've seen how they are used in a published module.

jamesonwilliams:
> It would be nice to get some plugins listed that are not of the 'category' type.
> 
> I've been wondering if they could help solve for some of the asks from customers/internally, without having to make core code changes.

For Android, iOS, and Flutter, anyway, that's the _only_ type of plugin that can be made. The contract gets defined in the core library, and then the plugin adds the business logic to carry out that contract.

It is not possible to create a plugin that augments/extends the API definition.

r0zar:
Interesting! Didn't know that. Thanks @jamesonwilliams

PR #2679, opened 2020-11-20
Clarify how to keep the file names for custom cf templates unique

toelke:
*Description of changes:*
The language on https://docs.amplify.aws/cli/usage/customcf is not clear how to keep the file names of custom templates unique. The code uses a glob here: https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-provider-awscloudformation/src/push-resources.js#L30


renebrandel (approved):

renebrandel:
I made a small edit in the wording for clarity. Thanks for catching this. cc @attilah to confirm accuracy as well.

PR #2683, opened 2020-11-20
Add documentation for clear, start and stop on Android

richardmcclellan:

### DIFF
--- a/docs/lib/datastore/other-methods.md
+++ b/docs/lib/datastore/other-methods.md
@@ -4,3 +4,4 @@ description: Other Amplify Datastore methods
 ---
 
 <inline-fragment platform="js" src="~/lib/datastore/fragments/js/other-methods.md"></inline-fragment>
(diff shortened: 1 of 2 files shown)


docs/lib/datastore/other-methods.md
```diff
@@ -4,3 +4,4 @@ description: Other Amplify Datastore methods
 ---
 
 <inline-fragment platform="js" src="~/lib/datastore/fragments/js/other-methods.md"></inline-fragment>
+<inline-fragment platform="android" src="~/lib/datastore/fragments/android/other-methods.md"></inline-fragment>
```
jpignata, line 7: Probably outside of the scope of this PR, but if the heading "Other methods" seems less than ideal to me.

jpignata (approved):
LGTM

docs/lib/datastore/other-methods.md
richardmcclellan: Yeah, I agree it's not the best.  I copied it from JS.  Maybe when iOS is ready to add their version and we can have one common file, we can also rethink the title then.

PR #2698, opened 2020-11-25
Adobe Analytics integration v1

tjleing:
- Adds Shortbread integration for EU user cookie consent
- Adds Adobe Analytics integration
  - page load event (seems to come with referrer, so we don't need to set that up manually?)
  - external link event (also done by default)
  - docsearch events -- including search query and result count

Theoretically TODO: internal/external campaign query search parameter
Will be able to see analytics data, including hopefully a Re:Invent bump, only after these changes are merged in


jamesonwilliams (commented):
`s.eVar26`, `s.eVar27`, `s.tl`, `s.t` -- very cryptic. Are these minified references, or something?

tjleing:
> `s.eVar26`, `s.eVar27`, `s.tl`, `s.t` -- very cryptic. Are these minified references, or something?

Unfortunately we can't change them -- they're required by Adobe Analytics!  Not entirely sure why it's so cryptic except possibly to cut down on script and query string parameter length.

mauerbac (approved):
LGTM - @sammartinez / @harrysolovay , can you also review ?

client/src/docs-ui/page/page.tsx
```diff
@@ -26,7 +26,12 @@ import {
   SetNewSelectedTabHeadings,
 } from "./page.types";
 import {pageContext} from "./page.context";
-import {track, AnalyticsEventType} from "../../utils/track";
+import {
+  track,
```
sammartinez, line 30: So I see `trackPageVisit()` and `trackPageFetchException()` in use, where is `track` or the `AnalyticsEventType` used?

client/src/utils/track.ts
```diff
@@ -7,6 +7,14 @@ let configured = false;
 if (!configured) {
   Auth.configure(awsexports);
   Analytics.configure(awsexports);
+  // @ts-ignore
+  if (typeof AWSCShortbread !== "undefined") {
```
sammartinez, line 11: Is there an additional check that we want to place here? Does it turn into a function? Does it need to be: 
```suggestion
  if (typeof AWSCShortbread !== undefined) {
```

client/src/utils/track.ts
```diff
@@ -7,6 +7,14 @@ let configured = false;
...
+    // @ts-ignore
+    const shortbread = AWSCShortbread({
+      domain: ".amplify.aws",
+    });
+    shortbread.checkForCookieConsent();
```
sammartinez, line 16: Is shortbread used anywhere else within this file?

client/src/utils/track.ts
```diff
@@ -7,6 +7,14 @@ let configured = false;
 if (!configured) {
   Auth.configure(awsexports);
   Analytics.configure(awsexports);
+  // @ts-ignore
+  if (typeof AWSCShortbread !== "undefined") {
+    // @ts-ignore
+    const shortbread = AWSCShortbread({
```
sammartinez, line 13: Do we need to create a new instance as follows? 
```suggestion
    const shortbread = new AWSCShortbread({
```

client/src/utils/track.ts
```diff
@@ -65,3 +73,35 @@ export const track = (event: AnalyticsEvent): Promise<unknown> | undefined => {
     }
   }
 };
+
+export const trackPageVisit = (): void => {
+  if (Build.isBrowser) {
+    // @ts-ignore
+    s.t();
```
sammartinez, line 80: We can talk offline but what does this do?

client/src/utils/track.ts
```diff
@@ -65,3 +73,35 @@ export const track = (event: AnalyticsEvent): Promise<unknown> | undefined => {
...
+    s.eVar26 = query;
+  }
+};
+
+export const trackSearchResult = (resultCount: number): void => {
+  if (Build.isBrowser) {
+    // @ts-ignore
+    s.eVar27 = resultCount;
+    // @ts-ignore
+    s.events = resultCount === 0 ? "event1" : "event2";
```
sammartinez, line 103: what do these events mean?

docs/scripts/shortbreadv1.js
```diff
@@ -0,0 +1,2 @@
...x-c #awsccc-sb-a.awsccc-Rtl #awsccc-cb-c #awsccc-cb-title{padding-right:40px}}#awsccc-sb-ux-c #awsccc-sb-a a,#awsccc-sb-ux-c #awsccc-sb-a a>span,#awsccc-sb-ux-c #awsccc-sb-a a svg path{color:#0073bb;text-decoration:none}#awsccc-sb-ux-c #awsccc-sb-a a:hover,#awsccc-sb-ux-c #awsccc-sb-a a>span:hover,#awsccc-sb-ux-c #awsccc-sb-a a svg path:hover{color:#0073bb;text-decoration:underline}#awsccc-sb-ux-c #awsccc-sb-a .awsccc-tab-helper{outline:0;text-decoration:none}.awsccc-cs-modal-open{overflow:hidden;-webkit-box-sizing:border-box;box-sizing:border-box}',""]),e.exports=a},function(e,a,t){"use strict";Object.defineProperty(a,"__esModule",{value:!0}),a.convertToArray=a.update=void 0,a.update=function(e,a){return Object.keys(a).forEach((function(t){e[t]=a[t]})),e},a.convertToArray=function(e){return Array.prototype.slice.call(e)}},function(e,a,t){"use strict";Object.defineProperty(a,"__esModule",{value:!0});a.default=function(e,a,t){function n(){a.removeEventListener("DOMContentLoaded",n),t.removeEventListener("load",n),e()}void 0===a&&(a=document),void 0===t&&(t=window),"loading"!==a.readyState?t.setTimeout(e):(a.addEventListener("DOMContentLoaded",n),t.addEventListener("load",n))}}])}));
```
sammartinez, line 1: Lets chat about this offline as well.

client/src/docs-ui/page/page.tsx
tjleing: `track` and `AnalyticsEventType` are used right above the `trackPageVisit` and `trackPageFetchException` calls -- they're being used for Pinpoint tracking right now.

tjleing: We'll rip out those `track()` calls once we're fully switched over to Adobe from Pinpoint.  I'm still unsure whether a `switch`-based approach (i.e. calling `track()` with a given event type) or this current approach with the specific named methods like `trackPageVisit` is more appropriate

client/src/utils/track.ts
tjleing: I tried that first and it doesn't work... actually it might be better to use `if (Build.isBrowser)` instead as is done in `track.ts` -- the code is there fully for testing

client/src/utils/track.ts
tjleing: Think that's it -- Shortbread does a lot automatically

client/src/utils/track.ts
tjleing: Strangely no, it's `AWSCShortbread({...})` -- not sure why 😅

client/src/utils/track.ts
sammartinez: can you shorten this to be a one line then? 
```suggestion
    AWSCShortbread({ domain: ".amplify.aws" }).checkForCookieConsent();
```

client/src/utils/track.ts
sammartinez: Interesting why that wouldn't work 🤔

harrysolovay (approved):
LGTM

client/src/utils/track.ts
tjleing: ooh probably.. not sure why I didn't see this before

tjleing: fixed!

Issue #2703, opened 2020-11-25
UI Components > Storage > Properties keeps showing the same tag info


watilde:
**Describe the bug**
In Storage of UI Components page, properties section reminds the tag I opened for the first time even I move to different items.

**To Reproduce**
1. Open https://docs.amplify.aws/ui/storage/s3-album/q/framework/react
2. Click the next menu in the side bar "S3 Images"
3. Check Properties section and it keeps showing `amplify-s3-album`'s property

**Expected behavior**
Properties shows `amplify-s3-image`'s property.

**Screenshots**
<img width="523" alt="Screen Shot 2020-11-25 at 21 19 27" src="https://user-images.githubusercontent.com/1716463/100227089-eed2eb00-2f63-11eb-9fd8-d97573193203.png">

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Chrome
 - Version Version 87.0.4280.67 (Official Build) (x86_64)

**Additional context**
Originally reported at https://twitter.com/tacck/status/1331446661743144962.

PR #2704, opened 2020-11-25
fix: handle re-render in ui-component-props

watilde:
To handle re-render in stencil's component lifecycle, it needs to use `componentWillRender` instead of `componentWillLoad` when app uses stencil-router.

Fixes: https://github.com/aws-amplify/docs/issues/2703

*Description of changes:*
Refs: https://stenciljs.com/docs/component-lifecycle


wlee221 (approved):
This LGTM, thanks for the catch!

Issue #2705, opened 2020-11-25
UI Components > Auth > "Select MFA Type" shows undefined


watilde:
**Describe the bug**
UI Components > Auth > "Select MFA Type" shows `undefined ` when I come from "Sign out" menu.

**To Reproduce**
1. Open https://docs.amplify.aws/ui/auth/select-mfa-type/q/framework/react
2. Click "Sign Out" in the side bar
2. Click "Select MFA Type" in the side bar
3. See `undefined`

**Expected behavior**
It should show `Properties` table.

**Screenshots**
<img width="660" alt="Screen Shot 2020-11-25 at 21 30 14" src="https://user-images.githubusercontent.com/1716463/100228032-6c4b2b00-2f65-11eb-9b8b-dd77a070a2d0.png">

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Chrome
 - Version Version 87.0.4280.67 (Official Build) (x86_64)

**Additional context**
Originally reported at https://twitter.com/tacck/status/1331446663299272706.

mauerbac:
Thanks @watilde . I see Sam will review your PR

watilde:
This seems to be still exisiting. I'd appriciate if I can ask for your team's review on PR :)

medelman17:
Thanks, @watilde, for bumping this. Will start to ping relevant stakeholders tomorrow to get resolution ASAP. My apologies for the delayed response.

PR #2706, opened 2020-11-25
fix: provide prop-type on select-mfa-type

watilde:
`ui-component-props` handles `prop-type` as `attr` with no value only for the initial load. Passing a correct `prop-type` fixes an issue that showing `undefined` when user comes from different page which is how the other page manages prop-type.

Fixes: https://github.com/aws-amplify/docs/issues/2705


watilde:
@sammartinez This is just a friendly reminder that we are waiting for your review. Hope we can fix the issue together :)

watilde:
Ping to keep opening

watilde:
@sammartinez Kindly asking for your review to ship this as the issue is still happening.

medelman17:
Circulated to @aws-amplify/amplify-js internally for review. Apologies for the delay, here.

medelman17 (approved):

PR #2673, opened 2020-11-18
Add datastore flutter code snippets

Amplifiyer:
This PR doesn't add all the snippets, still TBDs are

* Relational data access
* Sync
* Events
* Auth rules
* Conflict resolution

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2710, opened 2020-11-26
Add relational and sync data access examples for flutter datastore

Amplifiyer:
Continuation of #2673 


docs/lib/datastore/fragments/flutter/relational/delete-snippet.md
```diff
@@ -1,3 +1,11 @@
...
-TBD
+(await Amplify.DataStore.query(Post.classType, where: Post.ID.eq("123")))
+    .forEach((element) async {
+  try {
+    await Amplify.DataStore.delete(element);
+    print("Deleted a post");
+  } catch (e) {
+    print("Delete failed: " + e);
+  }
+});
```
jamesonwilliams, line 10: Do you need to include the outer-most `(` and `)`? 

Instead of doing `forEach` after the `await`, you could use `asStream()` and then iterator over it with the `forEach` on stream.

I wonder if there is a way to merge all of the `delete` calls, too, without that inner `await`.

This might be one way to do it.
```dart
await Amplify.DataStore.query(Post.classType, where: Post.ID.eq("123"))
    .then((queryResults) => {
        Future.forEach(queryResults, (item) => Amplify.DataStore.delete(item))
    })
```

Apologies if I'm bad at Dart. 😳  Syntax might be off.

docs/lib/datastore/fragments/flutter/relational/query-snippet.md
```diff
@@ -1,3 +1,8 @@
 ```dart
-TBD
+try {
+  List<Comment> comments = await Amplify.DataStore.query(Comment.classType,
+      where: Post.STATUS.eq(PostStatus.ACTIVE));
+} catch (e) {
+  print("Query failed: " + e);
+}
```
jamesonwilliams, line 7: In the Android and iOS docs, we don't show the blocking versions of the calls. I'd suggest something instead like:

```suggestion
Amplify.DataStore.query(Comment.classType, where: Post.STATUS.eq(PostStatus.ACTIVE))
    .asStream()
    .listen((result) => {
       // yooo, dude, sick.
    })
    .onError((error) => {
      // damn bruh.
    })
```

^ Note: syntax might be slightly off.

docs/lib/datastore/fragments/flutter/relational/save-many-snippet.md
```diff
@@ -1,11 +1,14 @@
 ```dart
-TBD
+Post post = Post(title: "My First Post");
+User editor = User(username: "Nadia");
+PostEditor postEditor = PostEditor(post: post, editor: editor);
+
+try {
+  await Amplify.DataStore.save(post);
+  await Amplify.DataStore.save(editor);
+  await Amplify.DataStore.save(postEditor);
```
jamesonwilliams, line 9: Maybe something like:

```dart
await Stream.fromIterable([post, editor, postEditor])
    .forEach((model) => await Amplify.DataStore.save(model))
```

docs/lib/datastore/fragments/flutter/relational/save-snippet.md
```diff
@@ -1,3 +1,16 @@
...
+Comment comment = Comment(
+    post: post, // Directly pass in the post instance
+    content: "Loving Amplify DataStore!");
+
+try {
+  await Amplify.DataStore.save(post);
+  print("Post saved");
+  await Amplify.DataStore.save(comment);
+  print("Comment saved");
+} catch (e) {
```
jamesonwilliams, line 13: Is there a way to remove these `catch` statements from the examples, for brevity? Maybe you can add an `.onError` hook to the end of a functional chain?

docs/lib/datastore/fragments/flutter/relational/delete-snippet.md
Amplifiyer: There is no outer `()` The opening parenthesis is ending on the first line itself so we can call `forEach` on the results and not promises and the last `)` is closing the `forEach`

I'm not seeing a benefit of converting these to streams to do the same, this seems more straightforward usage of the APIs. The `then` is not really recommended over `async/await`

docs/lib/datastore/fragments/flutter/relational/query-snippet.md
Amplifiyer: I'm not sure if this is the best way to show the usage of the APIs. `.asStream().listen()` is very similar to programming paradigm of using `then()` which creates a callback hell for developers to maintain. `async/await` never blocks the main thread, it just schedules the rest of the function body to execute after the await fulfills.

docs/lib/datastore/fragments/flutter/relational/save-many-snippet.md
Amplifiyer: Kept similar to the [JS version](https://docs.amplify.aws/lib/datastore/relational/q/platform/js#many-to-many) of the docs. Added step wise comments of saving first post, then editor and finally postEditor. Here the order matters and I think it's explicit this way.

haverchuck (approved):

docs/lib/datastore/fragments/flutter/relational/save-snippet.md
Amplifiyer: Removed catch statements on the relational data access snippets

jamesonwilliams (approved):

PR #2723, opened 2020-11-30
Fix Completable syntax

jpignata:
*Issue #, if available:* n/a

*Description of changes:* Completable.merge expects an array, which means we'd have to wrap the
completables as a list. Completable.mergeArray takes n Completables and
should be what we use here.

- Fix incorrect usage of Completable.merge with only one parameter


jpignata:
Closing in favor of #2724

PR #2743, opened 2020-12-04
Update flutter datastore sync docs for developer preview

Amplifiyer:
*Description of changes:* Temporarily remove the `add API plugin` section since we don't have an API plugin and we automatically add if we detect the api configuration in `amplifyconfiguration.dart`.

This change will be reverted once we release the API plugin.

### DIFF
--- a/docs/lib/datastore/fragments/flutter/sync/10-installPlugin.md
+++ b/docs/lib/datastore/fragments/flutter/sync/10-installPlugin.md
@@ -1,28 +1,3 @@
 ### Add the API plugin
 
-DataStore's cloud synchronization uses the [API category](~/lib/graphqlapi/getting-started.md) behind the scenes. Therefore, the first step is to add an API plugin.
-
-Make sure that you declare a dependency on the DataStore plugin in your pubspec.yaml:
-
-```yaml
-dependencies:
-  flutter:
-    sdk: flutter
-  amplify_api: '<1.0.0'
-```
-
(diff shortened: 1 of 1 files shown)


docs/lib/datastore/fragments/flutter/sync/10-installPlugin.md
```diff
@@ -1,28 +1,3 @@
 ### Add the API plugin
```
jamesonwilliams, line 1: Do you want to leave this header here? Or add it back later?

jamesonwilliams (approved):

docs/lib/datastore/fragments/flutter/sync/10-installPlugin.md
Amplifiyer: yeah I was actually thinking about removing the whole section since it's really a no-op for users in developer preview. But then I thought to leave this here to be explicit that we only add the API plugin implicitly in developer preview. What do you recommend, should we remove the whole thing or keep this or keep this without the header?

jamesonwilliams: If we leave it, maybe we should make the future call to action more explicit, like:

> “We are currently building the API (GraphQL) category for Flutter. When it becomes available, you will need to add it explicitly in order to use remote data synchronization. This step is not needed in the Developer Preview.”

But I think either approach will work just fine.

Amplifiyer: I like it, stealing.

jamesonwilliams (approved):

PR #2744, opened 2020-12-04
Added minor updates to javascript datastore

dabit3:
Added the following updates:

1. Instructions for querying for single item by ID
2. Closing a subscription

> Sidenote - I think that it would make sense to document __1__ on other platforms as well as, from what I can tell, we assume this knowledge as of now by just referencing it as _an instance of a model_ (i.e., [here](https://docs.amplify.aws/lib/datastore/data-access/q/platform/ios#delete))

### DIFF
--- a/docs/lib/datastore/fragments/js/data-access/query-single-item-snippet.md
+++ b/docs/lib/datastore/fragments/js/data-access/query-single-item-snippet.md
@@ -1,0 +1,7 @@
+### Querying for a single item
+
+To query for a single item, pass in the ID of the item as the second argument to the query.
+
+```js
+const post = await DataStore.query(Post, "1234567");
+```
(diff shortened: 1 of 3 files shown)


ericclemmons (commented):
Thanks for adding this!  This is now consistent with https://docs.amplify.aws/lib/datastore/examples/q/platform/js#example-application, which is what we want so there aren't dangling sockets in reactive applications.

drochetti (approved):
Thanks for this!

jamesonwilliams (approved):

sw-yx (approved):

mauerbac (approved):
Lgtm

dabit3:
Thanks everyone :)

PR #2765, opened 2020-12-11
Fix import module in Greetings example with React

watilde:
*Description of changes:*
Replace `AmplifySignOut` with a `AmplifyGreetings` which is actually used in the example code.


mauerbac (approved):
Thanks @watilde . LGTM

PR #2772, opened 2020-12-14
fix double quotation mark errors

jakemraz:
*Issue #, if available:*
https://docs.amplify.aws/sdk/push-notifications/setup-push-service/q/platform/android#handling-fcmgcm-push-notifications

*Description of changes:*
The sample code of `Adding Receiver` has double quotation mark errors which is not suitable for programming language.
So I fixed it.

### DIFF
--- a/docs/sdk/push-notifications/fragments/android/handle-fcm.md
+++ b/docs/sdk/push-notifications/fragments/android/handle-fcm.md
@@ -14,9 +14,9 @@ The SDK provides `PinpointNotificationReceiver` which handles the notification t
 
 ```xml
 <receiver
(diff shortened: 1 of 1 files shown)


jamesonwilliams (approved):

jamesonwilliams:
Passing thought: I wonder if we can add some tooling, to check for ” and “. In natural language portions of the documentation, we should prefer these. However, we should disallow them (instead preferring ") within code blocks, as they don't copy-paste well.

jakemraz:
I think we can make the policy which finds ” and “ between code blocks consisting of ```and ``` using regex

PR #2773, opened 2020-12-14
fix a case-sensitive error in the DeepLink example

jakemraz:
*Issue #, if available:*
https://docs.amplify.aws/sdk/push-notifications/setup-push-service/q/platform/android#open-a-deep-link

*Description of changes:*
The code example above uses DeepLinkActivity but the code example below uses DeeplinkActivity.
If you use this example as it is, it will not be compiled.
So I changed below DeeplinkActivity to DeepLinkActivity (Lower case letter changed to upper case)


jamesonwilliams (approved):
Google docs seem to refer to this as two distinct words, "deep link" ~= "DeepLink." Ours has a German-style compound noun: "deeplink" and "Deeplink."

Anyway, you change is moving things in the right direction. I note that this also fixes a technical bug, where the Activity name does not match the one specified in the manifest:
```xml
android:name=".DeepLinkActivity"
```

mauerbac (approved):
LGTM

PR #2783, opened 2020-12-21
fix: use sensible param names in arrow functions

jamesonwilliams:
In JavaScript's documentation of the Selective Sync feature for
DataStore, `(c)` is used as the single parameter name in most of the
lambdas that are shown. Ostensibly, this parameter name was selected to
represent a `Comment` model, which is one of the types referenced in the
documentation. It is *also* used for `Post` and `User`, which makes no
sense.

To remedy this, we just use full nouns, instead: `user`, `post`, `comment`.
In all places where non-`Comment` models are referenced, the updated name
is used, in place of `c`.

Additionally, the optional parentheses are dropped, as JavaScript allows
for single-parameter arrow functions.


wlee221 (approved):
LGTM! 👍

PR #2790, opened 2020-12-22
fix wrong code example

tklaas:
*Issue #, if available:* N/A

*Description of changes:*
copy and paste error, code example was showing the same sample as previous box. The text right before the code sample was referencing an example with graphqlOperation

### DIFF
--- a/docs/lib/graphqlapi/fragments/js/mutate-data.md
+++ b/docs/lib/graphqlapi/fragments/js/mutate-data.md
@@ -25,7 +25,7 @@ You can optionally import the `graphqlOperation` helper function to help you con
 ```javascript
 import { API, graphqlOperation } from 'aws-amplify';
 // ...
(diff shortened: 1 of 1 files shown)


amhinson (approved):
Good catch @tklaas! 🙏   LGTM

docs/lib/graphqlapi/fragments/js/mutate-data.md
```diff
@@ -25,7 +25,7 @@ You can optionally import the `graphqlOperation` helper function to help you con
 ```javascript
 import { API, graphqlOperation } from 'aws-amplify';
 // ...
-const newTodo = await API.graphql({ query: mutations.createTodo, variables: {input: todoDetails}})); // equivalent to above example
```
jamesonwilliams, line 28: "equivalent to above example," indeed. 😆

PR #2795, opened 2020-12-23
Rename custom cloudformation stacks to custom aws resources

dabit3:
Custom AWS Resources is a clearer and better name IMO, especially for devs new to AWS.


jamesonwilliams (approved):

kaustavghosh06 (approved):
LGTM.

PR #2809, opened 2020-12-28
Fix anchor links in React docs for Auth Props

mootrichard:
The links were not pointing to the correct id's, so it wasn't possible to jump to right place to see what properties you could be passing in.

*Issue #, if available:*
N/A

*Description of changes:*
I updated the anchor tags to point to `#props-attr-amplify-authenticator` instead of `#props-amplify-authenticator`, since `#props-amplify-authenticator` doesn't seem to link to anything on the page.


jamesonwilliams (approved):
Looks great. Thanks @mootrichard! These little docs changes may seem trivial, but I swear, they're the hardest to _find_, so they make a big difference.

Issue #2549, opened 2020-06-18
using @aws-amplify/ui-react how to fully customize the Auth flows documentation needed


ecomonestop:
**Is your feature request related to a problem? Please describe.**
Fully customizing the Auth flows by providing your own "slot"(not using AmplifyUIComponents at all; only using AmplifyAuthenticator and your custom component "slot" that uses Auth API itself ) is not well documented, especially how to handle AuthState changes.  Furthermore, the way I implemneted it seems to be very boilerplate code that I would think AmplifyAuthenticator would handle.  The old Authenticator did handle this boilerplate I believe; not sure why not in the new library. 

**Describe the solution you'd like**
I would like a clear documentation on how to fully customize your AuthFlows using the new library.  I believe there is a big gap when it comes to handling AuthState changes.  Further, maybe some of the boilerplate below can be handled by AmplifyAuthenticator?

**Describe alternatives you've considered**
Currently, I'm doing the following: ("See below code") Notice how I have to handle AuthState changes myself.  Should this be handled by the AmplifyAuthenticator?  As of today, is this best practice?  Finally, Please note that the <Cust.../> components are my own React components that have its own custom UI and calls the Auth API.

```
const AuthLayout = ({ children, initAuthState }) => {
  const [authState, setAuthState] = React.useState(initAuthState)
  const [authData, setAuthData] = React.useState(null)
  const handleAuthStateChange = (nextAuthState, authData) => {
    Hub.dispatch(UI_AUTH_CHANNEL, {
      event: AUTH_STATE_CHANGE_EVENT,
      message: nextAuthState,
      data: authData ? authData.user : null,
    });
  }
  React.useEffect(() => {
    return onAuthUIStateChange((nextAuthState, authData) => {
      setAuthState(nextAuthState)
      setAuthData(authData)
    });
}, []);

  return (
    <AmplifyAuthenticator initialAuthState={initAuthState}  >
      <CustSignUp authState={authState} authData={authData} handleAuthStateChange={handleAuthStateChange} ></CustSignUp>
      <CustConfirmSignUp authState={authState} authData={authData} handleAuthStateChange={handleAuthStateChange} ></CustConfirmSignUp>
      
      {authState === AuthState.SignedIn && authData && children}
    </AmplifyAuthenticator>
  )
}
```

dabit3:
[...]

Issue #2513, opened 2020-10-08
[Feedback]How to implement handleAuthStateChange should be documented


kjetilge:
**Page**: [`/lib/utilities/hub/q/platform/js`](https://docs.amplify.aws/lib/utilities/hub/q/platform/js#state-management)

**Feedback**:

There is a nice explanation [here:](https://github.com/aws-amplify/amplify-js/issues/6113)

At the very least include the following with a little explanation :

```
    handleAuthStateChange (nextAuthState, authData) {
      console.log('Next state: ', nextAuthState, 'auth data: ', authData)
      Hub.dispatch('UI Auth', {
        event: 'AuthStateChange',
        message: nextAuthState,
        data: authData ? authData.user : null,
      });
    }
```
And usage:
```
this.handleAuthStateChange(AuthState.ConfirmSignUp, { user: { username: user.username } })
```

mauerbac:
Thanks for calling this out. I moved the issue you [tagged](https://github.com/aws-amplify/docs/issues/2549) over and will work to update the documentation. Will close this one.

PR #2811, opened 2020-12-29
Documented handleauthstatechange for React

dabit3:
Documents `handleauthstatechange` in React, should address concerns in #2549 & #2513

### DIFF
--- a/docs/ui/auth/fragments/react/handleauthstatechange.md
+++ b/docs/ui/auth/fragments/react/handleauthstatechange.md
@@ -6,17 +6,17 @@ Similar to `onAuthUIStateChange`, `handleAuthStateChange` is a function you can 
 
 ```js
 import { AmplifyAuthenticator } from "@aws-amplify/ui-react";
+import { AuthState } from "@aws-amplify/ui-components";
 
 handleAuthStateChange((nextAuthState, authData) => {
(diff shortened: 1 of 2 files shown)


docs/ui/auth/fragments/react/handleauthstatechange.md
```diff
@@ -0,0 +1,22 @@
...
+
+```js
+import { AmplifyAuthenticator } from "@aws-amplify/ui-react";
+
+handleAuthStateChange((nextAuthState, authData) => {
+  if (nextAuthState === AuthState.SignedIn) {
+    console.log("user successfully signed in!");
+    console.log("user data: ", authData);
+  }
+  if (!authData) {
```
jamesonwilliams, line 15: Is it ever possible for `authData` to be falsey, but `nextAuthState` to be `AuthState.SignedIn`? If so, this might print both "user successfully signed in!" and "user is not signed in...".

```suggestion
  } else {
```

jamesonwilliams (approved):
Looks good!

docs/ui/auth/fragments/react/handleauthstatechange.md
dabit3: Not sure, but I updated the code to just print out the data if it exists.

PR #2826, opened 2021-01-05
Prerendering/SSG docs site

tjleing:
Stencil pre-rendering on the docs site has been broken since the beginning.  There are some current performance issues that this pull request is looking to reduce.  I think I've found a slightly hacky solution to the pre-rendering errors.

### The problem
The issue stemmed from a conflict between Stencil's prerendering mock document and Axios.  In short, Axios attempts to test if a request URL has the same origin as the base URL does by creating an <a> element and checking its pathname.  This doesn't exist because the page is a Stencil MockDocument and the <a> created is a Stencil MockAnchorElement, which doesn't have this property.

### The fix
Luckily, Axios also has a builtin function to determine whether or not it's running inside of a mock environment, in which case it doesn't bother with the request check.  Here's the snippet:
```
function isStandardBrowserEnv() {
  if (typeof navigator !== 'undefined' && (navigator.product === 'ReactNative' ||
                                           navigator.product === 'NativeScript' ||
                                           navigator.product === 'NS')) {
    return false;
  }
  ...
}
```
So all we need to do is set the navigator product to one of these values, and prerendering will pass.  Stencil provides a hook that runs before prerendering, so we can access the mock navigator and set this property.  Now that the change is made, there are no more Stencil errors when `$ yarn build` is ran.

TODO: does the prerendering actually work?  I'm not entirely sure how to test it, but you can almost certainly run the SSG build somehow.


ericclemmons (approved):
Confirmed it doesn't error, but probably doesn't "work" as we thought!

```shell
$ yarn build
$ npx serve client/www
$ open http://localhost:5000/
```

So let's compare before...

<view-source:https://docs.amplify.aws/start/q/integration/react>

```html
...

<body> <docs-router></docs-router> <script src="https://cdn.jsdelivr.net/npm/docsearch.js@2.6.3/dist/cdn/docsearch.min.js"></script>
--
  | <noscript>This website requires your browser to execute JavaScript.</noscript> <script>if ('serviceWorker' in navigator && location.protocol !== 'file:') {
  | // auto-unregister service worker during dev mode
  | navigator.serviceWorker.getRegistration().then(function(registration) {
  | if (registration) {
  | registration.unregister().then(function() { location.reload(true) });
  | }
  | });
  | }</script>
  | </body>
  | </html>


```

To after...
<view-source:http://localhost:5000/start/q/integration/react>

```html
...

<body> <docs-router class=hydrated s-id=1><!--r.1--><stencil-router class=hydrated c-id=1.0.0.0 s-id=2><!--r.2--><!--o.1.1--><!--s.2.0.0.0.--><stencil-route-switch
--
  | group="1666-4859-4727-7356" class=hydrated c-id=1.1.1.0 s-id=3><!--r.3--><!--o.1.2--><!--s.3.0.0.0.--><stencil-route
  | group="1666-4859-4727-7356" class=hydrated c-id=1.2.2.0 s-id=4><!--r.4--><docs-page class=hydrated c-id=4.0.0.0
  | s-id=5><!--r.5--></docs-page></stencil-route></stencil-route-switch></stencil-router></docs-router> <script
  | src="https://cdn.jsdelivr.net/npm/docsearch.js@2.6.3/dist/cdn/docsearch.min.js"></script>
  | <noscript>This website requires your browser to execute JavaScript.</noscript> <script>if ('serviceWorker' in navigator && location.protocol !== 'file:') {
  | // auto-unregister service worker during dev mode
  | navigator.serviceWorker.getRegistration().then(function(registration) {
  | if (registration) {
  | registration.unregister().then(function() { location.reload(true) });
  | }
  | });
  | }</script>
  | </body>
```

Notice how we get a _little_more markup inside the `<body>`?  But not the entire page.

My thinking is that there are asynchronous fetches/imports/code-splitting within the router, which **is not** being pre-rendered.

This is typical of React sites as well: nested, asynchronous renders aren't possible yet.

**So, what should we do with this PR**?

It _works_, so I'm inclined to approve it and accept it.  The only caveat I have is if it noticeably increases build times.  Honestly, I didn't see a significant difference, but you may have more data on that than I do.

Anyway, **great work** on identifying the root cause, and how to work around it!  What this does tell me is that, as we incrementally adopt Next.js going forward, that we should expect to see a benefit to pre-rendering that Stencil cannot provide currently.

tjleing:
I just ran `time` on both versions:

- with SSG fix:
```
✨  Done in 97.89s.
yarn build  229.83s user 27.89s system 262% cpu 1:38.09 total
```
- without SSG fix:
```
✨  Done in 181.55s.
yarn build  261.18s user 23.88s system 156% cpu 3:01.76 total
```

Not really sure what conclusion to come to from this.  SSG is much faster wallclock-wise but slower in terms of CPU time?  Also CPU utilization should probably be pretty important.

ericclemmons:
@tjleing I agree. Ultimately, the total is what matters most for us, and we won't really "feel" the difference in our CI.

At best, any improvements we make going forward will benefit customers. Today, there's not _as much_, but this at least unlocks the potential.

Thanks for getting this data!

tjleing:
Progress has been made!  Prerendered output of `http://localhost:5000/guides/q/platform/android` [on Pastebin](https://pastebin.com/d5waWHiE) (too large to fit in one comment)

Summary of changes:
- Regarding the problem seen above where SSG progress was made, but the prerendering shell was still empty: the issue simply was that `docs-page` had a specific check against non-browser environments that only rendered the outer shell.
- Stencil was infinitely looping when asked to prerender any page.  Originally, I thought this was an error with Stencil's page crawler repeatedly hitting each link, because the `docs-page` `render()` function was being called repeatedly.  However, it turns out that the issue was with `setSidebarStickyTop()`: the `docs-page` component was stuck in a loop of asking for the height of the Stencil mock window, receiving `NaN`, and rerendering because `NaN !== NaN`.  Disabling this for non-browser environments resolved the loop.
- `capi` prints an error whenever it finds a URL that it can't resolve to a docs page.  This includes external links to AWS documentation and so on.  This wouldn't normally be a problem, but Stencil checks for _any_ printed errors to determine whether or not the prerendering process completed successfully.  The quick fix is to change the `capi` message to a `console.warn()` instead, but we might want to resolve it a different way before this gets merged in.
- Authenticator pages embed the Amplify Authenticator UI component, including all of its Amplify logic.  Thus, they require Amplify to be configured first.  However, configuring Amplify leads to problems, because Pinpoint looks fails for some reason in a non-browser environment.  For now, Authenticator pages are excluded from the prerender list, and each Pinpoint `track()` call is guarded by a browser check.

Remaining issues:
- Look into the Pinpoint failures to see if there's some way to let Amplify configure and prerender Authenticator pages.
- The connection is closing before some pages finish prerendering, causing a lot of console ECONNCLOSED errors (maybe single-thread it?)
- I seem to have `rebase`d wrong, so that this PR contains 62 commits instead of the expected 3... will investigate tomorrow.
- A few layout issues with respect to menu + setSidebarStickyTop() (see attached picture): version switcher is being rendered way too low, and both mobile and desktop menu close buttons are rendering.

image

tjleing:
SSG preview branch metrics:
image

Master preview branch metrics:
image

Looks like an improvement!

- times are longer than the normal website because the preview branches are hosted on a different availability zone
- the home page doesn't have this improvement -- it seems to not be prerendered...

client/prerender.config.ts
```diff
@@ -20,6 +24,6 @@ export const config: PrerenderConfig = {
     };
   },
   filterUrl(url) {
-    return !!(url && entryUrls.includes(url.pathname) && url.pathname !== "/");
+    return !(url && url.toString().includes("authenticator"));
```
sammartinez, line 27: Why just the authenticator check?

client/src/docs-ui/page/page.tsx
```diff
@@ -101,7 +101,7 @@ export class DocsPage {
 
   @Listen("resize", {target: "window"})
   setSidebarStickyTop() {
-    if (this.pageData?.menu) {
+    if (Build.isBrowser && this.pageData?.menu) {
```
sammartinez, line 104: When is the build not a browser?

client/src/docs-ui/page/page.tsx
```diff
@@ -200,15 +200,16 @@ export class DocsPage {
...
-    track({
-      type: AnalyticsEventType.PAGE_VISIT,
-      attributes: {
-        url: currentRoute,
-        previousUrl: this.previousPathname,
-        referrer: document.referrer,
-      },
-    });
-    trackPageVisit();
+    if (Build.isBrowser) {
```
sammartinez, line 203: Do you need the `location.pathname === "/"` as well here?

sammartinez (commented):
Couple callouts and questions

client/prerender.config.ts
tjleing: The original check wasn't really doing anything -- entryUrls is exactly the crawled URL list, and prerendering used to fail on the home page, so I'm guessing it just got left in.  Now, authenticator pages are failing because they include the Amplify authenticator component, which errors if Amplify isn't configured (and we can't configure it because of pinpoint).  I'm still looking into all of this but the simple fix for now is to just not prerender authenticator pages.

client/src/docs-ui/page/page.tsx
tjleing: When prerendering and running tests -- I don't think there's a way to distinguish between the two.

client/src/docs-ui/page/page.tsx
tjleing: I don't understand why we would?  The page navigation event should fire whenever we switch pages, unless we're prerendering.

client/src/docs-ui/landing-hero-cta/__snapshots__/landing-hero-cta.spec.ts.snap
```diff
@@ -14,7 +14,7 @@ exports[`docs-landing-hero-cta Render logic should render 1`] = `
     <docs-internal-link class="scale-up-on-hover" href="/start/q/integration/vue">
       <img alt="Vue Icon" class="css-16ygk4v" src="/assets/integrations/vue.svg">
     </docs-internal-link>
-    <docs-internal-link class="scale-up-on-hover" href="/start/">
+    <docs-internal-link class="scale-up-on-hover" href="/start">
```
sammartinez, line 17: Does this break anything?

tjleing: No -- in fact it fixes the "no such page `/start/`" error that comes up in the console when visiting the home page before selecting a framework.

tjleing:
This most recent PR seems to only have addressed the toggle view issue locally, not on the preview branch...  I'm really not liking how inconsistent this is.

capi/src/write/get-page.ts
```diff
@@ -45,7 +45,7 @@ export async function getPage(config: t.Config, ctx: t.Ctx): Promise<void> {
 
         writer
           .writeLine("      default: {")
-          .writeLine('        console.error(`No such page "${route}"`);')
+          .writeLine('        console.warn(`No such page "${route}"`);')
```
tjleing, line 48: Maybe revert to `error`

sammartinez: does it break the page to where a user can not use it?

capi/src/visitors/fragment-tags.ts
```diff
@@ -1,13 +1,25 @@
 import * as t from "../types";
 import clone from "clone-deep";
 
+const isPContainingFragments = function(node: t.HyperscriptNode): boolean {
+  if (!Array.isArray(node)) return false;
+  if (node[0] !== "p") return false;
```
sammartinez, line 6: So this is writing to the logic that the first node will always be a `div` ? Is there a chance it can be something different like a header tag or anything?

capi/src/visitors/fragment-tags.ts
```diff
@@ -1,13 +1,25 @@
...
+  if (node.length < 3) return false;
+  const firstChild = node[2];
```
sammartinez, line 8: Is this the first child?

client/src/docs-ui/card/card.tsx
```diff
@@ -45,7 +45,7 @@ export class DocsCard {
   /*** query string parameters to attach to the link */
   @Prop() readonly QSPs: string = "";
   /*** link tag to use */
-  @Prop() readonly containerTag: string = "docs-internal-link";
+  @Prop() readonly containertag: string = "docs-internal-link";
```
sammartinez, line 48: Why the change?

client/src/docs-ui/landing-hero-cta/landing-hero-cta.tsx
```diff
@@ -41,7 +41,7 @@ export class DocsLandingHeroCTA {
             />
           </docs-internal-link>
           <docs-internal-link
-            href="/start/"
+            href="/start"
```
sammartinez, line 44: Shouldnt there be a trailing forward slash for this?

client/src/docs-ui/page/page.tsx
```diff
@@ -198,17 +197,19 @@ export class DocsPage {
       Object.values(routeFiltersEntry).reduce((acc, curr) => {
         return [...acc, ...curr];
       }, []);
+    console.log(currentRoute);
```
sammartinez, line 200: Lets kill any `console.log(...)` that are not needed :)

client/src/docs-ui/universal-nav/universal-nav.tsx
```diff
@@ -21,6 +21,7 @@ export class DocsUniversalNav {
   @Prop() readonly brandIconBlend?: string;
 
   render() {
+    console.log(this.blend);
```
sammartinez, line 24: Do we need this?

sammartinez (commented):
few callouts

capi/src/write/get-page.ts
tjleing: No, it just shows up in the console as an error or a warning.

tjleing: Oh also maybe it blocks the prerendering process if this is an error?  I'll look into it.

capi/src/visitors/fragment-tags.ts
tjleing: This code checks to see if the hyperscript node we're looking at is a `<p>` containing `<internal-fragment>`s.  If it is, Stencil freaks out for some unknown reason (that I can't reproduce locally with a simpler example...) but that seems to be related to this expected behavior where elements aren't supposed to be nested in `<p>`s: https://codesandbox.io/s/div-in-a-p-forked-d5q1e.

capi/src/visitors/fragment-tags.ts
tjleing: Yes, the structure goes like this: `[<node name>, <map of node properties, like class, id, onhover, or whatever>, <child 1>, <child 2>, ...]`.  Maybe there's some TS magic you can do to make this more clear but I couldn't think of anything.

client/src/docs-ui/card/card.tsx
tjleing: Conflict between Stencil and JSX on how camelCased props are supposed to be treated -- one turns them into dashed-case but the other keeps it as camelCased.  This was causing some of the "no such page `undefined`" errors.

client/src/docs-ui/landing-hero-cta/landing-hero-cta.tsx
tjleing: No, the `/start` page exists whereas the `/start/` page doesn't.  While one is redirected to the other, `/start` is correct here.  We could theoretically remedy this instead by making trailing slash pages correct for all URLs, but it seems better to have this conform to the standard already in place.

client/src/docs-ui/page/page.tsx
tjleing: Yep, just debugging in the preview branch over here...

client/src/amplify-ui/sidebar-layout/sidebar-layout-toggle/sidebar-layout-toggle.tsx
```diff
@@ -20,10 +20,10 @@ export class AmplifySidebarLayoutToggle {
...
    */
   getClass = () => {
-    return {
-      [sidebarLayoutSidebarToggleStyle]: true,
-      ...(this.inViewClass ? {[this.inViewClass]: !!this.inView} : {}),
-    };
+    return (
+      sidebarLayoutSidebarToggleStyle +
+      (this.inViewClass && this.inView ? " " + this.inViewClass : "")
+    );
```
tjleing, line 26: This change probably didn't need to occur, but it looks a lot cleaner this way to my eyes.

client/src/docs-ui/page/page.tsx
```diff
@@ -169,9 +168,9 @@ export class DocsPage {
     }
   }
 
-  componentWillLoad() {
+  async componentWillLoad() {
     this.restoreBlockSwitcherState();
-    return this.getPageData();
+    return await this.getPageData();
```
tjleing, line 173: This also didn't end up doing anything (i.e. fixing the 404 header being wrong), but I think this is more correct, so I'm going to leave it.

client/src/docs-ui/page/page.tsx
```diff
@@ -263,39 +264,44 @@ export class DocsPage {
...
-              selectedFilters: this.selectedFilters,
-              setSelectedFilters: this.setSelectedFilters,
-              selectedTabHeadings: this.selectedTabHeadings,
-              setNewSelectedTabHeadings: this.setNewSelectedTabHeading,
-            }}
-          >
+    return (
+      <Host class={pageStyle}>
+        <pageContext.Provider
+          state={{
```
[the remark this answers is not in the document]
tjleing: I think it's just indented differently, so it looks like a code change when it's not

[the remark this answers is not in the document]
tjleing: (indented differently because of removing the `if` statement around it)

client/prerender.config.ts
```diff
@@ -4,6 +4,10 @@ import {routes as entryUrls} from "./src/api/routes";
 export const config: PrerenderConfig = {
   crawlUrls: false,
   entryUrls,
+  beforeHydrate(document, _) {
+    // @ts-ignore
+    document.defaultView.navigator.product = "NS";
```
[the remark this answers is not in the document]
tjleing: (or in function defn.)

medelman17 (approved):
Everything seems to be in working order to my (untrained) eye.

client/src/docs-ui/version-switch/version-switch.tsx
```diff
@@ -10,23 +10,25 @@ export class DocsVersionSwitch {
...
-        <docs-internal-link
-          href={this.rightOption.href}
-          childActiveClass={activeSwitchStyle}
-        >
-          <span>{this.rightOption.title}</span>
-          <span class="subtitle">{this.rightOption.subTitle}</span>
-        </docs-internal-link>
-      </Host>
-    );
+    if (this.leftOption && this.rightOption) {
```
sammartinez, line 13: What happens if one of these flags is false?

sammartinez (commented):
One callout

client/src/docs-ui/version-switch/version-switch.tsx
tjleing: Resolving based on offline conversation.

sammartinez (approved):
LGTM 🌮  Thanks for your hard work on this @tjleing !!!

PR #2829, opened 2021-01-06
Dependabot version bump

tjleing:
Dependabot is flagging `node-notifier@5.4.2` as vulnerable.  Updating `jest` and `ts-jest` in the main package as well as `capi` should fix it.  Also updating `aws-amplify` and `@aws-amplify/ui-components` to deal with `axios` vulnerability.

Tests are now breaking due to one of these package upgrades (almost certainly `jest` or `ts-jest`).  Investigation continues.


mauerbac (approved):
LGTM

sammartinez:
We should be able to get this in now or at least the axios fix since `aws-amplify` is now released

sammartinez (approved):
LGTM 🌮

PR #2833, opened 2021-01-07
Fix adblocker breaking docs site

tjleing:
Fix: Just guard all `s` accesses
Locally validated!

The issue is that the `s_code` is blocked as an analytics library, which makes `s` undefined.


sammartinez (approved):
LGTM 🌮

mauerbac (approved):
LGTM.

PR #2838, opened 2021-01-08
Auth Directive Description Bugs

kneekey23:
*Description of changes:* If you read the words and code associated with these tables, they do not present correctly. This PR is to correct that so that the tables correctly reflect what the code means and also what the text is describing.


docs/cli/graphql-transformer/auth.md
```diff
@@ -145,20 +145,20 @@ type Todo @model
 }
 ```
 
-In this schema, only the owner of the object has the authorization to perform update (`updateTodo`) and delete (`deleteTodo`) operations on the owner created object, but anyone can read them (`getTodo`, `listTodos`). This prevents the object from being updated or deleted by users other than the creator of the object while allowing all authenticated users of the app to read them.
+In this schema, only the owner of the object has the authorization to perform update (`updateTodo`), create (`createTodo`) and delete (`deleteTodo`) operations on the owner created object, but anyone can read them (`getTodo`, `listTodos`). This prevents the object from being updated or deleted by users other than the creator of the object while allowing all authenticated users of the app to read them.
```
dabit3, line 148: I think this is the confusion I've always had with this table in the past. Basically, before create, there is no owner. I wonder if we can convey this better, because in reality anyone can create a todo because there is no owner before creation.

dabit3 (approved):

docs/cli/graphql-transformer/auth.md
wooj2: 1.  I suggest that we change the code snippet to order the operations in the following manner (which correspond to CRUD)
```
operations: [create, update, delete]
```
2.  The wording can then change to something like:
```
In this schema, only the owner of the object has authorization to perform create (`createTodo`), update (`updateTodo`), and delete (`deleteTodo`) on the owner created object, while anyone else can only read (`getTodo`, `listTodos`)
```

wooj2: Despite wanting to change the wording, I think this change is a step in the right direction, so, LGTM!

wooj2 (approved):

docs/cli/graphql-transformer/auth.md
kneekey23: addressed this feedback and made the changes

docs/cli/graphql-transformer/auth.md
```diff
@@ -125,7 +125,7 @@ type Todo @model
 }
 ```
 
-In this schema, only the owner of the object has the authorization to perform read (`getTodo` and `listTodos`), update (`updateTodo`), and delete (`deleteTodo`) operations on the owner created object. This prevents the object from being updated or deleted by users other than the creator of the object.
+In this schema, only the owner of the object has the authorization to perform read (`getTodo` and `listTodos`), update (`updateTodo`), and delete (`deleteTodo`) operations on the owner created object. However, any authenticated user can create the object (`createTodo`). This prevents the object from being updated or deleted by users other than the creator of the object.
```
renebrandel, line 128:
```suggestion
In this schema, only the owner of the object has the authorization to perform read (`getTodo` and `listTodos`), update (`updateTodo`), and delete (`deleteTodo`) operations on the owner created object. However, any authenticated user can create the object (`createTodo`). This prevents the object from being read, updated, or deleted by users other than the creator of the object.
```

docs/cli/graphql-transformer/auth.md
```diff
@@ -158,7 +158,7 @@ Next, let's say that you wanted to modify the schema to allow only the owner of
 
 ```graphql
 type Todo @model
-  @auth(rules: [{ allow: owner, operations: [create, delete] }]) {
+  @auth(rules: [{ allow: owner, operations: [delete] }]) {
```
renebrandel, line 161: This is in conflict with line no. 141 & 148. We should verify the requirement of the "create" operation.

renebrandel (changes requested):

docs/cli/graphql-transformer/auth.md
kneekey23: updated to put create back, I think anyone can create it's not owner specific so it's confusing. it can be in the list or not and its permissions are the same is my understanding.

PR #2839, opened 2021-01-08
Fix footer on Firefox

tjleing:
`overflow: overlay` doesn't work on non-Webkit browsers.

### DIFF
--- a/client/src/docs-ui/menu/menu.style.ts
+++ b/client/src/docs-ui/menu/menu.style.ts
@@ -3,8 +3,9 @@ import {css} from "emotion";
 export const menuStyle = css`
   display: block;
   padding: 0 2.5rem;
-  overflow-y: overlay;
-  padding-bottom: 6rem;
+  overflow-y: scroll; /* for Firefox */
+  overflow-y: overlay; /* for Webkit browsers */
+  margin-bottom: 6rem;
 `;
 
 export const menuItemContainerStyle = css`


mauerbac (approved):
LGMT. Thanks for fixing

client/src/docs-ui/menu/menu.style.ts
```diff
@@ -3,8 +3,9 @@ import {css} from "emotion";
 export const menuStyle = css`
   display: block;
   padding: 0 2.5rem;
-  overflow-y: overlay;
-  padding-bottom: 6rem;
+  overflow-y: scroll; /* for Firefox */
```
wlee221, line 6: We might want `auto` instead of `scroll` so that it hides the scroll bar if there's no overflow

client/src/docs-ui/menu/menu.style.ts
```diff
@@ -3,8 +3,9 @@ import {css} from "emotion";
...
+  overflow-y: overlay; /* for Webkit browsers */
```
wlee221, line 7: Have you verified that line 6 doesn't get overwritten by line 7 on firefox?

wlee221 (commented):
LGTM, nice catch! Just got some nits :D

wlee221:
wait how did I not see this as merged... please disregard this!

client/src/docs-ui/menu/menu.style.ts
tjleing: yeah, it just gets totally removed, which is admittedly pretty weird...

Issue #2837, opened 2021-01-08
Incorrectly documented method to retrieve model's name for DataStore events (Amplify JS)


nubpro:
![image](https://user-images.githubusercontent.com/762914/104036653-a1a69200-520e-11eb-94f2-69bbe1eb9eaa.png)
![image](https://user-images.githubusercontent.com/762914/104036679-ab2ffa00-520e-11eb-8579-b1c0353435a0.png)
![image](https://user-images.githubusercontent.com/762914/104036717-b71bbc00-520e-11eb-86fd-fa15418c51d7.png)

This `modelName` does not exists and is a lie.
To retrieve the model's name you need to do `model.name` from the model instance.

Related issue:
https://github.com/aws-amplify/amplify-js/issues/7219

amhinson:
Closing, as this appears to be addressed in #2844.

nubpro:
> Closing, as this appears to be addressed in #2844.

Hmm that pr didnt get merged as it requires changes from AmplifyJS instead of the doc.

amhinson:
Ah sorry my mistake 👍

nubpro:
> Ah sorry my mistake 👍

All good 😬

PR #2844, opened 2021-01-11
fix: datastore events model name

iartemiev:
*Issue #, if available:*
Fixes https://github.com/aws-amplify/docs/issues/2837

*Description of changes:*
`modelName` -> `model.name`

image

### DIFF
--- a/docs/lib/datastore/datastore-events.md
+++ b/docs/lib/datastore/datastore-events.md
@@ -52,7 +52,8 @@ HubPayload: N/A
 Dispatched when a local change has been newly staged for synchronization with the Cloud
 
 HubPayload `outboxMutationEvent` contains:
-- `modelName` (String): the name of the model that is awaiting publication to the Cloud
+- `model`:
+    - `name` (String): the name of the model that was synced
 - `element`: 
     - `model` (Model): the model instance that will be published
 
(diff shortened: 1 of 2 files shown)


docs/lib/datastore/datastore-events.md
```diff
@@ -52,7 +52,8 @@ HubPayload: N/A
 Dispatched when a local change has been newly staged for synchronization with the Cloud
 
 HubPayload `outboxMutationEvent` contains:
-- `modelName` (String): the name of the model that is awaiting publication to the Cloud
+- `model`:
+    - `name` (String): the name of the model that was synced
```
amhinson, line 56: Did you mean to replace the descriptions here and for the one below?

iartemiev: Bad copypasta 🍝.. Good catch!

iartemiev: Fixed!

amhinson (approved):
LGTM 👍

richardmcclellan (changes requested):
I believe this will break the documentation for iOS and Android, because they both have a `modelName` property, not `model.name`.  Would it be reasonable to add a modelName property to the JS event for platform consistency?   If not, please make sure this docs update only affects JS.

iartemiev:
> Instead of this docs change, would it be reasonable to add a `modelName` property to the JS `OutboxMutationEvent` for platform consistency? 

Yea, I think that makes more sense. I don't see an issue with that. Will confirm and then close this PR in favor of adding that property to JS.

iartemiev:
> Instead of this docs change, would it be reasonable to add a `modelName` property to the JS `OutboxMutationEvent` for platform consistency? 

@richardmcclellan, yea, I think that makes more sense. I don't see an issue with that. Will confirm and then close this PR in favor of adding that property to JS.

PR #2849, opened 2021-01-12
refactor(ui): use single file for duplicated markdowns

wlee221:
*Issue #, if available:*

*Description of changes:* We have common markdowns that are used across many ui pages. This includes `yarn install` instructions, `main.js` setup for vue, and `app.module.ts` setup for angular. They are copy-pasted across our fragments and thus hard to maintain. This PR centralize them at ~/docs/ui/fragments so that there is only one file to manage for each of them.


docs/ui/fragments/angular/configure-module.md
```diff
@@ -0,0 +1,19 @@
+```js
```
tjleing, line 1: Why not put `_app.module.ts_` at the top here so you don't have to duplicate it as well? (similar for `main.js` above vue/configure-app.md)

tjleing (approved):
Looks good!  Thank you @wlee221 😄

docs/ui/fragments/angular/configure-module.md
wlee221: Good point, I will do that.

docs/ui/auth/fragments/ionic/auth-state-management.md
```diff
@@ -1,22 +1,4 @@
...
-
-@NgModule({
-  declarations: [AppComponent],
-  imports: [AmplifyUIAngularModule, BrowserModule],
-  providers: [],
-  bootstrap: [AppComponent],
-})
-export class AppModule {}
-```
+<inline-fragment src="~/ui/fragments/angular/configure-module.md"></inline-fragment>
```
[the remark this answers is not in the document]
wlee221: Yeah, that's why I didn't made separate entry for ionic, but we can revisit this later

tjleing (approved):

wlee221:
I found that migration guides can share the same file too, so I'm working on that. Sorry I'm making you review this multiple times @tjleing 😢

tjleing (approved):
Good find!

wlee221:
Thanks @tjleing!

PR #2855, opened 2021-01-14
Adobe Analytics -- second round of changes!

tjleing:
## Changes:
- [x] Search query changes
  - [x] only send one request upon clicking a search result
  - [x] send special event for no search results
    - [x] send after a delay when there's no search results, so that it only sends once
  - [x] send more `eVar`s and `prop`s to give more details about search events
  - [x] fixed search events 1 and 2 not sending
- [x] Fix more than one `s.t()` event being sent when first visiting pages
- [x] Fix 404s not tracking
- [x] Fix external link tracking not sending the link location
- [x] QSPs on home page and getting started page links for internal campaign tracking 

## Fun facts:
- Adobe Analytics doesn't send an `eVar` if its value is equal to 0, even if it's in the list of `eVar`s to send.

### DIFF
--- a/client/src/docs-ui/choose-integration-anchor/choose-integration-anchor.tsx
+++ b/client/src/docs-ui/choose-integration-anchor/choose-integration-anchor.tsx
@@ -30,7 +30,7 @@ export class DocsChooseIntegrationAnchor {
                 <docs-card
                   key={label}
                   url={route}
-                  QSPs={`?sc_campaign=${filterValue}-start&sc_channel=choose-integration`}
+                  QSPs={`?sc_icampaign=${filterValue}-start&sc_ichannel=choose-integration`}
                 >
                   <img slot="graphic" src={graphicURI} alt={`${label} Logo`} />
                   <h4 slot="heading">{label}</h4>
@@ -54,7 +54,7 @@ export class DocsChooseIntegrationAnchor {
                 <docs-card
                   key={label}
                   url={route}
-                  QSPs={`?sc_campaign=${filterValue}-start&sc_channel=choose-integration`}
+                  QSPs={`?sc_icampaign=${filterValue}-start&sc_ichannel=choose-integration`}
                 >
                   <img slot="graphic" src={graphicURI} alt={`${label} Logo`} />
                   <h4 slot="heading">{label}</h4>
--- a/client/src/docs-ui/choose-integration-anchor/choose-integration-anchor.tsx
+++ b/client/src/docs-ui/choose-integration-anchor/choose-integration-anchor.tsx
@@ -27,7 +27,11 @@ export class DocsChooseIntegrationAnchor {
                 this.page && `${this.page.route}/q/integration/${filterValue}`;
 
               return (
-                <docs-card key={label} url={route}>
+                <docs-card
+                  key={label}
+                  url={route}
+                  QSPs={`?sc_campaign=${filterValue}-start&sc_channel=choose-integration`}
+                >
                   <img slot="graphic" src={graphicURI} alt={`${label} Logo`} />
                   <h4 slot="heading">{label}</h4>
                 </docs-card>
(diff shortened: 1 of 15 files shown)


client/src/docs-ui/page/page.tsx
```diff
@@ -254,6 +254,9 @@ export class DocsPage {
       });
       trackPageFetchException();
     }
+    if (this.pageData === undefined) {
+      trackPageFetchException();
+    }
```
tjleing, line 259: Move up to L215 ish

client/src/docs-ui/page/page.tsx
```diff
@@ -254,6 +254,9 @@ export class DocsPage {
       });
       trackPageFetchException();
     }
+    if (this.pageData === undefined) {
```
sammartinez, line 257: Can we look to move this to 214 ish?

mauerbac (approved):
LGTM. Paired with TJ on Chime and reviewed functionality.

client/src/docs-ui/choose-integration-anchor/choose-integration-anchor.tsx
```diff
@@ -27,7 +27,11 @@ export class DocsChooseIntegrationAnchor {
                 this.page && `${this.page.route}/q/integration/${filterValue}`;
 
               return (
-                <docs-card key={label} url={route}>
+                <docs-card
+                  key={label}
+                  url={route}
+                  QSPs={`?sc_icampaign=${filterValue}-start&sc_ichannel=choose-integration`}
```
sammartinez, line 33: should `QSPs` be `qsps` ?

client/src/utils/track.ts
```diff
@@ -78,46 +79,106 @@ export const track = (event: AnalyticsEvent): Promise<unknown> | undefined => {
...
+    s.linkTrackVars =
+      "prop39,prop41,prop50,prop61,prop62,eVar26,eVar27,eVar39,eVar41,eVar50,eVar61,eVar62,eVar69,events";
+    // @ts-ignore
+    s.linkTrackEvents = "event1";
+    // @ts-ignore
+    s.events = "event1";
     // @ts-ignore
     s.tl(true, "o", "internal search");
   }
+  window.location.assign(suggestion.url);
```
sammartinez, line 183: Why this?

sammartinez (commented):
Overall, looks good had a couple callouts for clarity and curious if we have tested this in staging

client/src/docs-ui/choose-integration-anchor/choose-integration-anchor.tsx
tjleing: It's only `qsps` instead of `QSPs` in snapshots and HTML and such, not JSX, as far as I can tell... in any case, it works in testing.

client/src/utils/track.ts
tjleing: We're overriding the default behavior when clicking on a search result in order to send the analytics event.  When we're done tracking, we need to also implement the default behavior: sending the user to the page they clicked on, which wouldn't happen at all if we didn't manually do it.

Do you think this deserves a comment explaining it?

tjleing:
> Overall, looks good had a couple callouts for clarity and curious if we have tested this in staging

Thank you @sammartinez!  I manually tested it, just demoed it to @mauerbac, and the Adobe folks also had a look at it on the preview branch.

tjleing:
Worth considering putting some testing around Adobe Analytics -- @mauerbac was mentioning considering some integration testing in Selenium or maybe Puppeteer -- the tests would definitely just be about checking the contents of event requests that are fired, which doesn't seem too difficult.

client/src/utils/track.ts
sammartinez: Nah, I just wanted clarity is all on it so thank you.

sammartinez:
> Worth considering putting some testing around Adobe Analytics -- @mauerbac was mentioning considering some integration testing in Selenium or maybe Puppeteer -- the tests would definitely just be about checking the contents of event requests that are fired, which doesn't seem too difficult. Probably not for this PR, anyways, because getting this out the door to collect valuable data sooner rather than later is a priority.

Yeah we can look into that or instead of re-inventing the wheel, we can look to do a similar integration test suite with Cypress.

sammartinez (approved):
LGTM 🌮

PR #2865, opened 2021-01-16
Update getting-started.md

anthonyhumphreys:
*Issue #, if available:* #4315
https://stackoverflow.com/questions/63605779/autherror-error-amplify-has-not-been-configured-correctly
*Description of changes:* Added Auth.configure()

Not sure if this is a valid PR/fix but it resolves the issue with amplify config not being loaded. I'm not sure if this is actually just masking a deeper issue with the library itself though, or if it is just missing documentation.

Thanks

### DIFF
--- a/docs/lib/auth/fragments/js/getting-started.md
+++ b/docs/lib/auth/fragments/js/getting-started.md
@@ -43,6 +43,7 @@ In your app's entry point (i.e. __App.js__, __index.js__, or __main.js__), impor
 import Amplify, { Auth } from 'aws-amplify';
 import awsconfig from './aws-exports';
 Amplify.configure(awsconfig);
+Auth.configure(awsconfig);
 ```
 
 ## Enable sign-up, sign-in, and sign-out


ericclemmons (commented):
I appreciate the PR! But first, I want to look into the underlying issue you called out. 🧐 

`Amplify.configure` internally calls `Auth.configure`, so if both are required for things to work, then there's an underlying bug.

Since that stackoverflow thread mentions `@aws-amplify/ui-components`, it may be related to the duplicate dependency issue tracked in this milestone:

https://github.com/aws-amplify/amplify-js/milestone/32

anthonyhumphreys:
Cheers Eric, had a feeling that was the case. I'd not spotted the duplicate dependency issue. Will keep an eye on that - the fix is working locally for now, will leave a TODO to fix it when the bug is identified and patched.

Feel free to close this issue if you feel if appropriate.

ericclemmons:
@anthonyhumphreys I appreciate your positivity :) We'll get the duplicate issue resolved shortly, so I'm going to close this since we want to encourage a singular import (`aws-amplify`). (Especially to help customers avoid conflicting dependency versions)

**Note to any customers who find this in the future – if `Amplify.configure` isn't sufficient, there's a duplicate dependency issue that has to be resolved. Please let us know 🙏 **

PR #2875, opened 2021-01-22
Update auth-state-management.md

bradwebber:
Replaced ChangeDetectorRef implementation with NgZone.

*Issue #, if available:*
N/A

*Description of changes:*
Replaced ChangeDetectorRef with NgZone. 


ericclemmons (commented):
Thanks for sending this in!

As I'm not that familiar with Angular, is this change due to how `ng` is generating `AppComponent` by default now?

Or was there another reason for this change?

Thanks!

PR #2921, opened 2021-02-08
feat(ios-graphql): add pagination support in API

lawmicha:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/lib/graphqlapi/fragments/ios/query-data.md
+++ b/docs/lib/graphqlapi/fragments/ios/query-data.md
@@ -73,13 +73,12 @@ You can get the list of items that match a condition that you specify using the 
 func listTodos() {
     let todo = Todo.keys
     let predicate = todo.name == "my first todo" && todo.description == "todo description"
-    Amplify.API.query(request: .list(Todo.self, where: predicate)) { event in
+    Amplify.API.query(request: .paginatedList(Todo.self, where: predicate)) { event in
         switch event {
         case .success(let result):
             switch result {
             case .success(let todo):
                 print("Successfully retrieved list of todos: \(todo)")
-
             case .failure(let error):
                 print("Got failed result with \(error.errorDescription)")
             }
@@ -98,7 +97,7 @@ func listTodos() {
 func listTodos() -> AnyCancellable {
     let todo = Todo.keys
     let predicate = todo.name == "my first todo" && todo.description == "todo description"
-    let sink = Amplify.API.query(request: .list(Todo.self, where: predicate))
+    let sink = Amplify.API.query(request: .paginatedList(Todo.self, where: predicate))
         .resultPublisher
         .sink {
             if case let .failure(error) = $0 {
@@ -121,3 +120,41 @@ func listTodos() -> AnyCancellable {
 </amplify-block>
 
 </amplify-block-switcher>
+
+> **Note**: This approach will only return up to the first 1,000 items.  To change this limit or make requests for additional results beyond this limit, use *pagination* as discussed below.
+
+## List subsequent pages of items
+
+A list query only returns the first 1,000 items by default, so for large data sets, you'll need to paginate through the results.  After receiving a page of results, you can check if there are subsequent pages and obtain the next page. The page size is configurable as well, as in the example below.
+
+
+```swift
+func listTodos() {
+    let todo = Todo.keys
+    let predicate = todo.name == "my first todo" && todo.description == "todo description"
+    Amplify.API.query(request: .paginatedList(Todo.self, where: predicate, limit: 1000)) { event in
+        switch event {
+        case .success(let result):
+            switch result {
+            case .success(let todo):
+                print("Successfully retrieved list of todos: \(todo)")
+                if todo.hasNextPage() {
+                    todo.getNextPage() { result 
+                        switch result {
+                            case .success(nextPageTodo):
+                                print("Successfully retrieved next page of todos: \(nextPageTodo)")
+                            case .failure(let error):
+                                print("Got failed result with \(error.errorDescription)")
+                        }
+                    }
+                }
+            case .failure(let error):
+                print("Got failed result with \(error.errorDescription)")
+            }
+        case .failure(let error):
+            print("Got failed event with error \(error)")
+        }
+    }
+}
+```
+
--- a/docs/lib/graphqlapi/fragments/ios/query-data.md
+++ b/docs/lib/graphqlapi/fragments/ios/query-data.md
@@ -63,7 +63,7 @@ func getTodo() -> AnyCancellable {
 
 ## List Query
 
-You can get the list of items that match a condition that you specify using the `where` parameter in `Amplify.API.query`
+You can get the list of items using `.paginatedList` with optional parameters `limit` and `where` to specific the page size and condition. By default, the page size is 1000.
 
 <amplify-block-switcher>
 
@@ -73,7 +73,7 @@ You can get the list of items that match a condition that you specify using the 
 func listTodos() {
     let todo = Todo.keys
     let predicate = todo.name == "my first todo" && todo.description == "todo description"
-    Amplify.API.query(request: .paginatedList(Todo.self, where: predicate)) { event in
+    Amplify.API.query(request: .paginatedList(Todo.self, where: predicate, limit: 1000)) { event in
         switch event {
         case .success(let result):
             switch result {
@@ -97,7 +97,7 @@ func listTodos() {
 func listTodos() -> AnyCancellable {
     let todo = Todo.keys
     let predicate = todo.name == "my first todo" && todo.description == "todo description"
-    let sink = Amplify.API.query(request: .paginatedList(Todo.self, where: predicate))
+    let sink = Amplify.API.query(request: .paginatedList(Todo.self, where: predicate, limit: 1000))
         .resultPublisher
         .sink {
             if case let .failure(error) = $0 {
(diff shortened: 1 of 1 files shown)


docs/lib/graphqlapi/fragments/ios/query-data.md
```diff
@@ -121,3 +120,41 @@ func listTodos() -> AnyCancellable {
...
+## List subsequent pages of items
+
+A list query only returns the first 1,000 items by default, so for large data sets, you'll need to paginate through the results.  After receiving a page of results, you can check if there are subsequent pages and obtain the next page. The page size is configurable as well, as in the example below.
+
+
+```swift
+func listTodos() {
+    let todo = Todo.keys
+    let predicate = todo.name == "my first todo" && todo.description == "todo description"
+    Amplify.API.query(request: .paginatedList(Todo.self, where: predicate, limit: 1000)) { event in
```
jamesonwilliams, line 135: To reduce the nesting in this example, you could show this example using the Combine API. The various `switch`/`if` statements could be flattened to [`filter(_:)` calls](https://developer.apple.com/documentation/combine/publishers/merge/filter(_:)).

docs/lib/graphqlapi/fragments/ios/query-data.md
```diff
@@ -121,3 +120,41 @@ func listTodos() -> AnyCancellable {
 </amplify-block>
 
 </amplify-block-switcher>
+
+> **Note**: This approach will only return up to the first 1,000 items.  To change this limit or make requests for additional results beyond this limit, use *pagination* as discussed below.
+
+## List subsequent pages of items
```
richardmcclellan, line 126: This example only shows how to get 2 pages of results.  Can you provide an example of getting all pages?

docs/lib/graphqlapi/fragments/ios/query-data.md
```diff
@@ -121,3 +120,41 @@ func listTodos() -> AnyCancellable {
...
+func listTodos() {
+    let todo = Todo.keys
+    let predicate = todo.name == "my first todo" && todo.description == "todo description"
+    Amplify.API.query(request: .paginatedList(Todo.self, where: predicate, limit: 1000)) { event in
+        switch event {
+        case .success(let result):
+            switch result {
+            case .success(let todo):
+                print("Successfully retrieved list of todos: \(todo)")
+                if todo.hasNextPage() {
```
richardmcclellan, line 141: What is the type of `todo`?  Is it a single todo, or a list of todos?    Maybe this should be named `todos`?

docs/lib/graphqlapi/fragments/ios/query-data.md
lawmicha: thanks for the sample, i'll have to visit this usage again to be more familiar with it. In the latest revision, i took the approach to build a view model class that helps collect the next pages into an array. I've added the combine example, that only used for the first call, holds a reference to the `List<Todo>`, and uses it get check and get the next page

docs/lib/graphqlapi/fragments/ios/query-data.md
lawmicha: Updated example to contain both `getNextPage()` and `getAllPages()`

docs/lib/graphqlapi/fragments/ios/query-data.md
lawmicha: good callout- updating to plural

docs/lib/graphqlapi/fragments/ios/query-data.md
richardmcclellan: I think this example is a little confusing now because it is showing two use cases in one code block.  I'd suggest only showing how to retrieve all pages, or splitting this into two independent code blocks.  

On another note, the overall DX does not seem great.  Customers basically have to implement the success and failure callbacks twice - once for fetching the first page, and a second time for fetching subsequent pages.   I doubt there's a use case when the logic would differ for these scenarios.  Has this new API gone through an API review yet?

lawmicha: thanks for the feedback, i've updated the sample and separated them into two. one that shows how to load the next page and another that retrieves all pages at once

in regards to the DX, I was working around a few things in mind such as 
- how to support pagination from the list instance for both DataStore and API users. Developers can create a List object from Amplify.DataStore APIs and Amplify.API APIs. 
- thus, how we can share the error handling, which introduced a new Amplify CoreError type that wraps the APIError or DataStore error. This is probably the reason why the DX seems different across the two calls, ie. Amplify.API returns two layers of results (which i think is a painpoint that we want to revisit) and `getNextPage` returns a `Result<List, CoreError>`. 
I'd be happy to grab some of your time to iterate on the DX, and I've gone through a couple of design/implementation reviews with tim and daniel on this. I believe the lack of cross platform reviews was a conscious decision on iOS side to move forward with these APIs to have feature parity with Android as we recognized that the solution was becoming very iOS platform dependent (if you recall, i actually wrote some code that `getNextPage` returns a `GraphQLRequest` and at that point we met to discuss the need to iterate on the solution). I expect these APIs to change or influence the overall cross platform efforts for introducing lazy loading in the future (loading associated data from the list instance) since they are all methods part of the List type

docs/lib/graphqlapi/fragments/ios/query-data.md
```diff
@@ -121,3 +119,87 @@ func listTodos() -> AnyCancellable {
...
+
+### List subsequent pages of items
+
+For large data sets, you'll need to paginate through the results. After receiving the first page of results, you can check if there is a subsequent page and obtain the next page.
+
+```swift
+var todos: [Todo] = []
+var currentPage: List<Todo>?
+
+func listTodos() {
```
palpatim, line 131: To make the lifecycle clearer from an app developer perspective, maybe this could be `listFirstPage` or `loadFirstPage`, and the method below could have a parallel construction `listNextPage` or `loadNextPage` or similar?

docs/lib/graphqlapi/fragments/ios/query-data.md
```diff
@@ -121,3 +119,87 @@ func listTodos() -> AnyCancellable {
...
+
+## List all pages
+
+If you want to get all pages, first get the initial page and then use it to recursively check if there is a subsequent page before retrieving the next page, aggregating the data from each page into the `todos` array.
+
+```swift
+var todos: [Todo] = []
+var currentPage: List<Todo>?
+
+func getAllPages() {
```
palpatim, line 174: Seems like it would be just as easy to reuse the code above, but change the success handlers in both the "initial" and "next" methods to invoke the "loadNext" method? Fewer concepts, and arguably better separation of concerns. I'm not super happy with the amount of work we've got going on in this method.

docs/lib/graphqlapi/fragments/ios/query-data.md
lawmicha: updated to `listFirstPage` and `listNextPage` since the section is called "List Query"

docs/lib/graphqlapi/fragments/ios/query-data.md
lawmicha: ah nice, i've refactored to use the above code, and updated the `listNextPage` to be called `listNextPageRecursively`

docs/lib/graphqlapi/fragments/ios/query-data.md
```diff
@@ -121,3 +119,91 @@ func listTodos() -> AnyCancellable {
...
+                print("Failed to get next page \(coreError)")
+            }
+        }
+    }
+}
+```
+
+## List all pages
+
+If you want to get all pages, retrieve the subsequent page when you have successfully retrieved the first or next page.
```
palpatim, line 168: If the code change is as small as I think it is, you can make understanding clearer (at the expense of copy/pasteability) by just calling out the one added line in the `listAllPages()` method, and then listing the code for the recursive method. Something like 

    If you want to get all pages, you can adapt the above code to recursively call `getNextPage()` after you retrieve the first page of results.

    - Create a new method named `listAllTodos()`, with the same content as the `listFirstPage()` method above.
    - After you append the first page of results to `self.todos`, add a call to `self.listNextPageRecursively()`:
        ```diff
         self.currentPage = todos
         self.todos.append(contentsOf: todos)
        +self.listNextPageRecursively()
        ```
    - Add the new `listNextPageRecursively()` method:
        ```swift
        func listNextPageRecursively() {
            if let current = currentPage, current.hasNextPage() {
                current.getNextPage { result in
                    switch result {
                    case .success(let todos):
                        self.todos.append(contentsOf: todos)
                        self.currentPage = todos
                        self.listNextPageRecursively()
                    case .failure(let coreError):
                        print("Failed to get next page \(coreError)")
                    }
                }
            }
        }
        ```

(Or with appropriate adaptations to use Combine as @jamesonwilliams suggested)

palpatim (approved):
Continued refactoring suggestion, but not blocking if this is impractical or if the change is bigger than I thought

docs/lib/graphqlapi/fragments/ios/query-data.md
lawmicha: the latest improvement to this describes the steps and then provides the final code snippet to allow copypasta. I think this provides the benefit of explaining what was done in the code and having the entire code snippet, i'm able to mark the step in the code like

1. Update the above method `listFirstPage()` to `listAllPages()` 
```
func listAllPages() { // 1. Updated from `listFirstpage()`
```

PR #2933, opened 2021-02-11
Fixes #2919, removes duplicate config

### LINKED ISSUE #2919: [Feedback]FEEDBACK_TITLE_HERE
**Page**: [`/ui/auth/authenticator/q/framework/angular`](https://docs.amplify.aws/ui/auth/authenticator/q/framework/angular)

**Feedback**:
You have two different versions of the app.component.html file in the Recommended Usage section. Which is correct? Also, you have two identical app.module.ts files there. You can delete one of them.

dabit3:


mauerbac (approved):
LGTM

PR #2939, opened 2021-02-12
fix-gql-custom-claim

### LINKED ISSUE #2481: Use `identityClaim: "sub"` instead of `identityClaim: "user_id"`
**Describe the bug**
In the custom claims section of our documentation located here:
https://docs.amplify.aws/cli/graphql-transformer/directives#custom-claims

We specify `identityClaim: "user_id"`, but if customers want to use the user_id, they need to specify this as `identityClaim: "sub"`

**To Reproduce**
1. Go to https://docs.amplify.aws/cli/graphql-transformer/directives#custom-claims
2.  Notice that we use `identityClaim: "user_id"`

**Expected behavior**
It should be `identityClaim: "sub"` instead

**Additional context**

yuth:
Updated the custom claims section to inform user_id is not available in cognito generated token
and customers should use sub when the token is generated by Cognito

fix #2481


SwaySway (approved):
lgtm 🚀

renebrandel (approved):

Issue #2911, opened 2021-02-04
Add note about keychain behavior to Auth docs


palpatim:
Customers are sometimes confused that uninstalling an app does not clear auth data from the iOS keychain. (e.g., https://github.com/aws-amplify/amplify-ios/issues/1016). We need to add the following explanation to our Auth docs, and consider ways of handling this on behalf of customers in the future.

---

**Add AWS SDK Auth docs**

```markdown
NOTE: The AWS SDK for iOS stores auth information in the local system keychain (https://developer.apple.com/documentation/security/keychain_services), which does not guarantee any particular behavior around whether data is removed when an app is uninstalled.

Deciding on when to clear this auth information is not something that the SDK can do in a generic way, so App developers should decide when to clear the data by signing out. One strategy for accomplishing this would be to use [UserDefaults](https://developer.apple.com/documentation/foundation/userdefaults) to detect whether or not the app is launching for the first time, and invoking [`AWSMobileClient.signOut()`](https://docs.amplify.aws/sdk/auth/working-with-api/q/platform/ios#signout) if the app has not been launched before.
```

**Add to Amplify Auth docs**

```markdown
NOTE: Amplify stores auth information in the local system keychain (https://developer.apple.com/documentation/security/keychain_services), which does not guarantee any particular behavior around whether data is removed when an app is uninstalled.

Deciding on when to clear this auth information is not something that the SDK can do in a generic way, so App developers should decide when to clear the data by signing out. One strategy for accomplishing this would be to use [UserDefaults](https://developer.apple.com/documentation/foundation/userdefaults) to detect whether or not the app is launching for the first time, and invoking [`Auth.signOut()`](https://docs.amplify.aws/lib/auth/signOut/q/platform/ios) if the app has not been launched before.
```

palpatim:
Fixed by https://github.com/aws-amplify/docs/pull/2941

PR #2941, opened 2021-02-12
feat: Add uninstall behavior for iOS; add Info section to SDK tree

palpatim:
*Issue #, if available:*
https://github.com/aws-amplify/docs/issues/2911

*Description of changes:*
- Adds an "Uninstalling the app" section to the "Info" menu
- Adds the "Info" section to the SDK tree. The "Data Information" section pulls the fragment from the `lib` path, but the "Uninstalling the app" section is custom for the SDK. See screenshots

**Amplify**
amplify

**SDK**
sdk


docs/lib/info/fragments/ios/app-uninstall.md
```diff
@@ -0,0 +1,5 @@
+Some Amplify categories such as Analytics, Auth, and DataStore persist data to the local device. Some of that data is automatically removed when a user uninstalls the app from the device.
```
royjit, line 1: Should we also mention what data get lost when the app is uninstalled in this doc?

palpatim: I thought about it, but in the absence of a full enumeration of data that falls into that category, felt it was more useful to simply explicitly call out the data that is *not* erased.

royjit (approved):
LGTM

PR #2964, opened 2021-02-18
Updated storage docs to use async await

dabit3:
Updated JavaScript storage docs to be more consistent across pages, using async await vs regular promise.

### DIFF
--- a/docs/lib/storage/fragments/js/upload.md
+++ b/docs/lib/storage/fragments/js/upload.md
@@ -4,39 +4,31 @@ The `Put` method uploads files into Amazon S3.
 It returns a `{key: S3 Object key}` object on success:
 
 ```javascript
-Storage.put('test.txt', 'Hello')
-    .then (result => console.log(result)) // {key: "test.txt"}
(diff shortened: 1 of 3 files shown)


docs/lib/storage/fragments/js/upload.md
```diff
@@ -60,67 +52,58 @@ const SSECustomerAlgorithm = 'string';
...
+      contentType: 'image/png' // contentType is optional
+    });
+  } catch (err) {
+    console.log('Error uploading file: ', error);
+  }  
 }
+
+<input
+  type="file"
+  onChange={onChange}
```
sammartinez, line 90: Dont we need to call here? 
```suggestion
  onChange={this.onChange}
```

dabit3: So I tried to refactor out of anything framework specific, so if they are using React hooks then they should not have to worry about the scope or `this`

sammartinez (approved):
LGTM 🌮

PR #2982, opened 2021-02-23
fix: minor grammatical correction in cli docs

rajrajhans:
*Issue #, if available:*

*Description of changes:*
Minor grammatical correction in cli graphql transformer docs

### DIFF
--- a/docs/cli/graphql-transformer/directives.md
+++ b/docs/cli/graphql-transformer/directives.md
@@ -7,7 +7,7 @@ The Amplify CLI provides GraphQL directives to enhance your schema with addition
 
 ## Amplify-provided directives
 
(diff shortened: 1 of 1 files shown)


ammarkarachi:
Why do you think the language is wrong in this PR?

palpatim:
FWIW, I agree that this PR makes sense:
1. Fixes the plural agreement between verb and object: "a top level object type...that are backed"
2. Keeps the structure parallel with the other directives described in this section. The other directive definitions speak in general terms about plural objects ("Configures custom index structures", "Defines authorization rules", etc) while the current `@model` doc refers to a single element ("Defines a top level object type").

palpatim (approved):

ammarkarachi:
@palpatim I agree to we had a discussion offline about this and LGTM too

ammarkarachi (approved):

PR #2987, opened 2021-02-25
fix: make the link inside warning box clickable

changxu0306:
Now, the link inside the warning box under Android getting started - Install Amplify libraries is not clickable.
Screen Shot 2021-02-25 at 12 59 49 AM

This PR puts the link inside an HTML <a> tag so that the link is clickable.


jamesonwilliams (approved):
Nice! I was wondering how we could fix it. :-)

PR #2990, opened 2021-02-26
fix: add error event in API subscription sample code

nubpro:
*Issue #, if available:*
https://discord.com/channels/705853757799399426/717109993416949811/813523850523836416
Some people has been getting `uncaught error` when the subscription fails. This is because they did not include the `error` event in their code as this is not demonstrated in the doc.

*Description of changes:*
add `error` event into sample code


amhinson (approved):
Thanks @nubpro! 👍  LGTM

PR #2998, opened 2021-02-26
fix(native): note that auth can be used without identity pool

lawmicha:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-ios/issues/1040#issuecomment-786949951

*Description of changes:*

### DIFF
--- a/docs/lib/auth/fragments/existing-resources.md
+++ b/docs/lib/auth/fragments/existing-resources.md
@@ -46,3 +46,5 @@ Existing Amazon Cognito identity pools and user pools can be used with the Ampli
     - **Region**: AWS Region where the resources are provisioned (e.g. `us-east-1`)
     
 Note that before you can add an AWS resource to your application, the application must have the Amplify libraries installed. If you need to perform this step, see [Install Amplify Libraries](~/lib/project-setup/create-application.md#n2-install-amplify-libraries). 
+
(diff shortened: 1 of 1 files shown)


docs/lib/auth/fragments/existing-resources.md
```diff
@@ -46,3 +46,5 @@ Existing Amazon Cognito identity pools and user pools can be used with the Ampli
     - **Region**: AWS Region where the resources are provisioned (e.g. `us-east-1`)
     
 Note that before you can add an AWS resource to your application, the application must have the Amplify libraries installed. If you need to perform this step, see [Install Amplify Libraries](~/lib/project-setup/create-application.md#n2-install-amplify-libraries). 
+
+If you are using Cognito user pool, without Cognito identity pool, you can omit the **CredentialsProvider** section in the configuration.
```
palpatim, line 50:
```suggestion
If you are using a Cognito User Pool without a Cognito Identity Pool, you can omit the **CredentialsProvider** section in the configuration.
```

Also, while we're in this section, could you fix the capitalization on Line 1 above?

palpatim (approved):
Approved w/minor changes

docs/lib/auth/fragments/existing-resources.md
lawmicha: updated this line and line 1. for some reason I thought "user pool" and "identity pool" should be lower case, but now i can't remember why

PR #3008, opened 2021-03-03
Coroutines project setup tweaks

jamesonwilliams:

### DIFF
--- a/docs/lib/project-setup/fragments/android/coroutines/coroutines.md
+++ b/docs/lib/project-setup/fragments/android/coroutines/coroutines.md
@@ -25,7 +25,7 @@ lifecycleScope.launch {
 
 Coroutines can greatly improve the readability of dependent, asynchronous calls. Moreover, you can use scopes, dispatchers, and other Kotlin coroutine primitives to get more control over your execution context.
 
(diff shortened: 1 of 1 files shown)


docs/lib/project-setup/fragments/android/coroutines/coroutines.md
```diff
@@ -23,27 +23,24 @@ lifecycleScope.launch {
 }
 ```
 
-Compared to the traditional callback API, this doesn't make a big difference when used for a single method call.
+Coroutines can greatly improve the readability of dependent, asynchronous calls. Moreover, you can use scopes, dispatchers, and other Kotlin coroutine primitives to get more control over your execution context.
 
-However, it greatly improves readability when chaining asynchronous calls. Moreover, you can use scopes, dispatchers, and other Kotlin coroutine primitives to get more control over your execution context.
-
-Let's revisit our nested example where we saved `Post`, `Editor`, and `PostEditor`. With Amplify's coroutines interface, we can write these operations sequentially:
+Let's consider what happens when we have three dependency operations. We want to save a `Post`, then an `Editor`, and finally a `PostEditor`. With Amplify's coroutines interface, we can write these operations sequentially:
```
richardmcclellan, line 28:
```suggestion
Let's consider what happens when we have three dependent operations. We want to save a `Post`, then an `Editor`, and finally a `PostEditor`. With Amplify's coroutines interface, we can write these operations sequentially:
```

richardmcclellan (approved):

PR #3009, opened 2021-03-03
Ensure API category is added before pushing to Amplify

Kilo-Loco:
*Description of changes:*
This PR is focused on ensuring that the API category is included in the local Amplify app before pushing to the cloud. If the customer is missing the category for any reason, they will be blocked from syncing data to the cloud.

Before:
Screen Shot 2021-03-03 at 9 40 27 AM

After:
Screen Shot 2021-03-03 at 9 40 09 AM


palpatim (approved):

docs/start/getting-started/fragments/ios/add-api.md
```diff
@@ -21,13 +21,27 @@ Now that you have DataStore persisting data locally, in the next step you'll con
...
+  This command will display a summary of the changes that will be pushed to your backend. It should look like this:
 
-    <amplify-callout>
-    
-    **Note:** sit back and relax since this command will generate all the required cloud resources on your AWS account and it might take a while to complete.
+  ![](~/images/lib/getting-started/ios/amplify-push-api-summary.png)
+
+  <amplify-callout>
+
+  **Note:** If the summary doesn't contain the API category, you will need to cancel this push process and run `$ amplify add api` before continuing.
```
drochetti, line 31: I have a few concerns with this, because `amplify add api` is not enough. The developer needs to pick specific options in order for DataStore to work (GraphQL, enable conflict resolution).

I believe the fact that someone ignored the `--quickstart` means we could improve the first steps, but I'm not sure this is the right fix for that.

drochetti (changes requested):
I'd like us to take the opportunity to dive deeper on the issues our tutorial has that would require that call-out.

Unfortunately `amplify add api` is not enough and if we go that route we need to provide more detailed instructions, like the ones here: https://docs.amplify.aws/lib/datastore/getting-started/q/platform/ios#setup-local-development-environment

Kilo-Loco:
> I'd like us to take the opportunity to dive deeper on the issues our tutorial has that would require that call-out.
> 
> Unfortunately `amplify add api` is not enough and if we go that route we need to provide more detailed instructions, like the ones here: https://docs.amplify.aws/lib/datastore/getting-started/q/platform/ios#setup-local-development-environment

This is a good point. I don't think it makes sense to list out all these steps in the event the user skipped around since this page is part of a sequence of steps. Instead, linking to the same link you provided would address both the concern of API not being an added category as well as ensuring ALL the appropriate steps and considerations are listed thoroughly.

So then maybe the callout should read more along the lines of:

> **Note:** If the summary doesn't contain the API category, you will need to cancel this push process and [add the API category](https://docs.amplify.aws/lib/datastore/getting-started/q/platform/ios#setup-local-development-environment) before continuing.

siegerts:
Closing for now after sync with @Kilo-Loco.

PR #3013, opened 2021-03-04
Update setup.md

evcodes:
Updating getting started guide's gradle dependencies with the Kotlin Amplify Facade dependency.


jamesonwilliams (changes requested):
Thanks for the bias for action!

I'm going through and adding the callback snippets back to all of the categories.

I think we should leave the Getting Started guide as is for now. I also pushed an update to ensure that the Getting Started guide shows only the callback snippets.

jamesonwilliams:
Closing for now, we'll revisit using the Kotlin module in the Getting Started guide at a later time.

PR #3015, opened 2021-03-04
Android Auth: Re-introduce Kotlin callbacks alongside coroutines

jamesonwilliams:
**This PR update the Auth docs for Android.**

Until the coroutines support achieves 1.0, we will document both types of Kotlin interaction alongside each other:

 1. Using Kotlin with the traditional callback API
 2. Using Kotlin with the Coroutine support


richardmcclellan (approved):
🚢

PR #3023, opened 2021-03-05
Bump Android Dependency Versions

jamesonwilliams:
Amplify Android 1.17.0
Android Gradle Plugin 4.1.2
Core Desugaring Library 1.1.5

### DIFF
--- a/docs/lib/datastore/fragments/android/getting-started/30_platformIntegration.md
+++ b/docs/lib/datastore/fragments/android/getting-started/30_platformIntegration.md
@@ -13,7 +13,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.1.1'
+        classpath 'com.android.tools.build:gradle:4.1.2'
         // Add this next line:
         classpath 'com.amplifyframework:amplify-tools-gradle-plugin:1.0.2'
     }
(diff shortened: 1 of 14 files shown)


richardmcclellan (approved):

docs/lib/datastore/fragments/android/getting-started/30_platformIntegration.md
```diff
@@ -13,7 +13,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.1.1'
+        classpath 'com.android.tools.build:gradle:4.1.2'
```
rjuliano, line 16: Is this the guy the breaks the build or did they finally patch it?

rjuliano: Not sure if the issue we observed when attempting to move to 4.1.2 for the library itself also happens when users are creating an app.

jamesonwilliams: This breaks the build _of our library itself_, but it should work fine for a consumer of our library. (I've been using it myself, in a sample app.)

richardmcclellan: The issue we are running into with the library is trying to jump from 4.0.x to 4.1.x, not 4.1.1 to 4.1.2, so I think this change is safe.

PR #3049, opened 2021-03-12
Fix reference to AmplifyTotpSetup.

andrewrmoreland:
*Issue #, if available:*

Corrected reference to AmplifyTotpSetup component which was incorrectly capitalised as AmpliftyTOTPSetup.


jamesonwilliams:
Wasn't clear to me from looking at amplify-js repo -- both stylizations are used. @aws-amplify/amplify-js this one needs your action.

wlee221 (approved):
Hi! Sorry about the response time. I checked `AmplifyTotpSetup` is the correct name. Thanks for the catch and pr!

PR #3054, opened 2021-03-16
Purge @ts-ignore statements by declaring types on globals

tjleing:
New global objects:
- Adobe `s` object
- Algolia docsearch object
- AWS Shortbread object

Also `ts-ignore` on Stencil `Watch` and `Listen` statements isn't necessary anymore.

### DIFF
--- a/client/src/globals.d.ts
+++ b/client/src/globals.d.ts
@@ -1,0 +1,28 @@
+interface AdobeS {
+  // Configuration properties
+  trackExternalLinks: boolean;
+
+  // Variables to set when tracking
+  linkTrackVars: string;
+  linkTrackEvents: string;
+  events: string;
+  pageURL: string;
+  eVar26: string;
+  eVar27: string;
+
+  // Tracking functions
+  t: () => void;
+  tl: (
+    linkObject: true | undefined,
+    linkType: string,
+    linkName: string,
+  ) => void;
+}
+
+interface AWSCShortbreadObject {
+  checkForCookieConsent: () => void;
+}
+
+declare const s: AdobeS;
+declare const docsearch: (obj: object) => void;
+declare const AWSCShortbread: (obj: object) => AWSCShortbreadObject;
--- a/client/src/utils/track.ts
+++ b/client/src/utils/track.ts
@@ -9,13 +9,11 @@ if (!configured) {
   Auth.configure(awsexports);
   Analytics.configure(awsexports);
   if (Build.isBrowser) {
-    // @ts-ignore
     AWSCShortbread({
       domain: ".amplify.aws",
     }).checkForCookieConsent();
   }
   if (Build.isBrowser) {
-    // @ts-ignore
     if (typeof s != "undefined") s.trackExternalLinks = false;
   }
   configured = true;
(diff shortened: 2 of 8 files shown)


client/src/utils/track.ts
```diff
@@ -78,87 +76,62 @@ export const track = (event: AnalyticsEvent): Promise<unknown> | undefined => {
...
     s.eVar26 = query;
   }
 };
 
 const triggerNoSearchResults = (query: string): void => {
-  // @ts-ignore
   const queryBackup: string = s.eVar26;
-  // @ts-ignore
-  const resultCountBackup: number = s.eVar27;
+  const resultCountBackup: number = parseInt(s.eVar27);
```
jakeburden, line 110: **Suggestion**:

What do you think about passing `10` as the second argument here to be sure the string is always being parsed as a base 10 number?  From MDN, [`10` is not always the default value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#parameters). My guess is that the default radix depends on the browser vendor implementation, so I think it would be safer to be explicit here.

```suggestion
  const resultCountBackup: number = parseInt(s.eVar27, 10);
```

client/src/utils/track.ts
```diff
@@ -78,87 +76,62 @@ export const track = (event: AnalyticsEvent): Promise<unknown> | undefined => {
...
-  // @ts-ignore
   s.events = "event2";
-  // @ts-ignore
   s.tl(true, "o", "internal search");
 
-  // @ts-ignore
   s.eVar26 = queryBackup;
-  // @ts-ignore
-  s.eVar27 = resultCountBackup;
+  s.eVar27 = resultCountBackup.toString();
```
jakeburden, line 121: **Question:**

I'm not too sure if we need `resultCountBackup`. Since `eVar27` is defined as a [string type here](https://github.com/aws-amplify/docs/blob/ed898cde261d4a9f74b67ce3e44ca6aad30bd3da/client/src/globals.d.ts#L11), do we need this line?

It looks like we're setting `resultCountBackup` to the number parsed from the string `s.eVar27`, and the setting `s.eVar27` back to the stringified value of the `resultCountBackup` number.

Let me know if I'm missing something 😄

client/src/globals.d.ts
```diff
@@ -0,0 +1,28 @@
...
+    linkType: string,
+    linkName: string,
+  ) => void;
+}
+
+interface AWSCShortbreadObject {
+  checkForCookieConsent: () => void;
+}
+
+declare const s: AdobeS;
```
jakeburden, line 26: *nitpick (non-blocking)*:

It's a difficult to know what `t`, `tl`, and `s` are supposed to mean as an outside contributor. Can we open a follow-up issue to replace these with more meaningful variable names?

jakeburden (changes requested):
Thanks @tjleing! This change looks great. I left one quick suggestion, and a couple questions.

Back to you 🏓

client/src/utils/track.ts
tjleing: Yeah, `resultCountBackup` is a little convoluted.  Here's the thought process.

1. We want to have a specific event to track search queries that have no results.
2. But we _don't_ want to send too many Adobe requests at once, because we pay per request or something like that.
3. So we really want to avoid a situation where a user types something like `dtastore` or pasting in a big block of text accidentally where it could just continue sending "no results found" events.
4. Solution: set a 1-second timer whenever a search query has no results, just refresh it every time the query changes, and only fire the event when the timer finishes.
5. Problem: now what if someone types out a query with no results, and then quickly deletes it, types in a new query, and selects a page (thus firing a "search query clicked" event)?  Well, one thing that can happen (depending on the order of everything) is that the event timer finishes, firing the "0 search results with query asdfgh" event, and then **the query and result count aren't updated** before the user selects the page.
6. Solution: in the callback called when the timer expires, back up the current search query and result count, use `eVar26` and `eVar27` to hold the values we received earlier when there were no results, and then restore the two `eVars` back to their original values.

So, ultimately, the problem arises because we're using `eVar26` and `eVar27` both to pass to Adobe Analytics (because that's just how you specify these things) but also to store the current values of two variables.  The problem would be solved with globals instead, but that seemed completely not worth it at the time.

As a side note, `eVar27` can't be a number because of "the '0' thing" -- passing the number 0 to Adobe Analytics results in it just being dropped, even when you literally want it to send the number 0.  The **official workaround** is to pass '0'.

Anyways, hope this helps you understand what's going on here -- if you have any more questions or ideas on how to do this whole thing more easily, please let me know!

client/src/globals.d.ts
tjleing: Heh, if only...  Don't worry, we've tried 😆

For real, this is just **required by Adobe Analytics**.  The meanings are all provided in their docs, but our internal doc can help provide a lot more detail ("Amplify Documentation: Adobe Analytics Dev Guide").

client/src/utils/track.ts
tjleing: Yeah, I totally forgot about that!

client/src/globals.d.ts
jakeburden: Thanks! I just looked through the docs. It makes sense where these are coming from now!

client/src/utils/track.ts
jakeburden: Ah, okay I think I'm following whats going on here. I'm probably going to need to spend some more time digging into this to fully grok the process, but for the sake of this review I think the changes here are consistent with how everything works. I really appreciate all the context provided here!


Speaking _very loosely_, I wonder if using a [`debounce`](https://www.freecodecamp.org/news/javascript-debounce-example/) function would be useful for simplifying this process. Debouncing would allow us to wait after a period of idle time from when the last character was entered in the search input before we fire the search query clicked event. If the user enters a new character (or deletes characters) before the idle timeout expires, the timeout is re-started.

I haven't spent much time looking at Adobe Analytics fully yet, so my suggestion might not be the direction to go, I just wanted to send the idea out just in case it sounds like something potentially useful :D

jakeburden (approved):
Thanks @tjleing for providing helpful context in the review comments. These changes look good to me! 🚀

PR #3067, opened 2021-03-18
Bump Android version

changxu0306:
Amplify Android 1.17.1
Core-kotlin 0.1.2


richardmcclellan (approved):

jamesonwilliams:
@tjleing @sammartinez does the docs site have a macro we can use for this? To date we just cut PRs to update the version, but it would make a nice variable.

tjleing:
@jamesonwilliams no macro as far as I know, and it could be pretty tricky to implement that sort of thing with `capi`.  However, it's useful feedback to have for our Next.js rewrite -- even just having all of these versions in one place could be a great improvement.

richardmcclellan:
> However, it's useful feedback to have for our Next.js rewrite -- even just having all of these versions in one place could be a great improvement.

Yes, please do, this would be great! 😀

PR #3072, opened 2021-03-22
feat: amplify-flutter isConfigured getter

haverchuck:
*Description of changes:*
Adds documentation for amplify-flutter 'isConfigured' API.

### DIFF
--- a/docs/lib/project-setup/fragments/flutter/create-application/40_verifyAmplifyLibraries.md
+++ b/docs/lib/project-setup/fragments/flutter/create-application/40_verifyAmplifyLibraries.md
@@ -47,8 +47,15 @@ class _MyHomePageState extends State<MyHomePage> {
 
 Note that all calls to `addPlugin()` or `addPlugins()` are made before `Amplify.configure()` is called.
 
-If you need to test whether Amplify has already been configured successfully, you can use the `isConfigured` getter:
+Currently Amplify cannot be configured more than once. You can use the `isConfigured` getter to check whether Amplify has already been configured successfully, which can be helpful if you are observing `AmplifyAlreadyConfiguredException` upon a hot restart.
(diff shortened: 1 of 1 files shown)


Ashish-Nanda (changes requested):
We need to update the code snippets as well for the getting started and category setups.

docs/lib/project-setup/fragments/flutter/create-application/40_verifyAmplifyLibraries.md
```diff
@@ -46,3 +46,9 @@ class _MyHomePageState extends State<MyHomePage> {
 ```
 
 Note that all calls to `addPlugin()` or `addPlugins()` are made before `Amplify.configure()` is called.
+
+If you need to test whether Amplify has already been configured successfully, you can use the `isConfigured` getter:
+
+```dart
+ var isConfigured = Amplify.isConfigured;
```
drochetti, line 53: general question, I was playing with Amplify/Flutter and realized that in our docs we're not very consistent when declaring variables, like recommended here: https://dart.dev/guides/language/effective-dart/usage#do-follow-a-consistent-rule-for-var-and-final-on-local-variables

it seems like `final` with type inference should be used more often, but I see `var` and the explicit type declared often in our snippets. Do we follow a pattern?

haverchuck: @drochetti We probably do not follow a consistent pattern (but obviously we should).

haverchuck:
@Ashish-Nanda I don't know that we need to clutter all of the category documentation with this check. Perhaps we can make that change if customers continue to have issues?

Ashish-Nanda (approved):

PR #3095, opened 2021-03-31
fix(cli/codegen): Migrate `--max-depth` flag to `--maxDepth`

medelman17:
*Issue #, if available:* closes aws-amplify/docs#2720

see also: aws-amplify/amplify-codegen#138

*Description of changes:*

Previously, users could execute the amplify codegen command with a --max-depth flag. Accordingly, when users would run amplify codegen --help the above usage would be described. However, at some point in 2019, the above flag was changed to --maxDepth (see aws-amplify/amplify-cli#2175). To date, neither the CLI help nor the official documentation reflect this change.

 

This PR addresses the official documentation.

### DIFF
--- a/docs/cli/graphql-transformer/codegen.md
+++ b/docs/cli/graphql-transformer/codegen.md
@@ -685,3 +685,5 @@ public class MainActivity extends AppCompatActivity {
     };
 }
 ```
+
+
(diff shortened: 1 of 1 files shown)


drochetti (approved):
Thanks for catching that and for fixing it! 🚀 

We should also engage someone from the CLI team to make a code change on the CLI to keep it backwards compatibility - both flags should be supported, with a deprecation message when the deprecated one is used until a next major release can remove it. @renebrandel

tjleing (approved):

ammarkarachi (approved):

PR #3100, opened 2021-04-01
fix(codeblock-switcher): Remedy inappropriate double-block rendering

### LINKED ISSUE #3091: Extra code blocks showing sometimes
**Describe the bug**
On some pages, more than one language's code blocks are rendering, one after the other.  Thanks for the callout @drochetti and @ericclemmons!

**To Reproduce**
Go to [this link](https://docs.amplify.aws/lib/analytics/identifyuser/q/platform/android).  Pick RxJava.  Refresh.

**Screenshots**
image

**Additional context**
Still looking at the DOM and CSS, but it doesn't quite look like the same `<slot>` issue as #3079.  Will link PR when created.

medelman17:
*Issue #, if available:* closes #3091

*Description of changes:*

This PR overhauls the way the `CodeblockSwitcher` component interacts with its `Codeblock` children. Specifically, with respect to how active children are identified/displayed. Previously, during render, the `CodeblockSwitcher` would generate and inject a css-selector based using the `:nth-child(i)` pseudo selector (where `i` was the index of the active child). This worked well enough in theory, but created a problem during pre-rendering where two versions of the selector would be created `:nth-child(<default-code-block>)` and `:nth-child(<selected-code-block>)`. As a result, as described in #3091, two code blocks would be displayed upon initial page-load. This PR addresses that issue. 

To fix the above, a single, always-existing, `.activeCodeBlock` css rule/selector was created. Control over whether that class would be applied to a given component was pushed down to the children. To facilitate the above process, the `CodeblockSwitcher` component exposes, via `data-*` attribute the initial `active-tab`. Using lifecycle methods, each child (before and after it loads) finds its parent `CodeblockSwitcher` and retrieves the value of the `active-tab` attribute. If the attribute value matches the child's name, the child sets its state to `isActive` = `true`, otherwise `isActive` is set to `false`. The `isActive` state determines whether a given component assigns itself the `activeCodeBlock` class. This effectively solves the above, double-block issue. 

However, the above approach introduced additional complexity in that future state changes to the `activeChild` index in the `CodeblockSwitcher` parent (e.g., when someone selects a new language to display) would no longer have any appreciable effect on child codeblocks. To overcome this, a connection between the parent and children was created using an `EventEmitter`. Now, when the `activeChild` is changed in the parent, that change is broadcast via an `active-codeblock-updated` event to which all children subscribe. When an event is observed in the child, the child checks to see whether the event detail matches its name and a similar process as described above is undertaken with respect to updating the child's `isActive` state. 

### DIFF
--- a/client/src/amplify-ui/block-switcher/block-switcher.style.ts
+++ b/client/src/amplify-ui/block-switcher/block-switcher.style.ts
@@ -48,4 +48,8 @@ export const contentStyle = css`
       margin-bottom: 0;
     }
   }
+
+  .activeCodeBlock {
+    display: initial;
+  }
 `;
--- a/client/src/amplify-ui/block-switcher/block-switcher.tsx
+++ b/client/src/amplify-ui/block-switcher/block-switcher.tsx
@@ -1,4 +1,14 @@
-import {Component, Host, h, Prop, State, Element, Watch} from "@stencil/core";
+import {
+  Component,
+  Host,
+  h,
+  Prop,
+  State,
+  Element,
+  Watch,
+  Event,
+  EventEmitter,
+} from "@stencil/core";
 import {
   hostStyle,
   activeTabStyle,
@@ -6,7 +16,6 @@ import {
   tabContainerStyle,
   contentStyle,
 } from "./block-switcher.style";
-import {css} from "emotion";
 import {pageContext} from "../../docs-ui/page/page.context";
 import {
   SelectedTabHeadings,
@@ -32,6 +41,7 @@ export class AmplifyBlockSwitcher {
 
   componentWillRender() {
     this.gatherHeadings();
+    this.handleInitSelectedTabHeading();
   }
 
   // recursively traverse the DOM tree to gather the language names from all child code blocks.
@@ -49,24 +59,43 @@ export class AmplifyBlockSwitcher {
     }
   }
 
+  handleInitSelectedTabHeading() {
+    if (this.tabHeadings?.length > 0 && this.selectedTabHeadings?.length > 0) {
+      for (const heading of this.selectedTabHeadings) {
+        if (this.tabHeadings.includes(heading)) {
+          this.activeChildI = this.tabHeadings.indexOf(heading);
+          return;
+        }
+      }
+    } else {
+      if (typeof window === "undefined") {
+        this.activeChildI = 0;
+      } else {
+        this.activeChildI = 0;
+      }
+    }
+  }
+
   gatherHeadings() {
     // gather tab headings from child `amplify-block` attrs
     this.tabHeadings = [];
     this.recursivelyFindBlocks(this.el);
+  }
 
-    // default to the first tab
-    this.activeChildI = 0;
+  /* eslint-disable */
+  @Event({
+    eventName: "active-codeblock-updated",
+    composed: true,
+    cancelable: true,
+    bubbles: true,
+  })
+  activeBlockUpdated: EventEmitter<string>;
 
-    if (this.tabHeadings?.length > 0 && this.selectedTabHeadings?.length > 0) {
-      // iterate through previous selections, and see if we have any shared
-      for (let i = 0; i < this.selectedTabHeadings.length; i++) {
-        const current = this.selectedTabHeadings[i];
-        if (this.tabHeadings.includes(current)) {
-          // if so, let's set its index in state!
-          this.activeChildI = this.tabHeadings.indexOf(current);
-          break;
-        }
-      }
+  @Watch("activeChildI")
+  activeChildWatcher(newValue: number, oldValue: number) {
+    if (newValue !== oldValue) {
+      const newEventDetail = this.tabHeadings[newValue];
+      this.activeBlockUpdated.emit(newEventDetail);
     }
   }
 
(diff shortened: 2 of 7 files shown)


tjleing:
Just really quickly before I review the code: `yarn test -u` to update the snapshots!

client/src/amplify-ui/block-switcher/block-switcher.tsx
```diff
@@ -81,42 +110,38 @@ export class AmplifyBlockSwitcher {
     }
   }
 
-  createActiveChildISetter = (i: number) => (_event: Event): void => {
-    this.activeChildI = i;
-    const current = this.tabHeadings[i];
-    this.setNewSelectedTabHeadings(current);
-  };
+  handleButtonClick(newActiveChild: number) {
```
tjleing, line 113: This reads more cleanly than the curried version we had before imo 😄

client/src/amplify-ui/block-switcher/block-switcher.tsx
```diff
@@ -49,24 +59,43 @@ export class AmplifyBlockSwitcher {
     }
   }
 
+  handleInitSelectedTabHeading() {
```
tjleing, line 62: Could this be just `initSelectedTabHeading()`?

client/src/amplify-ui/block-switcher/block-switcher.tsx
```diff
@@ -49,24 +59,43 @@ export class AmplifyBlockSwitcher {
...
+        if (this.tabHeadings.includes(heading)) {
+          this.activeChildI = this.tabHeadings.indexOf(heading);
+          return;
+        }
+      }
+    } else {
+      if (typeof window === "undefined") {
+        this.activeChildI = 0;
+      } else {
+        this.activeChildI = 0;
```
tjleing, line 74: Collapse to just `this.activeChildI = 0` with no `if` statement?  Or is this second one supposed to be something different...

client/src/amplify-ui/block-switcher/block-switcher.tsx
```diff
@@ -49,24 +59,43 @@ export class AmplifyBlockSwitcher {
...
   gatherHeadings() {
     // gather tab headings from child `amplify-block` attrs
     this.tabHeadings = [];
     this.recursivelyFindBlocks(this.el);
+  }
 
-    // default to the first tab
-    this.activeChildI = 0;
+  /* eslint-disable */
+  @Event({
```
tjleing, line 86: Really surprised that this needs a lint ignore statement; I was pretty sure that both `@Watch` and `@Event` statements were fine with respect to linting -- in any case, I think we usually use `// @ts-ignore` instead!

client/src/amplify-ui/block-switcher/block-switcher.style.ts
```diff
@@ -48,4 +48,8 @@ export const contentStyle = css`
       margin-bottom: 0;
     }
   }
+
+  .activeCodeBlock {
+    display: initial;
+  }
```
tjleing, line 54: Should this be in `block` CSS instead?

tjleing: Not sure if there's a file for that yet, but maybe there should be?

tjleing (commented):
I really like the approach you came up with!  Hopefully I can figure out what I'm doing wrong with my local copy 😅

docs/cli/graphql-transformer/codegen.md
```diff
@@ -11,7 +11,7 @@ When a project is configured to generate code with codegen, it stores all the co
 
 ## Statement depth
 
-In the below schema there are connections between `Comment` -> `Post` -> `Blog` -> `Post` -> `Comments`. When generating statements codegen has a default limit of 2 for depth traversal. But if you need to go deeper than 2 levels you can change the max-depth parameter either when setting up your codegen or by passing  `--max-depth` parameter to `codegen`
+In the below schema there are connections between `Comment` -> `Post` -> `Blog` -> `Post` -> `Comments`. When generating statements codegen has a default limit of 2 for depth traversal. But if you need to go deeper than 2 levels you can change the `maxDepth` parameter either when setting up your codegen or by passing  `--maxDepth` parameter to `codegen`
```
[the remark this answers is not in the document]
medelman17: Definitely need to rebase, here. Just didn't have the will/energy to do it yesterday.

client/src/amplify-ui/block-switcher/block.tsx
```diff
@@ -1,13 +1,53 @@
...
+
+  componentDidLoad() {
+    this.checkParentSwitcherAttributes();
+  }
+
+  checkParentSwitcherAttributes() {
+    if (typeof window !== "undefined") {
+      const switcher = this.el.closest("amplify-block-switcher");
+      const {value: activeTab} =
+        switcher?.attributes.getNamedItem("data-active-tab") ?? {};
```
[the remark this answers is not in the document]
medelman17: Reinventing ReactJS Context API with Webcomponents 😂

client/src/amplify-ui/block-switcher/block-switcher.tsx
medelman17: Agree. And wanted to eliminate as much potential for stale closures as possible since I have a child's understanding of how webcomponents work

client/src/amplify-ui/block-switcher/block.tsx
```diff
@@ -1,13 +1,53 @@
...
+  componentWillLoad() {
+    this.checkParentSwitcherAttributes();
+  }
+
+  componentDidLoad() {
+    this.checkParentSwitcherAttributes();
+  }
+
+  checkParentSwitcherAttributes() {
+    if (typeof window !== "undefined") {
```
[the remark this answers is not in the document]
medelman17: Did not know this API existed. I defer to you.

client/src/amplify-ui/block-switcher/block.tsx
```diff
@@ -1,13 +1,53 @@
...
+      const switcher = this.el.closest("amplify-block-switcher");
+      const {value: activeTab} =
+        switcher?.attributes.getNamedItem("data-active-tab") ?? {};
+
+      if (activeTab === this.name) {
```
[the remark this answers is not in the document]
medelman17: 🤷 

I was in and out, down and dirty style. I leave it to you. To the extent that we don't plan to maintain this code much longer, query whether a 'slightly-hacky' approach might provide best ROI?

client/src/amplify-ui/block-switcher/block-switcher.style.ts
medelman17: 🤷

I was in and out, down and dirty style. I leave it to you. To the extent that we don't plan to maintain this code much longer, query whether a 'slightly-hacky' approach might provide best ROI?

client/src/amplify-ui/block-switcher/block-switcher.tsx
medelman17: Agree. That's a better name.

client/src/amplify-ui/block-switcher/block-switcher.tsx
medelman17: 🤷

Apparently, the `stencil` lint plugin wants the `activeBlockUpdated: EventEmitter<string>` bit to be on the same line as the @Event decorator... but `prettier` don't like that.

client/src/amplify-ui/block-switcher/block-switcher.tsx
medelman17: Yeah. This should be cleaned up. Originally, I was setting the value to some nonsense number if outside of browser context, but then switched back and forgot to clean up.

client/src/amplify-ui/block-switcher/block.tsx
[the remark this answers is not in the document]
medelman17: I dig it. Inline IFFEs FTW.

tjleing:
Been looking at the reason it isn't working on my machine today.  Looks like the CSS class `activeCodeBlock` is still being applied despite the reinvention of the React context...

But the progress is that now that you used the `closest` DOM function, the CSS we have to reinvent is only displaying at all instead of having to reinvent the CSS `nth-child`.  Going to look into just removing the slot depending on the `activeCodeBlock` bool, and see what happens there...

tjleing:
Removing the `<slot>` doesn't work -- it still renders the content that would've gone into the slot.

tjleing:
Closing due to #3147 being merged!

Issue #3082, opened 2021-03-23
[missing-page] Push Notifications Docs


armenr:
**Page**: [`https://docs.amplify.aws/lib/push-notifications/getting-started/q/platform/js`](https://docs.amplify.aws/lib/push-notifications/getting-started/q/platform/js)

**Feedback**: 
Uncertain why or how, but looks like the docs for Push notifications don't exist/have gone missing.

medelman17:
Hi @armenr, can you be a bit more specific? If I travel to the URL you posted above, everything appears to be working properly. What do you see?

armenr:
Caching issue. Please disregard. Very bizarre. Thank you! I apologize for wasting your time!

medelman17:
Haha. No problem. Never a waste of time. You sure it was on your end and not ours?

armenr:
Pretty sure! Checked with an InCognito window, and all was well. No clue what the deal is, but I know for a fact that FireFox and/or Brave appear increasingly attractive to me, with each passing day. Chrome is just...*sigh*

Thanks again! :) You're awesome!

medelman17:
Are your running any ad blocking software in Chrome, by chance? If so, could you try and disable and see if it works as expected? We're having some issues related to the above and my radar is going off :-)

armenr:
Will do! Give me 10 minutes!

armenr:
You're right. 

On my phone, I run AdGuard. 

When enabled, I get the 404. 

When disabling adguard, I see the documentation. 

Adguard enabled:

![Screenshot_20210329-220101_Chrome.jpg](https://user-images.githubusercontent.com/7662191/112936291-61063880-90da-11eb-92dd-80c0114894e9.jpg)

Adguard disabled:
![Screenshot_20210329-220242_Chrome.jpg](https://user-images.githubusercontent.com/7662191/112936338-809d6100-90da-11eb-9174-30e2a6ea1855.jpg)

Sooooo, it sounds like your intuition is correct!

medelman17:
Thanks, @armenr! I'm going to close this issue, but link to it internally/in a new issue that tackles this issue globally.

tjleing:
Should be fixed now due to #3110 being merged in!  Please reopen if the problem persists 😄

Issue #3052, opened 2021-03-15
[missing-page] Ad Block Software Causes 404(ish)-Related Issue


harshvasoya008:
**Page**: [`https://docs.amplify.aws/lib/analytics/autotrack/q/platform/android`](https://docs.amplify.aws/lib/analytics/autotrack/q/platform/android)

**Feedback**: 
Page under the header "Automatically track sessions" gives 404 page not found error. Kindly look into this.

medelman17:
Hi @harshvasoya008, can you provide the URL returning a 404? I can't seem to reproduce the above.

harshvasoya008:
Hi @medelman17, URL is https://docs.amplify.aws/lib/analytics/autotrack/q/platform/android
Try opening this URL in new tab and you will get 404.

medelman17:
@harshvasoya008, I tried accessing the page as you suggested, and, unfortunately, I still can't recreate the issue--it seems to work fine on my end. Are you having trouble accessing any other content on the site? 

<img width="943" alt="Screen Shot 2021-03-16 at 11 05 13 AM" src="https://user-images.githubusercontent.com/14793389/111332433-2915e480-8648-11eb-91b6-b0f770a7dc94.png">

medelman17:
After much hair-pulling, we've determined that the above issue arises from use of ad-blocking browser extensions in chrome. Specifically, it appears that because the URL above contains the phrase 'autotrack', the underlying resources needed to fully render the page are blocked client-side. As a result, the website errors out and kicks over to its 404 page. Disabling ad-blocking extensions resolves the problem. 

I am going to close this issue and open another to discuss wider implications and potential global fixes.

tjleing:
Should be fixed now due to #3110 being merged in!  Please reopen if the problem persists 😄

Issue #3034, opened 2021-03-08
[missing-page] Analytics


bradlee96:
**Page**: [`https://docs.amplify.aws/lib/analytics/autotrack/q/platform/ios`](https://docs.amplify.aws/lib/analytics/autotrack/q/platform/ios)

**Feedback**: The page is missing!

palpatim:
@bradlee96 Can you clarify what page is missing? Do you mean that when you issue the `amplify console analytics` command, your browser displays a "missing page" error, or that the documentation site itself has missing content?

bradlee96:
Hi! I just meant that the documentation site itself has missing content. It does not load the page properly when attempting to access it from the rest of the documentation (for instance, from https://docs.amplify.aws/lib/analytics/record/q/platform/ios > "Automatically track sessions"). It doesn't affect me personally but I just noticed that it wasn't working so I figured I should report it.

palpatim:
@bradlee96 

I'm not seeing that same problem. Here are my steps:

- Navigate to https://docs.amplify.aws/lib/analytics/record/q/platform/ios
- Click **Automatically track sessions** in the left navigation

**Actual results**
- The page https://docs.amplify.aws/lib/analytics/autotrack/q/platform/ios loads, and displays the following content:
    ![image](https://user-images.githubusercontent.com/227982/110958457-900e6300-8301-11eb-97da-5c37fb32192c.png)

Can you confirm that you are seeing something different following those steps, or can you provide different steps so we can understand the issue?

bradlee96:
Here is a screenshot of what I see!

![image](https://user-images.githubusercontent.com/16856014/110966856-58acb000-831b-11eb-973b-259b5d3023b8.png)

lawmicha:
are there other pages missing for you? Do you see any errors in the browser console? have you tried private browse session?

bradlee96:
No, not as far as I was aware looking through the docs. I have tried private browse, which still does not work.

I get the following errors when:

1) Opening the page from the url directly:
![image](https://user-images.githubusercontent.com/16856014/110976966-2b65ff00-8327-11eb-952e-c22517183965.png)

2) Opening the page from another page:
p-9111c406.js:1725 POST https://pinpoint.us-west-2.amazonaws.com/v1/apps/53496f1ec97f4f57b6f9b1b5df579cda/events 400
FetchHttpHandler.handle @ p-9111c406.js:1725
(anonymous) @ p-6cf98df3.js:1882
[...]

PR #3110, opened 2021-04-03
Fix pages 404ing when adblocker is on

### LINKED ISSUE #1944: [missing-page]
**Page**: [`https://docs.amplify.aws/lib/push-notifications/overview/q/platform/js`](https://docs.amplify.aws/lib/push-notifications/overview/q/platform/js)

**Feedback**: "Push Notification" section of the amplify documentation is not available.

tjleing:
Hash filenames that are mistakenly triggering adblockers.  Should fix #1944, #3034, #3052 and #3082.

TODO:
- [ ] add .json ending back
- [ ] seems to only work on the pre-render build of the pages for now
- [ ] tests?

### DIFF
--- a/capi/src/write/get-page.ts
+++ b/capi/src/write/get-page.ts
@@ -9,6 +9,22 @@ const project = new Project({
   },
 });
 
+// DELETE
+function base64(path: string): string {
+  const paths = path.split("/");
+  let fileName = paths.pop();
+  if (!fileName) {
+    fileName = "";
+  }
+  let hashedFileName;
+  if (typeof btoa === "undefined") {
+    hashedFileName = Buffer.from(fileName).toString("base64");
+  } else {
+    hashedFileName = btoa(path);
+  }
+  return [...paths, hashedFileName].join("/");
+}
+
 export async function getPage(config: t.Config, ctx: t.Ctx): Promise<void> {
   await fs.ensureDir(config.outDir);
   return project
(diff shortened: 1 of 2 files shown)


capi/src/write/get-page.ts
```diff
@@ -9,6 +9,22 @@ const project = new Project({
   },
 });
 
+// DELETE
```
tjleing, line 12: Delete this after I figure out why this isn't working for the hydrated pages.

capi/src/write/pages.ts
```diff
@@ -135,9 +142,20 @@ export async function pages(ctx: t.Ctx): Promise<void> {
...
 
       await fs.ensureDir(outDir);
-      await fs.writeFile(pathDeduction.destinationPath, JSON.stringify(page), {
-        encoding: "utf8",
-      });
+      // hash file name to deal with aggressive adblockers
+      const paths = pathDeduction.destinationPath.split("/");
+      let fileName = paths.pop();
+      if (!fileName) {
+        fileName = "";
```
ericclemmons, line 149: In which cases will `fileName` be false-y?

Both `""` and `"/"` split & pop off `""`.

tjleing: This was for Typescript -- `Array.pop()` returns `string | undefined` where `undefined` is if the array is empty.  But it looks like you can just say `const fileName = paths.pop() as any` and it should compile, so thank you!

ericclemmons: Oooh, I was wondering why. This is was for type-safety!  Yes, I can see how `Array.pop()` could returned `undefined`, so no worries.

Since it's a false-y check, `|| ""` could also work.  Sorry for distracting!

tjleing:
Sorry for the bad timing @medelman17 -- as of _right now_ it's not a WIP 😅

medelman17:
Nice! I'm happy to be wrong, here :-)

medelman17:
@tjleing -- have we deployed to a staging environment to ensure that this works as expected? Just want to make sure we've functionally tested since it's sort of a big change.

medelman17 (approved):

PR #3128, opened 2021-04-10
Add note for using structured types with functions

camerondurham:
*Description of changes:*

Added a note to relevant Lambda documentation and pointed out requirements for using typed event input. 

As a first-time Lambda, I was unfamiliar with the expected input format for these functions. This note adds a link to the relevant Lambda documentation. While the examples in this document are helpful for the general expected design, it is difficult to understand the important parts required to implement a correct lambda resolver for a GraphQL function. I added the note to help other developers using more strongly typed languages such as Golang, Java, and C# should they encounter this error.

Thank you for considering this change!


akshbhu (approved):
Thanks for the Contribution
LGTM!!

camerondurham:
@akshbhu Thanks for reviewing! From [these docs](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks ), you may or may not have to manually kick off the CI to run, since it looks like a check isn't run yet? 

Sorry if this doesn't apply to how you do CI on this repo.

PR #3135, opened 2021-04-13
docs: Add page to troubleshoot multiple versions of a package

manueliglesias:
*Issue #, if available:*
N/A

*Description of changes:*
Added a page with upgrade instructions and troubleshooting

### DIFF
--- a/docs/lib/troubleshooting/fragments/js/upgrading.md
+++ b/docs/lib/troubleshooting/fragments/js/upgrading.md
@@ -57,7 +57,7 @@ npm ls -all 2>$null |
 
 ## Upgrade Amplify packages
 
-If you want to know update all Amplify packages to the `latest` version from npm and will make sure that only one version of every Amplify package is installed.
+If you want to update all Amplify packages to the `latest` versions from npm, run the following command, it will also make sure that only one version of every Amplify package is installed.
 
 <amplify-block-switcher>
 
@@ -93,7 +93,7 @@ npx npm-check-updates -i '/@?aws-amplify/' && npm update
 
 ## View latest Amplify package versions
 
-Sometimes it is useful to know which set of Amplify package versions belong to a release.
+If you want to see which set of Amplify package versions were published to npm as part of the latest release, you can run the following command:
 
 <amplify-block-switcher>
 
@@ -102,7 +102,7 @@ Sometimes it is useful to know which set of Amplify package versions belong to a
 ```sh
 # Show latest versions of amplify packages
 curl -s "https://api.github.com/repos/aws-amplify/amplify-js/commits/$(\
-    npm info aws-amplify .gitHead\
+    npm info aws-amplify@latest .gitHead\
   )" | jq -r '.commit.message'
 ```
 
@@ -114,7 +114,7 @@ curl -s "https://api.github.com/repos/aws-amplify/amplify-js/commits/$(\
 # Show latest versions of amplify packages
 (ConvertFrom-Json (
   iwr "https://api.github.com/repos/aws-amplify/amplify-js/commits/$(
-    npm info aws-amplify .gitHead
+    npm info aws-amplify@latest .gitHead
   )"
 ).Content).commit.message
 ```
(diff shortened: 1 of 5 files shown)


docs/lib/troubleshooting/fragments/js/upgrading.md
```diff
@@ -0,0 +1,154 @@
+When upgrading Amplify packages, it is important to make sure that there are no duplicate versions of Amplify packages in the `node_modules` folder tree as a result of the upgrade. Having multiple versions of packages can yield unexpected behavior as modules imported in your code might point to versions not configured by Amplify when calling `Amplify.configure`. To prevent this situation, you can [Check for duplicate versions](#check-for-duplicate-versions), and if duplicate versions exists, then [Upgrade Amplify packages](#upgrade-amplify-packages).
```
medelman17, line 1: Can you include the likely error-scenarios that would lead one to this documentation? For example, "If you're seeing... X, Y, Z, then you may want to check the following..."?

docs/lib/troubleshooting/fragments/js/upgrading.md
```diff
@@ -0,0 +1,154 @@
...
+
+</amplify-block>
+
+</amplify-block-switcher>
+
+---
+
+## Upgrade Amplify packages
+
+If you want to know update all Amplify packages to the `latest` version from npm and will make sure that only one version of every Amplify package is installed.
```
medelman17, line 60: I think there may be a typo in the above sentence--not sure I grasp the meaning as written.

docs/lib/troubleshooting/fragments/js/upgrading.md
```diff
@@ -0,0 +1,154 @@
...
+ - @aws-amplify/datastore@x.x.x
+ - @aws-amplify/interactions@x.x.x
+ - @aws-amplify/predictions@x.x.x
+ - @aws-amplify/pubsub@x.x.x
+ - @aws-amplify/pushnotification@x.x.x
+ - @aws-amplify/storage@x.x.x
+ - @aws-amplify/xr@x.x.x
+```
+
+All these versions were published as part of the same release, this means that if the versions you depend on are part of the same release, you won't have duplicated packages in your `node_modules`.
```
medelman17, line 154: Can we make it more clear what to do in the non-happy-path situation?

medelman17 (commented):
Great start! Thanks for this!

docs/lib/troubleshooting/fragments/js/upgrading.md
manueliglesias: Thanks Mike, the command outputs the latest versions of amplify packages **published to npm**, it doesn't look at your local dependencies. It is just a convenient way to see the `@latest` versions without having to go to the repo's homepage, look at the commits, search for the one for the release and look into the commit message.

The command is useful mostly for maintainers when helping customers. Maybe I should just remove it?

docs/lib/troubleshooting/fragments/js/upgrading.md
manueliglesias: Oops, thanks, fixed

docs/lib/troubleshooting/fragments/js/upgrading.md
manueliglesias: Once aws-amplify/amplify-js#8083 is merged, customers most likely will arrive here coming from the [bug issue template checklist](https://github.com/aws-amplify/amplify-js/pull/8083/files#diff-7e6dd26cf81256b658eb44be8a5664393c3d886d8de97606d7f89f691d98b48d) or the [messages logged by the library](https://github.com/aws-amplify/amplify-js/pull/8083/files#diff-db69e785d7fc5c73340d7a6b0eb17b7804601ee6143bb4d95df2968ace533c80)

Would you like to see the scenarios being mentioned here too?

medelman17: I would. We can't anticipate how customers will arrive at a given documentation resource. Theoretically, folks could arrive via search. Also, we're including the resource in our top-level navigation. In other words, "Every page is page one." We should provide more context, here.

docs/lib/troubleshooting/fragments/js/upgrading.md
medelman17: If it's mostly useful for maintainers, I would remove it. I empathize with the fact we lack a robust, internal help system / knowledge base.

medelman17 (approved):

PR #3137, opened 2021-04-13
fix(flutter): update flutter code snippets in datastore getting started

Jordan-Nelson:
*Issue #, if available:* N/A

*Description of changes:*
- update `codegen/ModelProvider.dart` to `models/ModelProvider.dart` in the initial code snippet for flutter datastore getting started (it was correct in the second snippet)
- fix inconsistencies between the plugin configuration & comment between the two snippets

### DIFF
--- a/docs/lib/datastore/fragments/flutter/getting-started/50_initDataStore.md
+++ b/docs/lib/datastore/fragments/flutter/getting-started/50_initDataStore.md
@@ -3,7 +3,7 @@ To initialize the Amplify DataStore, use the `Amplify.addPlugin()` method to add
 ```dart
 import 'models/ModelProvider.dart';
 
-// Add the following line to your app initialization to add the DataStore plugin
+// Add the following lines to your app initialization to add the DataStore plugin
 AmplifyDataStore datastorePlugin =
     AmplifyDataStore(modelProvider: ModelProvider.instance);
 Amplify.addPlugin(datastorePlugin);
@@ -30,7 +30,7 @@ class MyAmplifyApp extends StatefulWidget {
     }
 
     void _configureAmplify() async {
-        // Add the following line to your app initialization to add the DataStore plugin
+        // Add the following lines to your app initialization to add the DataStore plugin
         AmplifyDataStore datastorePlugin =
             AmplifyDataStore(modelProvider: ModelProvider.instance);
         Amplify.addPlugin(datastorePlugin);
--- a/docs/lib/datastore/fragments/flutter/getting-started/50_initDataStore.md
+++ b/docs/lib/datastore/fragments/flutter/getting-started/50_initDataStore.md
@@ -1,9 +1,9 @@
 To initialize the Amplify DataStore, use the `Amplify.addPlugin()` method to add the AWS DataStore Plugin. We also need to import the codegen dart file `ModelProvider.dart`
 
 ```dart
-import 'codegen/ModelProvider.dart';
+import 'models/ModelProvider.dart';
 
-// Add this in your app initialization
+// Add the following line to your app initialization to add the DataStore plugin
 AmplifyDataStore datastorePlugin =
     AmplifyDataStore(modelProvider: ModelProvider.instance);
 Amplify.addPlugin(datastorePlugin);
@@ -30,9 +30,10 @@ class MyAmplifyApp extends StatefulWidget {
     }
 
     void _configureAmplify() async {
-        // Add the following line to add DataStore plugin to your app
-        Amplify.addPlugin(AmplifyDataStore(modelProvider: ModelProvider.instance));
-
+        // Add the following line to your app initialization to add the DataStore plugin
+        AmplifyDataStore datastorePlugin =
+            AmplifyDataStore(modelProvider: ModelProvider.instance);
+        Amplify.addPlugin(datastorePlugin);
         try {
             await Amplify.configure(amplifyconfig);
         } on AmplifyAlreadyConfiguredException {


docs/lib/datastore/fragments/flutter/getting-started/50_initDataStore.md
```diff
@@ -1,12 +1,10 @@
...
 
 ```dart
-import 'codegen/ModelProvider.dart';
+import 'models/ModelProvider.dart';
 
 // Add this in your app initialization
-AmplifyDataStore datastorePlugin =
-    AmplifyDataStore(modelProvider: ModelProvider.instance);
-Amplify.addPlugin(datastorePlugin);
+Amplify.addPlugin(AmplifyDataStore(modelProvider: ModelProvider.instance));
```
cshfang, line 7: Actually think that the `datastorePlugin` declaration is more self-expressive even if it is slightly more verbose.

docs/lib/datastore/fragments/flutter/getting-started/50_initDataStore.md
```diff
@@ -16,28 +14,38 @@ Your code should look like this:
...
+  void _configureAmplify() async {
+    // Add the following line to add DataStore plugin to your app
+    Amplify.addPlugin(AmplifyDataStore(modelProvider: ModelProvider.instance));
+
+    try {
+      await Amplify.configure(amplifyconfig);
+    } on AmplifyAlreadyConfiguredException {
+      print(
+          "Tried to reconfigure Amplify; this can occur when your app restarts on Android.");
     }
```
Amplifiyer, line 43: Didn't we move away from this to `isConfigured`? cc: @haverchuck

docs/lib/datastore/fragments/flutter/getting-started/50_initDataStore.md
```diff
@@ -16,28 +14,38 @@ Your code should look like this:
...
+class _MyAmplifyAppState extends State<MyAmplifyApp> {
+  @override
+  void initState() {
+    super.initState();
+    _configureAmplify();
+  }
+
+  void _configureAmplify() async {
+    // Add the following line to add DataStore plugin to your app
+    Amplify.addPlugin(AmplifyDataStore(modelProvider: ModelProvider.instance));
```
cshfang, line 36: Hm, maybe you were just trying to be consistent - though maybe the more verbose code would be better for documentation.

I am a little confused why the plugin has to be added twice though, any ideas?

docs/lib/datastore/fragments/flutter/getting-started/50_initDataStore.md
Jordan-Nelson: Do you mean setting an `isConfigured` value and then removing the try catch?

This is how most of the examples are in the docs, but if there is another approach that is preferable I am happy to update them all. It actually would probably be good to decide on a consistent example of how to configure amplify. In searching around the docs for other examples I am realizing this is not the only one that doesn't correctly create a StatefulWidget.

The one example I found that has a boolean value for configured is [this](https://github.com/aws-amplify/docs/blob/master/docs/start/getting-started/fragments/flutter/integrate.md), but it still has the try catch.

docs/lib/datastore/fragments/flutter/getting-started/50_initDataStore.md
Jordan-Nelson: Yeah, I was just trying to be consistent. I don't have a strong preference between the two as long as they are consistent.

It doesn't have to be added twice. The first snippet is just the minimal code that someone would add if they were integrating Datastore into an app that already uses amplify, and the second is the full example of what your app would look like if you were integrating it into a new app.

docs/lib/datastore/fragments/flutter/getting-started/50_initDataStore.md
Jordan-Nelson: @Amplifiyer I tested this out with the hot restart changes. `Amplify.isConfigured` will be false after a hot restart, so the customer will still need to handle the exception for hot restarts.

I think `Amplify.isConfigured` being false after a hot restart is expected since that state will be lost. It *might* be ideal to have it maintain its state, but I don't think that was intended to be part of the hot restart fix. @haverchuck correct me if I am mistaken.

I think we should leave the try/catch, but update the print statement since I think most users will see this on hot restart now that hot restart works. I don't know the history of the current message ("this can occur when your app restarts on Android"). I assume that means re-launching after being moved to the background, but I am not sure.

My suggestion would be to update it to the following, but am open to suggestions. I took the Android portion out. If it is still relevant I can add it back, but I think we should clarify what "restarts on Android" means. As someone that is new to amplify, that wasn't clear to me. Once we decide, I can update all the docs to be consistent.

```dart
try {
  await Amplify.configure(amplifyconfig);
} on AmplifyAlreadyConfiguredException {
  print("Amplify was already configured. This can occur after a hot reload.");
}
```

haverchuck: We continue to throw (and not swallow) the AmplifyAlreadyConfiguredException because attempting to call configure multiple times is something customers might actually try (I think this is supported in JS?)... but now that I am revisiting the code, I think we could swallow the exception on the method channel response since redundant calls would only make it across the method channel on a hot restart.  But in the absense of making that change (which would admittedly be small) then the change proposed above makes sense.

Jordan-Nelson: I decided to make this PR more targeted to the original issue I was trying to fix (incorrect import & inconsistent configuration) since we should probably update all the docs after [this PR](https://github.com/aws-amplify/amplify-flutter/pull/527) lands.

cshfang (approved):

Ashish-Nanda (approved):
LGTM

PR #3162, opened 2021-04-26
Update AmplifySignOut buttonText prop to correct case

imjacobclark:
*Issue #, if available:* N/A

<AmplifySignOut> on React has a prop of `buttonText`, the docs show this as `button-text` which is incorrect. 

*Description of changes:*

Docs updated to reflect correct API of `<AmplifySignOut buttonText="Log out"></AmplifySignOut>` rather than `<AmplifySignOut button-text="Log out"></AmplifySignOut>`.

### DIFF
--- a/docs/ui/auth/fragments/react/sign-out.md
+++ b/docs/ui/auth/fragments/react/sign-out.md
@@ -1,4 +1,4 @@
-<amplify-sign-out button-text="Custom Text"></amplify-sign-out>
+<amplify-sign-out buttonText="Custom Text"></amplify-sign-out>
 
 **Usage**
 
--- a/docs/ui/auth/fragments/react/sign-out.md
+++ b/docs/ui/auth/fragments/react/sign-out.md
@@ -1,9 +1,9 @@
-<amplify-sign-out buttonText="Custom Text"></amplify-sign-out>
+<AmplifySignOut buttonText="Custom Text"></AmplifySignOut>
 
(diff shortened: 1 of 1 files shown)


docs/ui/auth/fragments/react/sign-out.md
```diff
@@ -1,4 +1,4 @@
-<amplify-sign-out button-text="Custom Text"></amplify-sign-out>
+<amplify-sign-out buttonText="Custom Text"></amplify-sign-out>
```
tjleing, line 1: It's also `AmplifySignOut` instead of `amplify-sign-out` too, right?

```suggestion
<AmplifySignOut buttonText="Custom Text"></AmplifySignOut>
```

docs/ui/auth/fragments/react/sign-out.md
```diff
@@ -1,4 +1,4 @@
...
 
 **Usage**
```
tjleing, line 4: Could you also make the same change down on line 6?  GitHub is preventing me from commenting directly on that line for some reason...

tjleing (changes requested):
Thank you for the feedback!  Just have a few nits before merging -- otherwise, looks good!

imjacobclark:
@tjleing Don!

tjleing (approved):
Thank you -- LGTM! 🚀

PR #3181, opened 2021-05-04
update: revamped flutter getting started tutorial

cshfang:
Revamp Amplify Flutter getting started tutorial to integrate with DataStore instead of Analytics category.


cshfang:
Github repo for final code is still not created. This PR will not be merged until that repo is created, readied and links to it are updated.

Jordan-Nelson (commented):
Looks pretty good! Most of my feedback was small typos or nits.

cshfang:
> Looks pretty good! Most of my feedback was small typos or nits.

The perfect type of feedback for docs imo

docs/start/getting-started/fragments/flutter/generate-model.md
```diff
@@ -0,0 +1,27 @@
+With the basic setup complete, you can now model the data your application will work with. Amplify DataStore persists the modeled data on your local device and can even synchronize to a backend API without writing any additional code. These models are specified as [GraphQL](http://graphql.org/) schemas. You can [learn more](~/cli/graphql-transformer/overview.md)) about GraphQL schemas and data modeling. For now, let’s start by generating some initial data models.
```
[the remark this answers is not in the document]
cshfang: 🕵️

docs/start/getting-started/fragments/flutter/generate-model.md
```diff
@@ -0,0 +1,27 @@
...
+
+<amplify-callout>
+
+**Did you know?** You can get started with Amplify’s [admin UI](https://sandbox.amplifyapp.com/) without even creating an AWS Account.
```
drochetti, line 5: This is an interesting selling point of the Admin UI and could potentially remove a lot of friction from the tutorial for users just looking for a quick try out. However, in the first page "Create an AWS account" is defined as a pre-requisite.

@renebrandel thoughts on this?

It sounds like we should rethink what the pre requisites for the Tutorial actually are. Maybe defer the AWS account creation until it's time to Sync data to the cloud?

docs/start/getting-started/fragments/flutter/generate-model.md
```diff
@@ -0,0 +1,27 @@
...
+
+</amplify-block>
+
+<amplify-block name="CLI">
+
+<inline-fragment src="~/start/getting-started/fragments/flutter/blocks/generate-model-cli.md"></inline-fragment>
+
+</amplify-block>
+
+</amplify-block-switcher>
```
drochetti, line 25: I like this pattern for the Amplify documentation in general, but for the tutorial I believe we should be more opinionated and offer less options (ideally offer a single option). For the "Getting Started" of each category under the "Libraries" menu I think this pattern is the way to go, for the tutorial I'd say going with Admin UI only is perfectly reasonable.

pinging @renebrandel again, thoughts? I'm hesitant to make customers face a decision that they might not be able to this early in the process.

docs/start/getting-started/fragments/flutter/add-api.md
```diff
@@ -1,65 +1,143 @@
...
+Update the `main.dart` file and import the new packages.
+
+```diff
+  import 'package:amplify_flutter/amplify.dart';
+  import 'package:amplify_datastore/amplify_datastore.dart';
++ import 'package:amplify_api/amplify_api.dart';
++ import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
+```
+
+Declare the additional plugins at the top of the `_TodosPageState` class.
```
drochetti, line 105: For these type of direct / single sentence instructions, another sentence just to get the customer aware of _why_ this is needed could be a nice addition. Something like:

> Declare the additional plugins at the top of the `_TodosPageState` class. This will ensure DataStore has access to the AppSync API.

docs/start/getting-started/fragments/flutter/add-api.md
```diff
@@ -1,65 +1,143 @@
-### Setup AWS Cloud Resources with Amplify CLI 
+## Deploy your Amplify sandbox backend
 
-We will now use the Amplify CLI to configure the AWS Cloud Resources that will power your app. 
+<amplify-callout>
 
-1. First install the amplify cli. Within your terminal run: 
+If you chose to generate your models earlier via the CLI, you can skip this step and proceed to [Add authentication](~/start/getting-started/add-api.md#add-authentication).
```
drochetti, line 5: Following my comment about the multiple options, keeping it opinionated towards Admin UI would avoid another branch out at this point. @renebrandel

docs/start/getting-started/fragments/flutter/add-api.md
```diff
@@ -1,65 +1,143 @@
-### Setup AWS Cloud Resources with Amplify CLI 
+## Deploy your Amplify sandbox backend
```
drochetti, line 1: I think this is the moment we could introduce the "create AWS account". My hypothesis is that at this point the customer has already seen stuff working, they're likely enjoying the smooth experience so far, with very little friction and are more willing to spend time creating an account and provisioning cloud resources, both time consuming tasks. @renebrandel

docs/start/getting-started/fragments/flutter/nextsteps.md
```diff
@@ -1,12 +1,12 @@
...
-- [REST API](~/lib/restapi/getting-started.md)
 - [GraphQL API](~/lib/graphqlapi/getting-started.md)
+- [REST API](~/lib/restapi/getting-started.md)
+- [Storage](~/lib/storage/getting-started.md)
 
-### Reference Sample App 
+### References
 
-For an end to end example application, check out the example available [in the GitHub repository](https://github.com/aws-amplify/amplify-flutter/tree/master/example).
+Find the source code of the final Todo App built in this tutorial [on our GitHub repository]().
```
drochetti, line 12: awesomeness! 🚀

drochetti (commented):
This is awesome! Such a huge improvement to the existing tutorials imho.

I left a few comments about the flow and options we're offering, but overall this seems great and I believe we should replicate this to all the other platforms. The link to a repo with the actual app working is 🔥.

@renebrandel @hdworld11 @ukhan-amazon @cgsea @sammartinez Can we coordinate a project to update the Tutorial of all platforms to follow this one? A lot of the content can even be shared, only a few code snippets and platform-specific callouts will need work, @cshfang did all the heavy lifting already.

docs/start/getting-started/fragments/flutter/generate-model.md
cshfang: I think it's a fair point to be opinionated, especially early on, and limit customer options to Admin UI. I was originally hesitant about making this change as it would be a large departure from our other platforms and so I opted for a more hybrid approach. But this, among other feedback, seems to indicate that maybe it's ok to just go ahead and offer what we feel is the recommended starting path. Our current block switching implementation also does not play as well with this pattern as it breaks the outline so that's another con for this hybrid approach.

docs/start/getting-started/fragments/flutter/add-api.md
cshfang: I tried to look these over and I think we can elaborate on the "why" at the top of this section without actually adding too much per small instruction simply because they're all just doing the same thing for the same reason. I.e. import, declare and add plugins in order to ____

I also don't want to introduce the concept of AppSync at all since a customer can realistically get through this tutorial without ever understanding that AppSync is behind some of the magic.

I will update this section to elaborate a bit on why we're doing this though!

docs/start/getting-started/fragments/flutter/add-api.md
```diff
@@ -32,49 +26,48 @@ On the creation form:
...
     ![Open admin UI](~/images/lib/getting-started/flutter/connect-to-cloud-open-admin-ui.png)
 
-### Update local project with deployed environment
+## Add authentication
 
-From your newly minted admin UI:
-1. Click **Local setup instructions** at the top of the admin UI
-2. Copy the command for pulling the updated environment and run it in your terminal.
-3. Your web browser should prompt you with the question **Are you sure you want to login to the Amplify CLI?**, and you can select **Yes** to move on.
+Since our **Todo** model specifies an @auth directive, we do need to first add authentication.
```
drochetti, line 31: nit: put @auth in backticks for inline code

docs/start/getting-started/fragments/flutter/generate-model.md
drochetti: nit: Admin UI instead of admin UI?

drochetti (approved):
This is so great! Thank you so much for being so thoughtful about every step of this tutorial.

I left a couple of nit-picking comments, but lgtm! 🚀 🚢

docs/start/getting-started/fragments/flutter/integrate.md
```diff
@@ -1,4 +1,4 @@
-In this section you’ll integrate Amplify DataStore with your app, and learn to use the generated data model to create, update, query, and delete Todo items by building an app. You can find the source code of the final Todo App [on our GitHub repository]().
+In this section you’ll integrate Amplify DataStore with your app, and learn to use the generated data model to create, update, query, and delete Todo items by building an app. You can find the source code of the final Todo App [on our GitHub repository](https://github.com/cshfang/amplify-flutter-datastore-starter/tree/main/amplified_todo).
```
cshfang, line 1: We will update these links when we have a more well-defined (and official) examples repo set up for amplify-flutter.

docs/start/getting-started/fragments/flutter/integrate.md
```diff
@@ -1,120 +1,470 @@
...
+  // Amplify.DataStore.observe() and when an update event occurs, fetch the
+  // todo list
+  // 
+  // note this strategy may not scale well with larger number of entries 
+  _subscription = Amplify.DataStore.observe(Todo.classType).listen((event) {
+    _fetchTodos();
+  });
+
+  // fetch Todo entries from DataStore
+  await _fetchTodos();
```
Amplifiyer, line 413: I think we should avoid this call pattern in our getting started guide. This is not really the right intended usage of `observe()` API and if customers are not careful and leave this in, it will be pretty bad for performance. Also what's the reason for calling `_fetchTodos()` twice here?

Amplifiyer (commented):
this is amazing improvement! Left a couple of suggestions. We should also remember to update the github code repository links.

docs/start/getting-started/fragments/flutter/integrate.md
cshfang: I tried to call out that this isn't optimal for larger datasets but, for the purposes of getting a simple app up and running, I think it's the simplest way to get the app displaying real time updates without getting too deep into the weeds about how to set up their app to actually respond to the granular events. I agree it would be bad for customers to leave things as is but I think the purpose of the guide isn't necessarily to help them with more advanced scenarios or UX patterns but to convey the value of Amplify quickly. I believe *everything* in this tutorial could/should be shored up as customers expand upon their use cases.

> Also what's the reason for calling _fetchTodos() twice here?

The `_fetchTodos()` on line 409 only fires in response to future streamed events while the one on on line 413 is part of the cold start.

docs/start/getting-started/fragments/flutter/integrate.md
```diff
@@ -1,120 +1,470 @@
...
 import 'package:flutter/material.dart';
+// amplify packages we will need to use
 import 'package:amplify_flutter/amplify.dart';
-import 'package:amplify_analytics_pinpoint/amplify_analytics_pinpoint.dart';
-import 'package:amplify_auth_cognito/amplify_auth_cognito.dart';
+import 'package:amplify_datastore/amplify_datastore.dart';
+// amplify configuration and models that should have been generated for you
 import 'amplifyconfiguration.dart';
+import 'models/ModelProvider.dart';
+import 'models/Todo.dart';
```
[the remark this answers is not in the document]
cshfang: Ah I wasn't aware we had collapsible segments. It would be nice if we could truncate some of these large code blocks where the display allows expansion but the copy still copies the whole snippet... do you know if that is possible today with the collapsible segments you alluded to?

Fwiw, I do think that the outline navigation allows for some skipping around as is but I can definitely see the initial boilerplate being a big of a deterrent (though I really did try to go through and trim the fat!)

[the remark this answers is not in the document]
Amplifiyer: to be honest it's more of a wishlist at this point. CC: @tjleing @wlee221 to see what's possible here.

Would breaking this page down in multiple pages help?
* Create a boilerplate flutter app
* Add Local DataStore
* Add online sync

docs/start/getting-started/fragments/flutter/integrate.md
Amplifiyer: >I agree it would be bad for customers to leave things as is but I think the purpose of the guide isn't necessarily to help them with more advanced scenarios or UX patterns but to convey the value of Amplify quickly.

We should assume the worst in my opinion though it's not a security risk but a performance one, but still significant. Observe is not necessarily an advanced case and if it is we should just leave it out. For local updates, we can update the UI for every `save` and call it quits. If we do want to show cross device sync, we should demonstrate `observe` accurately.

cshfang: Hmm given that I think the value of cross device sync is quite large with DataStore, it's probably worth leaving in, imo. I'm not sure that just updating UI for every `save` is enough to show that things are actually being persisted since we'd essentially just be optimistically updating the UI that way. Could you please share what you mean with demonstrating usage of `observe` accurately?

PR #3186, opened 2021-05-04
Wrap async build function in try/catch

jakeburden:
*Issue #, if available:*

*Description of changes:*
This change ensures that the build process exits upon error.
This is accomplished by wrapping a promise used in an async function that throws an error in a try/catch statement.


tjleing (approved):
LGTM!

PR #3208, opened 2021-05-07
fix: datastore events model name

### LINKED ISSUE #2837: Incorrectly documented method to retrieve model's name for DataStore events (Amplify JS)

amhinson:
*Issue #, if available:*
fixes #2837

*Description of changes:*
- Change JS DataStore events to be `model.name` instead of `modelName`
- Add platform specific fragments so that this change only applies to JS, since the other platforms still have `modelName`
- Update fragment naming convention & folder structure to be more specific

image

image

---


sammartinez (approved):
LGTM 🌮

PR #3222, opened 2021-05-12
Modified imports for IE 11 Support

davida26:
Fix IE 11 Imports

*Issue #, if available:*

*Description of changes:*
As of core-js 3+ entry points do not need es6 or es7 specified. They should be replaced with just "es".

See CommonJS entry points section
https://github.com/zloirock/core-js/blob/master/README.md#ecmascript-typed-arrays


ericclemmons (approved):
Thanks for this update!

(The assumption is that customers install `core-js@latest`, which is `3.12.1` at the time of writing.)

PR #3229, opened 2021-05-13
Fix issue with incorrect attribute name

bricef:
**Description of changes:**
The documentation is incorrect. It took me ages to find the right invocation to make the `formFields` input work correctly.

I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


ErikCH (changes requested):
@bricef Thanks so much for your request!

I have tested both with `form-fields.prop` and `formFields.prop` and both work as expected. You should be able to use both. I'm wondering if you have an example app that it's not working with `form-fields.prop` in?. Were you using Vue2 or Vue3? Or Vue CLI or Vite? Please let me know.

Issue #3151, opened 2021-04-16
[Feedback] documentation for recording custom event attributes is inconsistent/wrong


gspeicher:
**Page**: [`/lib/analytics/record/q/platform/js#record-a-custom-event-with-attributes`](https://docs.amplify.aws/lib/analytics/record/q/platform/js#record-a-custom-event-with-attributes)

**Feedback**:

Under "Record a Custom Event with Attributes", the code example contains the comment "Attribute values must be strings" but then below the code block, the documentation states that "Attributes must have the type `String` or be an array of strings". Which is correct? Are array values allowed?

If I actually attempt to record an event with a custom attribute containing an array of strings, the operation never completes, never returns an error, and the event never gets recorded. It just hangs indefinitely. Not sure if that is the expected behavior but can't tell from the documentation whether or not this is a bug.

nickarocho:
Hi @gspeicher, just wanted to check in here and let you know I'm working on replicating the issue to further understand the underlying functionality.

Once I am getting custom event attributes to show up in my sample app I'll update you to see if we need to simply make a documentation change or if there is a problem in the underlying method's source code.

nickarocho:
Hi @gspeicher, I am working with an internal team to resolve this issue on a technical level. For now, we've removed the line in question from our documentation for now as a safeguard to avoid any further confusion.

Thanks for pointing this out to us!

PR #3231, opened 2021-05-14
chore(analytics-js): remove custom event attribute value type line

nickarocho:
*Issue #, if available:*
https://github.com/aws-amplify/docs/issues/3151

*Description of changes:*
Removes the following line from the Analytics (JS) section describing recording custom events:
`Attribute values must have the type String or be an array of strings.`

It appears Pinpoint API documentation does not specify exactly what types of values they accept within the `Attributes` object. When attempting to pass an `array` of `strings` in my sample app, the server responds with a `400` error.

Removing this line from the documentation will help avoid further confusion as seen from https://github.com/aws-amplify/docs/issues/3151.

Screen Shot 2021-05-14 at 12 14 07 PM

### DIFF
--- a/docs/lib/analytics/fragments/js/record.md
+++ b/docs/lib/analytics/fragments/js/record.md
@@ -18,8 +18,6 @@ Analytics.record({
 });
 ```
 
-Attribute values must have the type `String` or be an array of strings.
-
 ## Record Engagement Metrics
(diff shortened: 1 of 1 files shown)


sammartinez (approved):
LGTM 🌮

PR #3236, opened 2021-05-18
Add Guide for parallel processing in DataStore on iOS

Kilo-Loco:
*Issue #, if available:*
N/A

*Description of changes:*
Added a Guide that shows user how to perform batch CRUD operations using Combine for iOS. New section named "DataStore" was also added to the Guides.

### DIFF
--- a/docs/guides/datastore/fragments/parallel-processing-ios.md
+++ b/docs/guides/datastore/fragments/parallel-processing-ios.md
@@ -1,14 +1,14 @@
-Amplify DataStore provides APIs that allow you to create, read, update, and delete items individually. Sometimes you need to perform batch operations like creating and updating multiple objects or querying and deleting object by their IDs. In this guide, you will learn how to perform batch operations using Combine to extend the functionality of the Amplify DataStore APIs.
+Amplify DataStore provides APIs that allow you to create, read, update, and delete items individually. Sometimes you need to perform multiple operations in parallel, like creating and updating several objects or querying and deleting objects by their IDs. In this guide, you will learn how to perform parallel processing using the Combine framework to extend the functionality of the Amplify DataStore APIs.
 
 <amplify-callout warning>
 
-For simplicity, this Guide demonstrates how to perform batch operations using the Combine framework. To follow along, you must include `import Combine` at the top of your file.
+For simplicity, this Guide demonstrates how to perform parallel processing using the Combine framework. To follow along, you must include `import Combine` at the top of your file.
 
 </amplify-callout>
 
-## Batch create and update
+## Create and update in parallel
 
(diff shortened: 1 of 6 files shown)


undefobj:
One suggestion - maybe we use a different phrase other than "batch" for this guide? [Batch terminology is something very specific with DynamoDB](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html) and I fear customers will get confused.

drochetti:
> One suggestion - maybe we use a different phrase other than "batch" for this guide? [Batch terminology is something very specific with DynamoDB](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html) and I fear customers will get confused.

I second this. Even outside of the DynamoDB scenarios, batching has deeper implications, It think what it's happening in this PR is more about parallel processing than batching.

aspittel:
I like the "in parallel" phrasing change and think this looks valuable, I'd love to hear more from some iOS focused folks on the team too.

docs/guides/datastore/fragments/parallel-processing-ios.md
```diff
@@ -0,0 +1,81 @@
...
+To write data to the DataStore in parallel, pass an array of models to a Combine `Publishers.Sequence` and perform `Amplify.DataStore.save(_ model:)` on each object:
+
+```swift
+let todos = [
+    Todo(name: "Learn Amplify"),
+    Todo(name: "Build app"),
+    Todo(name: "Profit")
+]
+
+let sink = Publishers.Sequence<[Todo], Never>(sequence: todos)
```
palpatim, line 20: Here and below: can simplify this to `todos.publisher`

docs/guides/datastore/fragments/parallel-processing-ios.md
```diff
@@ -0,0 +1,81 @@
...
+    .sink {
+        if case let .failure(error) = $0 {
+            print("Error on save() for type Todo - \(error.localizedDescription)")
+        }
+    } receiveValue: { savedTodos in
+        print("Todos: \(savedTodos)")
+    }
+```
+
+`Publishers.Sequence` creates a `Publisher` where each model is passed to `.flatMap()` individually, allowing for each object to be saved to DataStore. `.collect()` emits a single array of all the elements collected, returning the saved models in `receiveValue`.
```
palpatim, line 34: "`todos.publisher` creates a ..."

palpatim (approved):
Very cool guide, and I also learned more about Combine. :) Minor notes below, but LGTM

aspittel (approved):

PR #3260, opened 2021-05-25
chore: Add user state information for AWSMobileClient

royjit:
*Description of changes:*

Screen Shot 2021-05-26 at 2 58 36 PM

### DIFF
--- a/docs/sdk/auth/fragments/ios/how-it-works.md
+++ b/docs/sdk/auth/fragments/ios/how-it-works.md
@@ -66,11 +66,11 @@ User was signed out from the app by calling the signedOut api and the app is not
 
 #### signedOutFederatedTokensInvalid
 
-Only `signedIn` state can transition to `signedOutFederatedTokensInvalid` if the following condition are true: i) user was in `signedIn` by federating to Cognito Identity Pool ii) AWS credential fetch failed because the social provider token passed during sign in got invalid.
+Only `signedIn` state can transition to `signedOutFederatedTokensInvalid` if the following condition are true: i) user was in `signedIn` by federating to Cognito Identity Pool ii) AWS credential fetch failed because the social provider token passed during sign in got invalid. You can now re-authenticate the user by calling `federatedSignIn` with a refreshed token. If you make a `getAWSCredentials` call it will wait till the user is signed in to return a result.
 
 #### signedOutUserPoolsTokenInvalid
 
-Only `signedIn` state can transition to `signedOutUserPoolsTokenInvalid` if the following condition are true: i) user was in `signedIn` by signing to Cognito User Pool ii) Cognito user pool token fetch failed because the JWT refresh token got invalid.
+Only `signedIn` state can transition to `signedOutUserPoolsTokenInvalid` if the following condition are true: i) user was in `signedIn` by signing to Cognito User Pool ii) Cognito user pool token fetch failed because the JWT refresh token got invalid. You can now re-authenticate the user by calling `signIn` or `showSignIn`. If you make a `getAWSCredentials` or `getToken` call it will wait till the user is signed in to return a result.
 
 #### unknown
 
--- a/docs/sdk/auth/fragments/ios/how-it-works.md
+++ b/docs/sdk/auth/fragments/ios/how-it-works.md
@@ -56,9 +56,9 @@ More detailed explanation of different states are given below:
 
 You can invoke different `AWSMobileClient` methods like `signIn`, `showSignIn`, `federatedSignIn`, etc.,  which will sign in a user as either a Cognito User Pool user, or federate them to a Cognito Identity Pool. If user has successfully signed in using any of these apis, `AWSMobileClient` will move to `signedIn` state. The available tokens/credentials depend on the signed in method and configuration. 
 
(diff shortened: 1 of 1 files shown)


docs/sdk/auth/fragments/ios/how-it-works.md
```diff
@@ -46,6 +46,33 @@ AWSMobileClient.default().addUserStateListener(self) { (userState, info) in
...
+
+#### guest
+
+`AWSMobileClient` will be in `guest` state if you are not signed in to the app and unauthenticated access is enabled. Only Cognito Identity Pool related credentials will be available in `guest` state ie, identity id and aws temporary credentials for the unauthenticated role.
+
+#### signedIn
+
+User can invoke  different apis like `signIn`, `showSignIn`, `federatedSignIn` etc.. using `AWSMobileClient` which will sign them as either a Cognito User pool user or federate them to Cognito Identity Pool. If user has successfully signed in using any of these apis, `AWSMobileClient` will move to `signedIn` state. The available tokens/credentials depend on the signed in method and configuration. 
+If the user is signed in to Cognito User pool ie using `signIn` api or `showSignIn` with hostedUI, `getToken` will be available with Cognito User Pool tokens. They will also have access to identity id and aws credentials for authenticated role if Cognito Identity Pool is configured with the Cognito User Pool. On the other hand if the user is signed in via federated signIn to Cognito IdentityPool, only identity id and aws credentials for authenticated role will be available.
```
palpatim, line 58:
```suggestion
You can invoke different `AWSMobileClient` methods like `signIn`, `showSignIn`, `federatedSignIn`, etc.,  which will sign in a user as either a Cognito User Pool user, or federate them to a Cognito Identity Pool. If user has successfully signed in using any of these apis, `AWSMobileClient` will move to `signedIn` state. The available tokens/credentials depend on the signed in method and configuration. 

If the user is signed in to Cognito User Pool (i.e., using `signIn` or `showSignIn` with hostedUI), `getToken` will return Cognito User Pool tokens. If your Cognito User Pool is configured to federate into a Cognito Identity Pool, you will also have access to `identityId`, and AWS credentials for the authenticated role. If no Identity Pool is configured, both `identityId` and `getCredentials` will <Return null? Crash? Return garbage?>?

On the other hand if the user is signed in via `federatedSignIn` to a Cognito IdentityPool, only `identityId` and AWS credentials for authenticated role will be available, and `getToken` will <Return null? Crash? Return garbage?>
```

docs/sdk/auth/fragments/ios/how-it-works.md
```diff
@@ -46,6 +46,33 @@ AWSMobileClient.default().addUserStateListener(self) { (userState, info) in
...
+
+#### signedOut
+
+User was signed out from the app by calling the signedOut api and the app is not configured to be in guest mode.
+
+#### signedOutFederatedTokensInvalid
+
+Only `signedIn` state can transition to `signedOutFederatedTokensInvalid` if the following condition are true: i) user was in `signedIn` by federating to Cognito Identity Pool ii) AWS credential fetch failed because the social provider token passed during sign in got invalid.
```
palpatim, line 66: What methods are valid to call? What is the behavior if invalid method is called?

docs/sdk/auth/fragments/ios/how-it-works.md
```diff
@@ -46,6 +46,33 @@ AWSMobileClient.default().addUserStateListener(self) { (userState, info) in
...
+
+#### signedOutUserPoolsTokenInvalid
+
+Only `signedIn` state can transition to `signedOutUserPoolsTokenInvalid` if the following condition are true: i) user was in `signedIn` by signing to Cognito User Pool ii) Cognito user pool token fetch failed because the JWT refresh token got invalid.
```
palpatim, line 70: What methods are valid to call? What is the behavior if invalid method is called?

docs/sdk/auth/fragments/ios/how-it-works.md
```diff
@@ -46,6 +46,33 @@ AWSMobileClient.default().addUserStateListener(self) { (userState, info) in
...
+
+#### unknown
+
+`unknown` should never occur, it is the initial state set when the app starts and after the `AWSMobileClient` initializes, it will transition to either `guest`, `signedOut` or `signedIn` state.
```
palpatim, line 74:
```suggestion
It is very unlikely you will see the `unknown` state, as it is the initial state set when the app starts. After the `AWSMobileClient` initializes, it will transition to either `guest`, `signedOut` or `signedIn` state.
```

medelman17: Omit 'very' (editorializing) and remove comment after 'state' before 'as' -- not acting as a coordinating conjunction; rather, 'as', here, is a subordinating conjunction (like 'because') -- thus, no comma is necessary.

docs/sdk/auth/fragments/ios/how-it-works.md
royjit: Ideally developer should call one of the signIn method again to re-authenticate the user. All other method calls which are valid for signedIn user are still valid. ie getIdentity should return back the cached identity, getAWSCredentials will put the request into the queue and wait for signIn .. etc

docs/sdk/auth/fragments/ios/how-it-works.md
royjit: Same as above, all calls are valid similar to the signed in user.

docs/sdk/auth/fragments/ios/how-it-works.md
royjit: Added more content

palpatim (approved):

PR #3290, opened 2021-06-04
fix(Storage): iOS storage operations cancel/pause/resume

lawmicha:
_Issue #, if available:_

_Description of changes:_
- add operation functionality such as cancel/pause/resume to the upload and download docs
- mention multipart upload after 5MB for upload API

### DIFF
--- a/docs/lib/storage/fragments/ios/download.md
+++ b/docs/lib/storage/fragments/ios/download.md
@@ -9,7 +9,7 @@ You can download to in-memory buffer [Data](https://developer.apple.com/document
 <amplify-block name="Listener (iOS 11+)">
 
 ```swift
-Amplify.Storage.downloadData(
+let storageOperation = Amplify.Storage.downloadData(
     key: "myKey", 
     progressListener: { progress in
         print("Progress: \(progress)")
@@ -57,7 +57,7 @@ let downloadToFileName = FileManager.default.urls(for: .documentDirectory,
                                                   in: .userDomainMask)[0]
     .appendingPathComponent("myFile.txt")
 
-Amplify.Storage.downloadFile(
+let storageOperation = Amplify.Storage.downloadFile(
     key: "myKey",
     local: downloadToFileName,
     progressListener: { progress in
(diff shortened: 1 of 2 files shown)


docs/lib/storage/fragments/ios/download.md
```diff
@@ -96,7 +96,26 @@ receiveValue: {
 
 </amplify-block-switcher>
 
-### Generate a download URL
+## Cancel, Pause, Resume
+
+After you call `downloadData` or `downloadFile` as above, your containing class retains a reference to the operation that is actually performing the download.
+
+To cancel the download (for example, in response to the user pressing a **Cancel** button), you simply call `cancel()` on the upload operation.
```
royjit, line 103:
```suggestion
To cancel the download (for example, in response to the user pressing a **Cancel** button), you simply call `cancel()` on the download operation.
```

docs/lib/storage/fragments/ios/download.md
```diff
@@ -96,7 +96,26 @@ receiveValue: {
...
+
+```swift
+func cancelDownload() {
+    storageOperation.cancel()
+}
+```
+
+You can also pause then resume the operation.
```
royjit, line 111:
```suggestion
You can also pause and then resume the operation.
```

docs/lib/storage/fragments/ios/download.md
```diff
@@ -96,7 +96,26 @@ receiveValue: {
 
 </amplify-block-switcher>
 
-### Generate a download URL
+## Cancel, Pause, Resume
+
+After you call `downloadData` or `downloadFile` as above, your containing class retains a reference to the operation that is actually performing the download.
```
ameter, line 101:
```suggestion
Calls to `downloadData` or `downloadFile` return a reference to the operation that is actually performing the download.
```

ameter: Take this suggestion with a grain of salt, but it's up to the calling class whether to retain the reference, so it may be simpler just to say that the reference is returned.

docs/lib/storage/fragments/ios/upload.md
```diff
@@ -109,3 +109,27 @@ receiveValue: { data in
 </amplify-block>
 
 </amplify-block-switcher>
+
+## Cancel, Pause, Resume
+
+After you call `uploadFile` as above, your containing class retains a reference to the operation that is actually performing the upload.
```
ameter, line 115:
```suggestion
Calls to `uploadData` or `uploadFile` return a reference to the operation that is actually performing the upload.
```

ameter: Same recommendation as download, except that I added uploadData for completeness and to be consistent with the download documentation.

ameter (commented):
I think the documentation is good and makes it clear.  I added a couple of minor suggestions.

ameter (approved):
Looks good.

PR #3291, opened 2021-06-04
fix(CLI): admin queries API with updated code snippets

lawmicha:
_Issue #, if available:_
https://github.com/aws-amplify/amplify-ios/issues/1103

_Description of changes:_
This PR updates the Admin Queries API example to include platform specific examples using Amplify. It mostly replicates the existing JS code snippet in the different platform languages

### DIFF
--- a/docs/cli/auth/admin.md
+++ b/docs/cli/auth/admin.md
@@ -113,7 +113,19 @@ export default withAuthenticator(App, true);
 
 <amplify-block name="iOS">
 
-1. Configure Amplify. See [Getting started with Amplify.API for REST](~/lib/restapi/getting-started.md) to learn more.
+1. Initialize Amplify API. Refer to [Getting Started with Amplify.API for REST](~/lib/restapi/getting-started.md) for more details. 
+
+You should have the initialization code including the imports:
+
+```swift
+import Amplify
+// If you are using Swift Package Manager
+import AWSAPIPlugin
+// If you are using Cocoapods
+import AmplifyPlugins
+```
+
+and code that adds `AWSCognitoAuthPlugin`, `AWSAPIPlugin`, and configures it.
 
 ```swift
 try Amplify.add(plugin: AWSCognitoAuthPlugin())
--- a/docs/cli/auth/admin.md
+++ b/docs/cli/auth/admin.md
@@ -121,7 +121,7 @@ You should have the initialization code including the imports:
 import Amplify
 // If you are using Swift Package Manager
 import AWSAPIPlugin
-// If you are using Cocoapods
+// or if you are using Cocoapods
 import AmplifyPlugins
 ```
 
@@ -221,6 +221,13 @@ try Amplify.API.add(interceptor: MyCustomInterceptor(), for: "[YOUR-RESTENDPOINT
 Set up the custom interceptor to return the ID token for the request.
 
 ```swift
+// `URLRequestInterceptor` comes from AWSAPIPlugin, add the following imports
+import Amplify
+// If you are using Swift Package Manager
+import AWSAPIPlugin
+// or if you are using Cocoapods
+import AmplifyPlugins
+
 class MyCustomInterceptor: URLRequestInterceptor {
     func getLatestAuthToken() -> Result<String, Error> {
         let semaphore = DispatchSemaphore(value: 0)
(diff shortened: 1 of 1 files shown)


docs/cli/auth/admin.md
```diff
@@ -104,3 +108,173 @@ function App() {
...
+
+<amplify-block name="Flutter">
+
+For Amplify Flutter
+```dart
+dart code
+
+```
+
+</amplify-block>
```
lawmicha, line 277: will remove these blocks

docs/cli/auth/admin.md
```diff
@@ -104,3 +108,173 @@ function App() {
...
+                    case .success(let tokens):
+                        result = .success(tokens.idToken)
+                    case .failure(let error):
+                        result = .failure(error)
+                    }
+                }
+            case .failure(let error):
+                result = .failure(error)
+            }
+        }
```
royjit, line 233: Can we shorten this code like this:

```swift

Amplify.Auth.fetchAuthSession { (event) in
            do {
                defer {
                    semaphore.signal()
                }
                let session = try event.get()
                if let tokens = try (session as? AuthCognitoTokensProvider)?.getCognitoTokens().get() {
                    result = .success(tokens.idToken)
                }
            } catch {
                result = .failure(error)
            }
            
        }
```

docs/cli/auth/admin.md
```diff
@@ -104,3 +108,173 @@ function App() {
...
+        return result
+    }
+    
+    func intercept(_ request: URLRequest) throws -> URLRequest {
+        guard let mutableRequest = (request as NSURLRequest).mutableCopy() as? NSMutableURLRequest else {
+            throw APIError.unknown("Could not get mutable request", "")
+        }
+        
+        let tokenResult = getLatestAuthToken()
+        guard case let .success(token) = tokenResult else {
```
royjit, line 244: Bit confusing to read the code, can we do this? 

```swift
switch tokenResult {
        case .success(let token):
            mutableRequest.setValue(token, forHTTPHeaderField: "authorization")
        case .failure(let error):
            throw APIError.operationError("Failed to retrieve Cognito UserPool token.", "", error)
        }
return mutableRequest as URLRequest
```

royjit (approved):
Approved with minor code changes suggestions.

docs/cli/auth/admin.md
lawmicha: https://github.com/aws-amplify/docs/issues/3329

docs/cli/auth/admin.md
```diff
@@ -104,3 +108,143 @@ function App() {
...
+Add a custom interceptor to the API
+```swift
+try Amplify.configure()
+try Amplify.API.add(interceptor: MyCustomInterceptor(), for: "[YOUR-RESTENDPOINT-NAME]")
+```
+
+Set up the custom interceptor to return the ID token for the request.
+
+```swift
+class MyCustomInterceptor: URLRequestInterceptor {
```
frimfram, line 212: which library should be imported for URLRequestInterceptor?  It would be useful to include import statements in case people are not clear.

lawmicha: URLRequestInterceptor comes from AWSAPIPlugin, which is imported differently based on the dependency manager, ie. Cocoapods is `import AmplifyPlugins` and Swift Package Manager is `import AWSAPIPlugin`. I'll mention it here

docs/cli/auth/admin.md
```diff
@@ -104,3 +108,162 @@ function App() {
...
+```
+
+4. Use the following to list the users in the Group.
+
+```swift
+func listEditors(groupName: String, limit: Int, nextToken: String? = nil) {
+    let path = "/listUsersInGroup"
+    var query = ["groupname": groupName,
+                  "limit": String(limit)]
+    if let nextToken = nextToken {
```
attilah, line 170: nit: I'm not not a swift guru, is this a good practice to reuse the same name as the parameter?

attilah (approved):
LGTM, one nit comment about an expression

docs/cli/auth/admin.md
lawmicha: I think this is fine because we need to unwrap the parameter `nextToken` which is optional and the type completion in IDE controls what can be accessed from the variable at line 169 and line 171. Similarly, in https://docs.swift.org/swift-book/LanguageGuide/OptionalChaining.html swift docs also shows a few examples of doing the same
```swift
if let buildingNumber = buildingNumber, let street = street {
```

attilah: Thanks for the explanation :)

PR #3304, opened 2021-06-09
Code correction proposal.

santaclawn:
_Issue #, if available:_

_Description of changes:_I'm not sure, but it seems the correct parameters were missing.


gpanshu (commented):
You just changed the parameter called variables and then assigned the same to the request. This change is not required.

gpanshu:
Unfortunately this PR does not change anything and thus can not be approved or committed into the repo.

PR #3326, opened 2021-06-19
Correcting theme property name to customise the background colour

lammichael:
_Issue #:_ I don't have one.

_Description of changes:_ Correcting theme property name to customise the background colour


ericclemmons (approved):
Good catch!  Thanks for the submission

(For reference: `sectionHeader` is coded here https://github.com/aws-amplify/amplify-js/blob/fc398b4d26f76fe128c4de806f7e85f0207618d3/packages/aws-amplify-react-native/src/AmplifyTheme.ts#L50)

david-mcafee:
@lammichael this looks good to me! However, we just removed the `docs` folder, so you'll need to remove that file from this PR in order to resolve the merge conflict.

PR #3341, opened 2021-06-24
[Android] Add Android documentation for checking authentication state

changxu0306:
Added Android code snippets for checking user sign in state under `Getting Started` - `Authentication` - `Check authentication state`. 
The same change is required for iOS and Flutter as well. 


docs/lib/auth/fragments/native_common/signin/common.md
```diff
@@ -117,3 +117,9 @@ As a result, for testing purposes, you'll at least need an input field where you
 <inline-fragment platform="ios" src="~/lib/auth/fragments/ios/signin/50_multi_factor_confirm_signin.md"></inline-fragment>
 <inline-fragment platform="android" src="~/lib/auth/fragments/android/signin/50_multi_factor_confirm_signin.md"></inline-fragment>
 <inline-fragment platform="flutter" src="~/lib/auth/fragments/flutter/signin/50_multi_factor_confirm_signin.md"></inline-fragment>
+
+## Check user sign-in state
+
+Invoke the following api to determine whether the current user is signed in or not.
+
```
changxu0306, line 124: iOS and Flutter fragments should be added here

docs/lib/auth/fragments/native_common/signin/common.md
```diff
@@ -117,3 +117,9 @@ As a result, for testing purposes, you'll at least need an input field where you
 <inline-fragment platform="ios" src="~/lib/auth/fragments/ios/signin/50_multi_factor_confirm_signin.md"></inline-fragment>
 <inline-fragment platform="android" src="~/lib/auth/fragments/android/signin/50_multi_factor_confirm_signin.md"></inline-fragment>
 <inline-fragment platform="flutter" src="~/lib/auth/fragments/flutter/signin/50_multi_factor_confirm_signin.md"></inline-fragment>
+
+## Check user sign-in state
```
richardmcclellan, line 121: Since this applies to all types of sign in (Sign In, Sign In with web UI, Social sign in with web UI), can we create a new page for this instead of adding it to the existing "Sign In" page?    I'm thinking it could be a page named "Check sign in state", right after the "Sign in next steps" page.

changxu0306: Yeah, this is reasonable 👍

docs/lib/auth/fragments/native_common/signin/common.md
```diff
@@ -117,3 +117,9 @@ As a result, for testing purposes, you'll at least need an input field where you
...
+
+Invoke the following api to determine whether the current user is signed in or not.
```
richardmcclellan, line 123: The main customer use case we want to document here is how clients should determine on app start whether to show a sign in screen or their primary application screen which requires authenticated content.  That's not necessarily the same as whether the current user is signed in or not (if the refresh token expires, the end user might still be signed in according to fetchAuthSession().isSignedIn, but they are no longer authenticated because the refresh token is expired.

```suggestion
Use the following method to check if the current user is authenticated.   This is the recommended approach for determining whether you should display a sign in screen when your application starts.
```

docs/lib/auth/fragments/android/signin/60_check_sign_in_state.md
```diff
@@ -0,0 +1,55 @@
+<amplify-block-switcher>
+<amplify-block name="Java">
+
+```java
+Amplify.Auth.fetchAuthSession(
+    result -> Log.i("AuthQuickstart", result.isSignedIn() ? "User is signed in" : "User is signed out"),
```
richardmcclellan, line 6: Per https://github.com/aws-amplify/amplify-android/issues/1247, `result.isSignedIn()` is still true even after the session is expired.  I believe we determined that we need to explicitly try to obtain the token, right and verify that an exception is not thrown right?  Can you add this check?  We also need to make sure our sample app is updated and tested with this approach.

docs/lib/auth/fragments/android/signin/60_check_sign_in_state.md
```diff
@@ -0,0 +1,55 @@
+<amplify-block-switcher>
+<amplify-block name="Java">
+
+```java
+Amplify.Auth.fetchAuthSession(
```
richardmcclellan, line 5: Let's wrap this in a method that returns a boolean, since a customer would likely want to do the same anyway.  Something like:
```
private boolean isAuthenticated() {
    ...
}
```

changxu0306: Is that necessary? The code snippets provided in getting started guide are for customers to call in `onCreated()` method

richardmcclellan: Now that I think about it, we can't actually return a boolean since `fetchAuthSession` is async, so you can disregard this!

richardmcclellan: Now that I think about it, we can't actually return a boolean since `fetchAuthSession` is async, so you can disregard this!

changxu0306:
Add a new column `Check authentication state` under `Authentication`. 
Screen Shot 2021-06-29 at 12 13 03 AM
This PR also fixes some typos inside `Auth Events` and `Access AWS credentials`.

docs/lib/auth/fragments/native_common/check_authentication_state/common.md
```diff
@@ -0,0 +1,6 @@
+Use the following method to check if the current user is authenticated.   This is the recommended approach for determining whether you should display a sign in screen when your application starts.
+
+<amplify-callout>
+`fetchAuthSession().isSignedIn()` returns true even after the session is expired. It's necessary to check whether refresh token and AuthSession is still valid or not before calling `fetchAuthSession().isSignedIn()`.
+</amplify-callout>
+<inline-fragment platform="android" src="~/lib/auth/fragments/android/check_authentication_state/10_checkSignInState.md"></inline-fragment>
```
changxu0306, line 6: Note: iOS and Flutter documentation should be added here.

docs/lib/auth/fragments/native_common/check_authentication_state/common.md
```diff
@@ -0,0 +1,6 @@
+Use the following method to check if the current user is authenticated.   This is the recommended approach for determining whether you should display a sign in screen when your application starts.
```
lawmicha, line 1: suggestion

> Use the following method to check if the current user is authenticated.

Use the following method to check if the current user's authentication session is still valid.

docs/lib/auth/fragments/android/check_authentication_state/10_checkSignInState.md
```diff
@@ -0,0 +1,146 @@
...
+<amplify-block name="Java">
+
+```java
+Amplify.Auth.fetchAuthSession(
+    result -> {
+        AWSCognitoAuthSession cognitoAuthSession = (AWSCognitoAuthSession) result;
+            // Check refresh token is valid or not
+            switch (cognitoAuthSession.getUserPoolTokens().getType()) {
+                case SUCCESS:
+                    Log.i("AuthQuickStart", "Refresh token is valid.");
```
lawmicha, line 11: what is `getType()`?

docs/lib/auth/fragments/android/check_authentication_state/10_checkSignInState.md
```diff
@@ -0,0 +1,146 @@
...
+                    break;
+                case FAILURE:
+                    Log.e("AuthQuickStart", "Refresh token expired. User must re-authenticate by signing in again.", cognitoAuthSession.getUserPoolTokens().getError());
+                    // Check whether AuthSession is expired
```
lawmicha, line 15: this comment is confusing, i thought at this point in the code, the auth session is expired already

docs/lib/auth/fragments/android/check_authentication_state/10_checkSignInState.md
```diff
@@ -0,0 +1,146 @@
...
+                    subscribeToAuthEvents();
```
lawmicha, line 16: why do we need to subscribe to auth events anyways? I was expecting at this point a comment like
```
// User is no longer authenticated, display the login screen for the user to re-authenticate
``` 

why don't we subscribe to auth events regardless of the result of `cognitoAuthSession.getUserPoolTokens().getType()` ?

docs/lib/auth/fragments/android/check_authentication_state/10_checkSignInState.md
```diff
@@ -0,0 +1,146 @@
...
+        Log.i("AuthQuickstart", result.isSignedIn() ? "User is signed in" : "User is signed out");
+    },
+    error -> Log.e("AuthQuickstart", "Failed to determine user sign-in state" + error.toString())
+);
+
+private void subscribeToAuthEvents() {
+    Amplify.Hub.subscribe(HubChannel.AUTH,
+            hubEvent -> {
+                if (AuthChannelEventName.valueOf(hubEvent.getName()) == AuthChannelEventName.SESSION_EXPIRED) {
+                    Log.i("AuthQuickstart", "Auth session just expired.");
```
lawmicha, line 27: what to do here? should there be a comment like
```
// User is no longer authenticated, navigate to the login screen for the user to re-authenticate
```

docs/lib/auth/fragments/native_common/check_authentication_state/common.md
```diff
@@ -0,0 +1,6 @@
...
+
+<amplify-callout>
+`fetchAuthSession().isSignedIn()` returns true even after the session is expired. It's necessary to check whether refresh token and AuthSession is still valid or not before calling `fetchAuthSession().isSignedIn()`.
```
lawmicha, line 4: - In what use case do I use ``fetchAuthSession().isSignedIn()`? When do I not care whether the session is expired or not?
- as a reader, i don't know what a refresh token is, perhaps "A valid refresh token indicates that the session is still valid".
- what is AuthSession? is it the authentication session? why do we have to check both the AuthSession and the refresh token? why do I have to check `fetchAuthSession().isSignedIn()` afterwards?

docs/lib/auth/fragments/android/check_authentication_state/10_checkSignInState.md
```diff
@@ -0,0 +1,91 @@
...
+```kotlin
+try {
+    val result = Amplify.Auth.fetchAuthSession()
+    val cognitoAuthSession = result as AWSCognitoAuthSession
+    // Check refresh token is valid or not
+    when (cognitoAuthSession.userPoolTokens.type) {
+        AuthSessionResult.Type.SUCCESS ->
+            Log.i("AuthQuickStart", "Refresh token is valid.")
+        AuthSessionResult.Type.FAILURE -> 
+                // User is no longer authenticated, display the login screen for the user to re-authenticate
```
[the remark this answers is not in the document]
changxu0306: It seems that the answer is yes.

src/fragments/lib/auth/native_common/check_auth_state/common.mdx
```diff
@@ -0,0 +1,11 @@
...
+
+import android0 from "/src/fragments/lib/auth/android/check_auth_state/10_checkSignInState.mdx";
+
+<Fragments fragments={{android: android0}} />
+
+You can determine the action of your app when the authentication session expires by invoking authentication hub events:
+
+import android1 from "/src/fragments/lib/auth/android/check_auth_state/20_sessionExpiredEvent.mdx";
+
+<Fragments fragments={{android: android1}} />
```
eeatonaws, line 11: Note: iOS and Flutter documentation should be added here.

docs/lib/auth/fragments/native_common/check_authentication_state/common.md
```diff
@@ -0,0 +1,6 @@
+Use the following method to check if the current user's authentication session is still valid.   This is the recommended approach for determining whether you should display a sign in screen when your application starts.
```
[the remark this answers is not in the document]
eeatonaws: From @changxu0306's [comment](https://github.com/aws-amplify/docs/pull/3341#discussion_r662573744), it seems that if the device is offline and the token is expired, using this approach will produce a result of type AuthSessionResult Failure.

mauerbac:
Closing old PR

PR #3259, opened 2021-05-25
Flutter 2.2.0 Instructions

offlineprogrammer:
*Issue #, if available:* [Instructions for Flutter 2.2.0 compatibility with Amplify](https://github.com/aws-amplify/docs/issues/3258)

*Description of changes:* Add instructions to disable sound null safety

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3347, opened 2021-06-26
chore(flutter): adds null safety docs for amplify-flutter

haverchuck:
_Description of changes:

Adds null safety docs for amplify-flutter.

Note: feature flag section of CLI docs will need to be updated once we know the version number of the CLI which will first include the dart feature flag.

### DIFF
--- a/docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
+++ b/docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
@@ -3,11 +3,13 @@
 
 The Amplify Flutter library supports [Dart null safety](https://dart.dev/null-safety) beginning with version 0.2.0.  This means that:
 
+
+
 * Applications utilizing [sound null safety](https://dart.dev/null-safety) must use Amplify Flutter version 0.2.0 or greater.
 
 * Applications utilizing [unsound null safety](https://dart.dev/null-safety/unsound-null-safety) can use versions of Amplify Flutter above or below 0.2.0, but some application changes may be necessary to migrate to 0.2.0.
 
-* Applications that are not null safe must use a version of Amplify Flutter under 0.2.0.
+* Applications that are not using Flutter 2.0 or higher must use a version of Amplify Flutter lower than 0.2.0.
 
 
 **Using Generated Codegen Models with Null Safety**
--- a/docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
+++ b/docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
@@ -15,4 +15,4 @@ The Amplify Flutter library supports [Dart null safety](https://dart.dev/null-sa
 
 The Amplify CLI's `codegen models` command can generate Dart models with or without null safety. To generate null safe models, make sure that the `pubspec.yaml` file at the root of your projects defines a Dart SDK version of 2.12.0 or greater, and make sure that the `enableDartNullSafety` [feature flag](https://docs.amplify.aws/cli/reference/feature-flags) is set to true in your `amplify/cli.json` file.
 
-In order to migrate to Amplify Flutter version 0.2.0 or greater, you will need to regenerate your codegen models with null safety if you are using them.
+In order to migrate to Amplify Flutter version 0.2.0 or greater, you will need to regenerate your codegen models with null safety (using the `amplify codegen models` command) if you are using them.
--- a/docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
+++ b/docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
@@ -23,10 +23,10 @@ To migrate to null safe models, you can simply regenerate them following the ins
 1. Make sure that the `pubspec.yaml` file at the root of your projects defines a Dart SDK version of 2.12.0 or greater.
 2. Update your Amplify CLI to version 5.1.0 or higher.(snippet for npm command)
 3. Make sure that the `enableDartNullSafety` [feature flag](https://docs.amplify.aws/cli/reference/feature-flags) is set to "true" in your `amplify/cli.json` file. 
-4. Re-run amplify codegen models if you have a schema defined.
+4. Re-run `amplify codegen models` for your schema.
 
 ***Opting-out of Null Safety***
 
 If you wish to continue using non-null safe models:
 1. Make sure that the `enableDartNullSafety` [feature flag](https://docs.amplify.aws/cli/reference/feature-flags) is set to "false".
-2. Re-run amplify codegen models for your schema 
+2. Re-run `amplify codegen models` for your schema 
--- a/docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
+++ b/docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
@@ -13,6 +13,6 @@ The Amplify Flutter library supports [Dart null safety](https://dart.dev/null-sa
 **Using Generated Codegen Models with Null Safety**
 
 
-The Amplify CLI's `codegen models` command can generate Dart models with or without null safety. To generate null safe models, make sure that the `pubspec.yaml` file at the root of your projects defines a Dart SDK version of 2.12.0 or greater, and make sure that the `enableDartNullSafety` [feature flag](https://docs.amplify.aws/cli/reference/feature-flags) is set to true.
+The Amplify CLI's `codegen models` command can generate Dart models with or without null safety. To generate null safe models, make sure that the `pubspec.yaml` file at the root of your projects defines a Dart SDK version of 2.12.0 or greater, and make sure that the `enableDartNullSafety` [feature flag](https://docs.amplify.aws/cli/reference/feature-flags) is set to true in your `amplify/cli.json` file.
 
 In order to migrate to Amplify Flutter version 0.2.0 or greater, you will need to regenerate your codegen models with null safety if you are using them.
--- a/docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
+++ b/docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
@@ -1,0 +1,18 @@
+**Amplify Flutter and Null Safety**
+
+
+The Amplify Flutter library supports [Dart null safety](https://dart.dev/null-safety) beginning with version 0.2.0.  This means that:
+
+* Applications utilizing [sound null safety](https://dart.dev/null-safety) must use Amplify Flutter version 0.2.0 or greater.
+
+* Applications utilizing [unsound null safety](https://dart.dev/null-safety/unsound-null-safety) can use versions of Amplify Flutter above or below 0.2.0, but some application changes may be necessary to migrate to 0.2.0.
+
+* Applications that are not null safe cannot use a version of Amplify Flutter 0.2.0 or greater.
+
+
+**Using Generated Codegen Models with Null Safety**
+
+
+The Amplify CLI's `codegen models` command can generate Dart models with or without null safety. To generate null safe models, make sure that the `pubspec.yaml` file at the root of your projects defines a Dart SDK version of 2.12.0 or greater, and make sure that the `enableDartNullSafety` [feature flag](https://docs.amplify.aws/cli/reference/feature-flags) is set to true.
+
+In order to migrate to Amplify Flutter version 0.2.0 or greater, you will need to regenerate your codegen models with null safety if you are using them.
--- a/docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
+++ b/docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
@@ -11,9 +11,22 @@ The Amplify Flutter library supports [Dart null safety](https://dart.dev/null-sa
 
 
 
-**Using Generated Codegen Models with Null Safety**
+**DataStore with Code Generated Models and Null Safety**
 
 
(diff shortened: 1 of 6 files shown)


docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
```diff
@@ -0,0 +1,18 @@
...
+
+* Applications utilizing [unsound null safety](https://dart.dev/null-safety/unsound-null-safety) can use versions of Amplify Flutter above or below 0.2.0, but some application changes may be necessary to migrate to 0.2.0.
+
+* Applications that are not null safe must use a version of Amplify Flutter under 0.2.0.
+
+
+**Using Generated Codegen Models with Null Safety**
+
+
+The Amplify CLI's `codegen models` command can generate Dart models with or without null safety. To generate null safe models, make sure that the `pubspec.yaml` file at the root of your projects defines a Dart SDK version of 2.12.0 or greater, and make sure that the `enableDartNullSafety` [feature flag](https://docs.amplify.aws/cli/reference/feature-flags) is set to true.
```
fjnoyp, line 16: Can we say “set to true in your cli.json file under (it’s file path).”  This way customers don’t need to read all that documentation

docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
```diff
@@ -0,0 +1,18 @@
+**Amplify Flutter and Null Safety**
+
+
+The Amplify Flutter library supports [Dart null safety](https://dart.dev/null-safety) beginning with version 0.2.0.  This means that:
+
+* Applications utilizing [sound null safety](https://dart.dev/null-safety) must use Amplify Flutter version 0.2.0 or greater.
+
+* Applications utilizing [unsound null safety](https://dart.dev/null-safety/unsound-null-safety) can use versions of Amplify Flutter above or below 0.2.0, but some application changes may be necessary to migrate to 0.2.0.
+
```
fjnoyp, line 9: Do you think it would be easier to make a table of what works and doesn’t.  This way we can have less text and make it faster for customers to understand what works with what.

docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
```diff
@@ -0,0 +1,18 @@
...
+* Applications that are not null safe must use a version of Amplify Flutter under 0.2.0.
+
+
+**Using Generated Codegen Models with Null Safety**
+
+
+The Amplify CLI's `codegen models` command can generate Dart models with or without null safety. To generate null safe models, make sure that the `pubspec.yaml` file at the root of your projects defines a Dart SDK version of 2.12.0 or greater, and make sure that the `enableDartNullSafety` [feature flag](https://docs.amplify.aws/cli/reference/feature-flags) is set to true.
+
+In order to migrate to Amplify Flutter version 0.2.0 or greater, you will need to regenerate your codegen models with null safety if you are using them.
```
fjnoyp, line 18: Nit: maybe saying something like regenerate by calling the ‘amplify codegen models’ just in case they forgot?

fjnoyp (commented):
Some minor comments to make things clearer for customers

docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
haverchuck: Personally I prefer it as-is, but we can change it if you feel strongly.

docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
haverchuck: Done.

fjnoyp (approved):

docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
offlineprogrammer: +1 for the table idea

offlineprogrammer (approved):
Looks good 
small comment re using table

docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
Jordan-Nelson: What are you think for a table? Are you thinking of something like the table below? I personally don't see a ton of value in a table since its not much info. I also sort of prefer the way it is, but I think either way gets the info across.


|                                	| amplify-flutter 0.1.6 and below   	| amplify-flutter 0.2.0 and above  	|
|-------------------	|---------------------------------	|---------------------------------	|
| Non Null Safe App 	| Supported                                    	| Supported                                    	|
| Null Safe App     	    | Not Supported                             	| Supported                                    	|

Jordan-Nelson: I missed the third bullet initially. I guess you could add a third row to that table for the third bullet. I don't think you can have a null safe app that isn't running w/flutter 2+ so I don't think we really have to consider that.

|                                	                     | amplify-flutter 0.1.6 and below   	| amplify-flutter 0.2.0 and above  	|
|-------------------------------	|---------------------------------	|---------------------------------	|
| Null Safe App     	                         | Not Supported                             	| Supported                                    	|
| Non Null Safe App w/ flutter v2  | Supported                                    	| Supported                                    	|
| Non Null Safe App w/ flutter v1 	| Supported                                    	| Not Supported                                    	|

docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
```diff
@@ -0,0 +1,20 @@
+**Amplify Flutter and Null Safety**
+
+
+The Amplify Flutter library supports [Dart null safety](https://dart.dev/null-safety) beginning with version 0.2.0.  This means that:
+
+
+
+* Applications utilizing [sound null safety](https://dart.dev/null-safety) must use Amplify Flutter version 0.2.0 or greater.
```
Jordan-Nelson, line 8:
```suggestion
* Applications utilizing [sound null safety](https://dart.dev/null-safety) should use Amplify Flutter version 0.2.0 or greater.
```

Technically someone with a null safe app can use any version if they run it with `--no-sound-null-safety`. We actually have that in our docs for people that created apps with flutter 2.2+. I would just change this to "should" since it isn't technically a requirement.

docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
```diff
@@ -0,0 +1,20 @@
...
+
+* Applications utilizing [unsound null safety](https://dart.dev/null-safety/unsound-null-safety) can use versions of Amplify Flutter above or below 0.2.0, but some application changes may be necessary to migrate to 0.2.0.
+
+* Applications that are not using Flutter 2.0 or higher must use a version of Amplify Flutter lower than 0.2.0.
+
+
+**Using Generated Codegen Models with Null Safety**
+
+
+The Amplify CLI's `codegen models` command can generate Dart models with or without null safety. To generate null safe models, make sure that the `pubspec.yaml` file at the root of your projects defines a Dart SDK version of 2.12.0 or greater, and make sure that the `enableDartNullSafety` [feature flag](https://docs.amplify.aws/cli/reference/feature-flags) is set to true in your `amplify/cli.json` file.
```
Jordan-Nelson, line 18: This will also require the latest version of the CLI, right? We might want to include what version that is and how to check what version they are using.

docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
```diff
@@ -0,0 +1,20 @@
...
+
+In order to migrate to Amplify Flutter version 0.2.0 or greater, you will need to regenerate your codegen models with null safety (using the `amplify codegen models` command) if you are using them.
```
Jordan-Nelson, line 20: I think we should include a migration path to null safety for apps using datastore. For all other packages, the upgrade path is no different that any other package (upgrade all dependencies, then upgrade the app). Data store has a third step of generating the new models.

The steps aren't really that much different, but I think we should just be specific:
1. Confirm your app is working with the latest non null safe version of amplify-flutter
2. Upgrade amplify-flutter to 0.2.0 and make updates for breaking changes (see package change logs for details).
3. Upgrade app to null safety: For the application code follow flutter docs, for the model code re-generate the models

Open question if steps 2 and 3 should combined. I believe the app should work in the mixed state (non null safe app, non null safe models, null safe amplify-flutter). If it doesn't, they should probably be listed as one step w/ a specific callout that both steps need to be done together.

Jordan-Nelson:
Do we want to back out the changes from this PR - https://github.com/aws-amplify/docs/pull/3259/files? We could do it in a separate PR as well

docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
Jordan-Nelson: I think the open question has been answered. Steps 2 and 3 can be distinct steps and the application should function as expected after each step. With that in mind, rather than listing out the migration steps as I suggested above, I think we could probably just update this line to be:

> If your app uses generated models from amplify, you will need to migrate these models to be null safe as part of migrating your app to be null safe. You should follow the [flutter null safety migration docs](https://dart.dev/null-safety/migration-guide) to migrate your application code, excluding the generated models. To migrate the models to be null safe, you can simply regenerate them following the instructions above.

I think as is, the line is not correct. You do not have to regenerate models to use 0.2.0. You just have to regenerate models to migrate your app to null safety.

Jordan-Nelson (approved):

cshfang (approved):
:shipit:

docs/lib/project-setup/fragments/native_common/prereq/flutter_null_safety.md
```diff
@@ -1,13 +1,6 @@
 
-<amplify-callout warning>
+<amplify-callout>
```
cshfang, line 2: Actually, I guess I should call out that the follow-up PR is just removing these callouts altogether except where we talk about codegen. Seeing as how null safety will become the new default, I think it's probably reasonable to not mention it explicitly going forward?

ragingsquirrel3 (approved):

docs/lib/project-setup/fragments/native_common/prereq/flutter_null_safety.md
haverchuck: I think we should leave these up, at least temporarily.

Ashish-Nanda: I agree we should leave it in for now, especially while its not enabled by default in the CLI codegen for existing projects. Right now its opt-in for existing projects. Its only enabled by default for new projects.

Once we move to opt-out as the default for all projects, we can remove these callouts.

cshfang: Ok, I will restore the callouts and let this PR update the copy

docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
```diff
@@ -0,0 +1,19 @@
...
+
+|                                	                     | amplify-flutter 0.1.6 and below   	| amplify-flutter 0.2.0 and above  	|
+|-------------------------------	|---------------------------------	|---------------------------------	|
+| Null Safe App     	                         | Not Supported                             	| Supported                                    	|
+| Non Null Safe App w/ flutter v2  | Supported                                    	| Supported                                    	|
+| Non Null Safe App w/ flutter v1 	| Supported                                    	| Not Supported                                    	|
+
+
+
+**Using Generated Codegen Models with Null Safety**
```
Ashish-Nanda, line 14: I would just say `DataStore with Code Generated Models and Null Safety`

docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
```diff
@@ -0,0 +1,19 @@
...
+
+
+The latest version of the Amplify CLI can generate Dart models with or without null safety using the `codegen models` command. To generate null safe models, make sure that the `pubspec.yaml` file at the root of your projects defines a Dart SDK version of 2.12.0 or greater, and make sure that the `enableDartNullSafety` [feature flag](https://docs.amplify.aws/cli/reference/feature-flags) is set to true in your `amplify/cli.json` file. You may need to update your Amplify CLI to version 5.1.0 or higher.
```
Ashish-Nanda, line 17: You may need to -> You will need to update your Amplify CLI to version 5.1.0 or higher

Ashish-Nanda: Consider having these instructions as separate bullet points or a numbered list. 
Also we need to have the instructions.

docs/lib/project-setup/fragments/flutter/null-safety/null-safety.md
```diff
@@ -0,0 +1,32 @@
...
+The latest version of the Amplify CLI can generate Dart models with or without null safety using the `codegen models` command. 
+
+***Opting-in to Null Safety***
+
+* If you have a null safe app, or are migrating to null safety and your app uses generated models from amplify, you will need ensure the models are null safe as well. You should follow the [flutter null safety migration docs](https://dart.dev/null-safety/migration-guide) to migrate your application code, excluding the generated models. 
+To migrate to null safe models, you can simply regenerate them following the instructions:
+1. Make sure that the `pubspec.yaml` file at the root of your projects defines a Dart SDK version of 2.12.0 or greater.
+2. Update your Amplify CLI to version 5.1.0 or higher.(snippet for npm command)
+3. Make sure that the `enableDartNullSafety` [feature flag](https://docs.amplify.aws/cli/reference/feature-flags) is set to "true" in your `amplify/cli.json` file. 
+4. Re-run amplify codegen models if you have a schema defined.
```
Ashish-Nanda: Can we make this the same as the one for opt-out:
`Re-run amplify codegen models for your schema`

Ashish-Nanda (approved):
Looks good 🚀

PR #3378, opened 2021-07-02
chore(auth): SMS flows for Flutter

dnys1:
_Issue #, if available:_
- User request for SMS/OTP based flow docs

_Description of changes:_
- Add docs describing SMS based flows with Flutter snippets


offlineprogrammer (approved):

docs/lib/auth/fragments/flutter/phone/add_mfa.md
```diff
@@ -0,0 +1,23 @@
+Turning MFA "ON" will make it required for all users, while "Optional" will make it available to enable on a per-user basis.
```
Jordan-Nelson, line 1: *Question:* The majority of the docs below apply to iOS and Android as well (maybe even JS?), correct? Should these docs be added for all platforms, and then just the code snippets be added as fragments for each platform?

I am not sure if there is a norm for how this is done for totally new sections of docs

dnys1: For sure, yeah, and they can easily be turned on for other teams by just removing `platform="flutter"` but it didn't seem right to enable it for all teams at once.

dnys1: @Jordan-Nelson I've updated the tags to make it clear which ones are actually flutter-specific and moved the common text to a common folder.

docs/lib/auth/fragments/common/sms/flows.md
```diff
@@ -0,0 +1,130 @@
+There are a few ways to integrate phone numbers into an Amplify app's sign-in and verification process. 
+- **As a username**: Users login with a username and password where their phone number acts as the username.
```
haverchuck, line 2: I think people often get confused on the distinctions between using a phone number AS as username and signin in WITH a phone_number.  In this first bullet point we collapse that distinction a bit; but making the differences clear would take a fair amount of space. Maybe to keep this section a little cleaner we could stick to only the SMS stuff (and possibly discuss these distinctions in another section)?

haverchuck (changes requested):
Left one comment in line.

haverchuck (approved):

Issue #3380, opened 2021-07-04
[Feedback]FEEDBACK_TITLE_HERE


DanielCollins96:
**Page**: [`/ui/auth/authenticator/q/framework/angular`](https://docs.amplify.aws/ui/auth/authenticator/q/framework/angular#secondary-nav)

**Feedback**:
in Custom CSS Properties, typo of wegigh?
--footer-wegight

drochetti:
It seems this was fixed in [this commit](https://github.com/aws-amplify/amplify-js/commit/9bc6a3c785a8b288c35db243541187303c2103df) / [this PR](https://github.com/aws-amplify/amplify-js/pull/8316) but the process that regenerates that part of the docs is not being triggered as part of the docsite build.

I don't know what exactly pulls the source code content into the docs, but find that and you will find redemption. 😄

drochetti:
Based on [this](https://github.com/aws-amplify/docs/blob/81a8849c6178171988c74d7b53ed29fa4c8f8cac/client/src/docs-ui/ui-component-props/ui-component-props.tsx#L2), the docs are rendered by the `ui-component-props` component/tag. It imports `@aws-amplify/ui-components/dist/docs`:

https://github.com/aws-amplify/docs/blob/81a8849c6178171988c74d7b53ed29fa4c8f8cac/client/src/docs-ui/ui-component-props/ui-component-props.tsx#L2

That file is indeed outdated in the `dist/` folder of `@aws-amplify/ui-components`. It seems like a Stencil build phase is responsible for generating that file:

https://github.com/aws-amplify/amplify-js/blob/53175248fa73d43860974dcd565b6681882e7a45/packages/amplify-ui-components/stencil.config.ts#L79

drochetti:
Okay, did some more digging and turns out the Amplify dependencies are outdated. The versions in `package.json` are set to the string `"latest"`, which is a side-effect of a Yarn bug: https://github.com/yarnpkg/yarn/issues/1306

https://github.com/aws-amplify/docs/blob/5ff409c5c4a0cc97197f02c39da29f0a9b7df8b7/package.json#L15-L16

```
docs@2.0.0 /dev/aws-amplify/amplify-docs
└── aws-amplify@3.3.15
docs@2.0.0 /dev/aws-amplify/amplify-docs
└── @aws-amplify/ui-components@0.10.1
```

PR #3415, opened 2021-07-22
fix amplify dependencies

drochetti:
_Issue #, if available:_ #3380 

_Description of changes:_

Some of the docs that are auto-generated from the `@aws-amplify/ui-components` source code are out-dated due to a wrong dependency definition in the `package.json`. Due to a Yarn bug, when trying to add the dependency tagged with `latest`, Yarn adds the string `latest` as the dependency version, which results in the wrong version getting installed:

https://github.com/aws-amplify/docs/blob/5ff409c5c4a0cc97197f02c39da29f0a9b7df8b7/package.json#L15-L16

Currently, on `main` the following versions are present (`npm list` command):

```
docs@2.0.0 /dev/aws-amplify/amplify-docs
└── aws-amplify@3.3.15  invalid
docs@2.0.0 /dev/aws-amplify/amplify-docs
└── @aws-amplify/ui-components@0.10.1  invalid
```

This PR fixes it by reinstalling the dependencies to get their most recent version. After the changes the dependency tree is:

```
docs@2.0.0 /dev/aws-amplify/amplify-docs
└── aws-amplify@4.2.0 
docs@2.0.0 /dev/aws-amplify/amplify-docs
└── @aws-amplify/ui-components@1.6.2 
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,9 +12,9 @@
     "capi"
   ],
   "dependencies": {
-    "aws-amplify": "latest",
-    "@aws-amplify/ui-components": "latest",
+    "@aws-amplify/ui-components": "^1.6.2",
     "array-flatten": "^3.0.0",
+    "aws-amplify": "^4.2.0",
     "copy-to-clipboard": "^3.2.1",
     "emotion": "^10.0.23"
   },
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,89 +2,92 @@
 # yarn lockfile v1
 
 
-"@aws-amplify/analytics@4.0.6":
-  version "4.0.6"
-  resolved "https://registry.yarnpkg.com/@aws-amplify/analytics/-/analytics-4.0.6.tgz#1c62a616caee1587660296a05eda316b1094af6b"
-  integrity sha512-HVFb6NxpvAomeZ+caai6qz0kIzsKwp3lpNUM32cmYID+1qgg4BfH5u+Ol3/0nSAJO076+Nl+GwLuuoOTiEEanQ==
+"@aws-amplify/analytics@5.0.6":
+  version "5.0.6"
+  resolved "https://registry.yarnpkg.com/@aws-amplify/analytics/-/analytics-5.0.6.tgz#e65afb0f2d4eddb583b767816a72fc8ab1c35c31"
+  integrity sha512-81VO8zBgaTXPagqHrzxvVjpH/shmjCYbMLeMe15aGRjqHXpmWFFUNFomBb3aeXyB3fftKMcMIs8F7gS5qrl6ug==
   dependencies:
-    "@aws-amplify/cache" "3.1.43"
-    "@aws-amplify/core" "3.8.10"
-    "@aws-sdk/client-firehose" "1.0.0-rc.4"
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -12,9 +12,9 @@
     "capi"
   ],
   "dependencies": {
-    "aws-amplify": "latest",
-    "@aws-amplify/ui-components": "latest",
+    "@aws-amplify/ui-components": "^1.6.2",
     "array-flatten": "^3.0.0",
+    "aws-amplify": "^4.2.0",
```
ericclemmons, line 17: I believe that `latest` is correct because we want our docs to use the latest artifacts & published code.  Even with `latest`, we can see that a very old version was "locked" here:
https://github.com/aws-amplify/docs/pull/3415/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deL147

So the problem appears to be `yarn.lock` doing what it's intended to do – deterministic installs.

I don't know the best way to tell `yarn` or `npm` "yes, this is a dependency, but don't lock the version because we always want the `latest` from NPM".

Technically, that'd make these `peerDependencies`, but then that means both contributors and CI would need to `yarn install aws-amplify @aws-amplify/ui-components` before running the project.

drochetti: Só how to use latest and get Yarn to behave correctly and update `yarn.lock`?

jakeburden: In the [amplify prebuild phase](https://github.com/aws-amplify/docs/blob/5c17716ca465d0e3dab64f8e3573c4e06f79feaa/amplify.yml#L4-L6), could we add the [`upgrade --latest`](https://classic.yarnpkg.com/en/docs/cli/upgrade/#toc-yarn-upgrade-package-latest-l-caret-tilde-exact-pattern) flag for `aws-amplify` and `@aws-amplify/ui-components`?

This doesn't solve the issue for local development, but I think this would at least allow the deployed site to always be get the most recent version.

```
yarn upgrade aws-amplify @aws-amplify/ui-components --latest
```

drochetti: I understand the intent of always pointing to the latest `amplify` in the docs, but it seems like `latest` is not giving us what we need. The previous config remained outdated almost a year, which is a concerning outcome. So I agree with the premise but it doesn't seem like keeping `latest` is correct unless we can adapt the development and the CI/CD process to get the result we intend.

@mauerbac Do you have any thoughts on this? It seems like there's a missing link between the Amplify release process and the Docs dependency on it.

jakeburden: @drochetti  @ericclemmons  What if we add a [`dependabot config`](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates) so that dependabot opens a docs PR when `amplify` updates?

image

jakeburden: @drochetti @ericclemmons I've opened a PR to add a dependabot config for this:
https://github.com/aws-amplify/docs/pull/3607

mauerbac:
ping on this :)

ericclemmons:
This was resolved & replaced with https://github.com/aws-amplify/docs/pull/3607

PR #3470, opened 2021-08-12
update clear and start method notes

chrisbonifacio:
_Issue #, if available:_
https://github.com/aws-amplify/amplify-js/issues/8658

_Description of changes:_

Adding note for `DataStore.clear` that mentions it will only clear local data if called after DataStore has started and refers to the more detailed description of starting DataStore either manually or automatically further below.

As suggested by @iartemiev, and to keep with the other framework documentation, we're including the other DataStore methods that trigger the start/sync process.


siegerts (approved):

src/fragments/lib/datastore/js/other-methods.mdx
```diff
@@ -24,4 +26,4 @@ import { DataStore } from 'aws-amplify';
 await DataStore.start();
 ```
 
-Synchronization starts automatically whenever you run `DataStore.query()` however you can explicitly begin the process with `DataStore.start()`.
\ No newline at end of file
+Synchronization starts automatically whenever you run any of the following methods: `DataStore.query()`, `DataStore.save()`, `DataStore.delete()`, or `DataStore.observe()`.
```
svidgen, line 29: `observeQuery()` should also `start()` it, right?

svidgen (approved):
Approving. I'm not sure my comment about `observeQuery` is relevant. (I don't think it's documented yet, is it?)

chrisbonifacio:
> Approving. I'm not sure my comment about `observeQuery` is relevant. (I don't think it's documented yet, is it?)

doesn't seem to be documented yet

PR #3482, opened 2021-08-16
Fix bugs in Vue tutorial code

coisme:
Fixed two bugs below:

* Name mismatch between `template` and `script` section, which leads to showing initial page
* Requires `async` for `subscribe()`

### DIFF
--- a/docs/start/getting-started/fragments/vue/data-model.md
+++ b/docs/start/getting-started/fragments/vue/data-model.md
@@ -197,7 +197,7 @@ import { createTodo } from './graphql/mutations';
 import { listTodos } from './graphql/queries';
 
 export default {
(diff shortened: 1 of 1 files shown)


Milan-Shah (approved):

docs/start/getting-started/fragments/vue/data-model.md
```diff
@@ -249,7 +249,7 @@ export default {
   },
   methods: {
     // other methods
-    subscribe() {
+    async subscribe() {
```
ErikCH, line 252: This code has no 'await', why does it have an `async` here?

ErikCH:
@coisme Thanks for the PR. Please also fix the conflict.

PR #3485, opened 2021-08-16
Migrate to Next.js

jakeburden:
### Preview URL:
https://next-docs.amplify.aws/

_Description of changes:_

* Moved away from using Capi/Stencil as our docs build system.
* Moved to using Next.js and React, utilizing continuously improving open source tools.
* By utilizing React + Next.js, we also enable more contributions, a faster development feedback loop, and more feature-rich improvements to the site.
* Docs markdown is now parsed with MDX - meaning you can use React components in markdown files.
* As Amplify Console natively supports Next.js Server-Side Rendering (SSR), we’ll be able to leverage our own product to add more advancements to the docs site such as internationalization (i18N) and image optimization.
* By utilizing Next.js/MDX we can fine-tune and have greater control over meta information parsed by search engines to help with SEO ranking. We can do this by passing SEO friendly meta-data to the `next/head` component. Since all content is static generated (and SSR in the future), search engines will be able to parse all content immediately.
 * Next.js enables us to choose which pages will be statically generated for speed and performance, and which pages will be server-side generated to provide more feature-rich experiences that require a sever runtime.
 
 
 The content authoring experience should remain mostly the same, with three key differences:
1. Rather than using `<inline-fragment>` we now use `<Fragments>`, and those fragments live in the `/src/fragments` directory.
     Example:
     ```jsx
    import js from "/src/fragments/lib/datastore/js/conflict.mdx";

    <Fragments fragments={{js: js}} />;
    ```
2. Rather than using `<amplify-block>` and `<amplify-block-switcher>` we now use `<Block>` and `<BlockSwitcher>`
    Example:
     ````md
    <BlockSwitcher>
    
    <Block name="JavaScript">
    
    ```js
    const a = "a";
    ```
    
    </Block>
    
    <Block name="TypeScript">
    
    ```ts
    const a: "a" = "a";
    ```
    
    </Block>
    
    <Block name="Rust">
    
    ```rust
    let mut a = String::from("a");
    ```
    
    </Block>
    
    </BlockSwitcher>
    ````
  3. Rather than using `<ui-component-props>`, we now use `<UiComponentProps>`
      Example:
      `<UiComponentProps tag="amplify-s3-image" propType="attr" useTableHeaders />`


tjleing (approved):

mauerbac (approved):
Reviewed offline. Approved

PR #3498, opened 2021-08-17
Use menuIsOpen state to determine the width of content

jakeburden:
_Issue #, if available:_

Resolves the issue of content being too close to the right edge of the mobile screen:

image

_Description of changes:_

Passes `menuIsOpen` to content styles to determine if content should be full width or not.

When the menu is open, set the content to be 100vw, so that it doesn't appear "scrunched" on the side.
When the menu is closed, set the content to be fluid, so it fills the space with proper padding.

|  Menu Open  | Menu Closed  | 
|---|---|
|  image  | image |

### DIFF
--- a/src/components/Menu/index.tsx
+++ b/src/components/Menu/index.tsx
@@ -21,6 +21,7 @@ type MenuProps = {
   filterKey: string;
   pathname: string;
   href: string;
+  setMenuIsOpen?: any;
 };
 
 type MenuState = {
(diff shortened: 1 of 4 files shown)


medelman17:
LGTM

medelman17 (approved):

tjleing (approved):
Unfortunate that it has to be all of this overhead to track whether the menu is open or not.  Sorry, I think I was the one who made that 100vw change 😅

PR #3508, opened 2021-08-18
(fix) FilterSelect not showing correct text on mobile

tjleing:
soln: pass `filterKind` through dependency chain so that we don't have to rederive it from the router url which doesn't exist at render time

image

as a bonus, this also fixes the screamingfrog 404s -- they were coming from the prerendered FilterSelect dropdown:
image
vs.
image


jakeburden:
Staging URL: https://leint-next-mobile-mount.d1oosfztpocl9c.amplifyapp.com/

jakeburden (approved):

offlineprogrammer (approved):

PR #3509, opened 2021-08-18
(fix) UI switcher has correct hrefs for both sides on ChooseFilterPages

tjleing:
Problem: /ui pages like https://docs.amplify.aws/ui/ where you have to choose a framework (`ChooseFilterPage`s) have broken links on the UI/UI-legacy version switcher.  Hovering the "Legacy" button, see the link at the bottom of the image:

image

Solution: add hrefs of `ChooseFilterPage`s to the lists of hrefs that are allowed to appear in a UI switcher; also don't result in undefined when on a page with no filterKey (/ui/**/q/framework/[filterKey]).

For example, `/ui` results in `filter = ""` instead of `filter = undefined`, and then is allowed to appear in the UI switcher.

- also the href lists for UI and UI-legacy are now created differently.  I don't think this ever manifested into a problem, but the way in this PR is more accurate to the variable names.

### DIFF
--- a/src/components/Menu/VersionSwitcher/index.tsx
+++ b/src/components/Menu/VersionSwitcher/index.tsx
@@ -2,22 +2,29 @@ import Link from "next/link";
 import {ActiveSwitchStyle, SwitchStyle} from "./styles";
 import directory from "../../../directory/directory";
 
-const uiLegacy = directory["ui-legacy"];
-const {items} = uiLegacy;
+const ui = directory["ui"].items;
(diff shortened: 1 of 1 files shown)


offlineprogrammer (approved):

jakeburden:
Staging URL: https://leint-next-fix-ui-switcher.d1oosfztpocl9c.amplifyapp.com

jakeburden (approved):

PR #3578, opened 2021-09-13
chore: Add docs of Flutter DataStore start and stop APIs

HuiSF:
_Issue #, if available:_

_Description of changes:_

Add "Other methods" section to amplify-flutter DataStore documentation.

----

### DIFF
--- a/src/directory/directory.js
+++ b/src/directory/directory.js
@@ -381,7 +381,7 @@ const directory = {
           {
             title: "Other methods",
             route: "/lib/datastore/other-methods",
-            filters: ["android", "ios", "js"],
+            filters: ["android", "flutter", "ios", "js"],
           },
           {
             title: "Schema updates",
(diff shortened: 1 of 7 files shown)


src/directory/directory.js
```diff
@@ -381,7 +381,7 @@ const directory = {
           {
             title: "Other methods",
             route: "/lib/datastore/other-methods",
-            filters: ["android", "ios", "js"],
+            filters: ["android", "flutter", "ios", "js"],
```
Jordan-Nelson, line 384: I am not sure how the filters work, but if all platforms have this section, can this simply be empty?

HuiSF: Yes that works, but I decided to follow what exists in the codebase already... listing all possible platforms may also be a bit clearer than an empty array in the case that any new platform could be added in the future.

Jordan-Nelson: Ah okay. I saw all the empty filters below this in the diff and assumed that generally the filters were excluded if it applied to all, but now I see those are for the CLI. I do see that generally all the platforms are listed in the category docs 👍

Jordan-Nelson (approved):

dnys1 (approved):

offlineprogrammer (approved):

PR #3605, opened 2021-09-22
Remove page from Flutter listing

dnys1:
_Issue #, if available:_
- Erroneously lists a getting started guide for Flutter which it does not support

_Description of changes:_


Jordan-Nelson (approved):
Approved. We should let someone from iOS/Android know that this is showing for iOS and Android as well.

PR #3613, opened 2021-09-23
Update node options fix in GitHub workflow

jakeburden:
_Issue #, if available:_
  

_Description of changes:_
Moves the NODE_OPTIONS env variable to it's own line


tjleing (approved):
If the last one looked good, this one looks even better!

Issue #3376, opened 2021-07-02
[Feedback]Tons of errors in Angular 12 app


rajinder-yadav:
**Page**: [`/start/getting-started/data-model/q/integration/angular`](https://docs.amplify.aws/start/getting-started/data-model/q/integration/angular)

**Feedback**:

Here are the following errors I am seeing initially, I attempted to fix some.

1. Need to move the code from onInit into the constructor.
2. Rrename file, 'aws-exports.js' to 'aws-exports.ts'
3. Assign empty array to class field `restaurants: Array<Restaurant> = [];`

```ts
Error: src/app/app.component.ts:13:10 - error TS2564: Property 'createForm' has no initializer and is not definitely assigned in the constructor.

13   public createForm: FormGroup;
            ~~~~~~~~~~


Error: src/app/app.component.ts:15:3 - error TS2564: Property 'restaurants' has no initializer and is not definitely assigned in the constructor.

15   restaurants: Array<Restaurant>;
     ~~~~~~~~~~~


Error: src/app/app.component.ts:16:3 - error TS7008: Member 'subscription' implicitly has an 'any' type.

16   subscription;
     ~~~~~~~~~~~~


Error: src/app/app.component.ts:29:7 - error TS2322: Type '({ __typename: "Restaurant"; id: string; name: string; description: string; city: string; createdAt: string; updatedAt: string; } | null)[] | null | undefined' is not assignable to type 'Restaurant[]'.
  Type 'undefined' is not assignable to type 'Restaurant[]'.

29       this.restaurants = event.items;
         ~~~~~~~~~~~~~~~~

Error: src/main.ts:5:25 - error TS7016: Could not find a declaration file for module './aws-exports'. '/home/yadav/dev/tmp/amplify-app/src/aws-exports.js' implicitly has an 'any' type.

5 import aws_exports from "./aws-exports";
                          ~~~~~~~~~~~~~~~
```

After making these changes, there is still one more error I am unable to fix at the moment.

```sh

Error: src/app/app.component.ts:26:7 - error TS2322: Type '({ __typename: "Restaurant"; id: string; name: string; description: string; city: string; createdAt: string; updatedAt: string; } | null)[]' is not assignable to type 'Restaurant[]'.
  Type '{ __typename: "Restaurant"; id: string; name: string; description: string; city: string; createdAt: string; updatedAt: string; } | null' is not assignable to type 'Restaurant'.
    Type 'null' is not assignable to type 'Restaurant'.

[...]

Issue #3140, opened 2021-04-14
[Feedback]Multiple errors in the meat of the tutorial - giving up


codeedog:
**Page**: [`/start/getting-started/nextsteps/q/integration/angular`](https://docs.amplify.aws/start/getting-started/nextsteps/q/integration/angular)

**Feedback**:

I tried following along with your tutorial, but there are too many errors and I don't feel like continuing through with figuring out how to fix them. Normally, I try to help improve the projects I'm interested in using, however, I already had to fight through a bug on the previous [page](https://docs.amplify.aws/start/getting-started/setup/q/integration/angular) which I encountered when running `amplify init` and was able to work around by using `amplify add env`, but only after stumbling upon a link in a comment in this StackOverflow [issue](https://stackoverflow.com/questions/67068703/typeerror-cannot-read-property-projectpath-of-undefined-when-create-amplify) filed by someone else.

I'm more interested in understanding how to create a backend in REST for my Angular front end. I want the backend to be NestJs compliant, which means Typescript. Some of the errors below are Angular typescript errors and are in your provided libraries. Other of the errors are code that I copied and pasted from your tutorial. It's possible that I made a mistake, but since GraphQL is not of my interest and there are many errors, I'm giving up.

Which, tbh, is a shame, because I would like to understand GQL better, and I will, down the road.

You should fix your libraries, at a minimum, so that they are typescript compliant or provide the `.d.ts` type extensions. The fact that you don't leads me to believe that the rest of this tutorial has a similar level of inattention and is not worth my time.

I apologize for being blunt. It is not intended as disparagement. Rather, it is intended to let you know that I'm new to AWS but very much not new to programming and am judging your company by the quality its product including its teaching products.

````
Error: node_modules/@aws-amplify/api-graphql/lib-esm/types/index.d.ts:1:30 - error TS7016: Could not find a declaration file for module 'graphql/error/GraphQLError'. '/Users/andrewphilips/Documents/Websites/test-aws/aws-ng-tutorial/amplify-app/node_modules/graphql/error/GraphQLError.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/graphql` if it exists or add a new declaration (.d.ts) file containing `declare module 'graphql/error/GraphQLError';`

1 import { GraphQLError } from 'graphql/error/GraphQLError';
                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~


[...]

Issue #2918, opened 2021-02-07
[Feedback]the angular code doesn't compile


jmagaram:
**Page**: [`/lib/datastore/examples/q/platform/js`](https://docs.amplify.aws/lib/datastore/examples/q/platform/js)

**Feedback**: the onQuery method is missing, "subscription" declaration generates error when no data type is provided (using TypeScript/Angular strict mode)

Come on guys, give a little love to Angular!

wlee221:
Hello @jmagaram, apologies on the late response. The subscription typing should be fixed from #3606. Specifically, we now are importing `Restaurant` type directly from `"./API.service"`, which you can use to strictly type API functions. Please let us know if you see the same issue with the latest snippets. 

And +1 on giving some love to Angular! There are a lot of Angular gaps we have seen with current UI components, and we'll work our best to bridge that gap in our new [Angular UI components](https://ui.docs.amplify.aws/ui/components/authenticator) that we're developing now 🎉

PR #3625, opened 2021-09-28
Strict mode troubleshooting

wlee221:
_Issue #, if available:_ #2918, #3140, #3376, [stackoverflow](https://stackoverflow.com/questions/69353284/integration-issue-in-angular-with-aws-amplify-for-google-sign-up/69356521)

_Description of changes:_ This adds troubleshooting section for using Typescript strict mode. This is prominent on Angular 12+ users as Angular 12 turns on strict mode by default, but this is relevant to all users who uses stricter TypeScript settings. I pulled it out to a separate troubleshooting for that reason, and we can reference this document from issues and [UI docs](https://ui.docs.amplify.aws/ui) as well.

Screen Shot 2021-09-28 at 9 32 09 AM

### DIFF
--- a/src/fragments/lib/troubleshooting/js/strict-mode.mdx
+++ b/src/fragments/lib/troubleshooting/js/strict-mode.mdx
@@ -1,0 +1,43 @@
+This section lists known TypeScript errors and their workarounds when [strict mode](/lib/troubleshooting/strict-mode/q/platform/js/) is enabled on your application.
+
+## Declaration file for GraphQL
+
+If you see the error
+
+```
+Could not find a declaration file for module 'graphql/error/GraphQLError'` 
+Could not find a declaration file for module 'graphql/language/ast'` 
+```
+
+Add a declaration file `src/types.d.ts` with the following content to resolve the error:
+
+```ts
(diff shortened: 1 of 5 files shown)


offlineprogrammer (approved):

src/fragments/lib/troubleshooting/js/strict-mode.mdx
```diff
@@ -0,0 +1,43 @@
...
+## Storage Type Error
+
+If you see the error
+
+```
+TS2344: Type 'T[U]' does not satisfy the constraint '(...args: any) => any'
+TS2344: Type 'T[api]' does not satisfy the constraint '(...args: any) => any'.
+```
+
+You can resolve this by setting [`strictNullChecks`](https://www.typescriptlang.org/tsconfig#strictNullChecks) to false within your `tsconfig.json`.
```
wlee221, line 43:
```suggestion
```
This is being addressed from https://github.com/aws-amplify/amplify-js/pull/8956, removing. Thanks @jamesaucode ✨

slaymance (approved):
LGTM

PR #3684, opened 2021-10-14
Add amplify init step

friedcode:
When following these steps, the `amplify add auth` command will return `You are not working inside a valid Amplify project.
Use 'amplify init' in the root of your app directory to initialize your project, or 'amplify pull' to pull down an existing project.` So I think the amplify init step should be here, too.

_Issue #, if available:_

_Description of changes:_
- Adding the `amplify init` command and description.


mauerbac:
Spoke with Geo PM and closing as this content has been removed

Issue #3355, opened 2021-06-29
[Feedback]Deploy and host app


rajinder-yadav:
**Page**: [`/start/getting-started/hosting/q/integration/js`](https://docs.amplify.aws/start/getting-started/hosting/q/integration/js)

**Feedback**: Would be nice to also share the command to destroy the deployment and release all the aws resources created.

A really nice amplify feature would be able to list all the aws resources created, not sure if this can be done already?

PR #3697, opened 2021-10-21
Added instruction for resource clean-up

nilisha-jais:
_Issue :_ #3355 

_Description of changes:_ Added "Resource Clean-Up" Instructions to End of Getting Started Guides 
 


renebrandel:
Hi - thank you for the PR but I think the original issue is more talking about how to delete the entire Amplify project and not just the hosted web app. I think we should reword this into its own ">" block using `amplify delete` instead of Amplify remove hosting.

PR #3724, opened 2021-11-01
fix:table cutoff

nilisha-jais:
Issue : #3677 

_Description of changes:_
The content of the table seems to cutoff [here](https://docs.amplify.aws/cli/graphql-transformer/resolvers/#custom-resolvers). 
Fixed them by adding a overflow where scrollbars only appear when the content is overflowing. The table now even looks responsive.


david-mcafee (approved):
Thanks for your PR @nilisha-jais!! Pulled these changes and this looks good! 🥇 There may be a better solution longterm for how we format these tables that we could consider, but this is a great solution to unblock the documentation as these values are currently unreadable. Thanks again!

offlineprogrammer (approved):

PR #3725, opened 2021-11-01
docs(geo): updates from DA feedback

TreTuna:
_Description of changes:_

- docs(geo): add info about map.remove
- docs(geo): add import in snippets
- docs(geo): all long, lat comments to reinforce order
- docs(geo): add geo to js configuration page
- docs(geo): add link to CLI search index provisioning

### DIFF
--- a/src/fragments/lib/geo/js/maps.mdx
+++ b/src/fragments/lib/geo/js/maps.mdx
@@ -217,11 +217,22 @@ map.resize(); // forces the map to re-render
 ## Removing a map from the DOM
 
 When it's time to remove the map from the DOM, you can use the `.remove` method of the generated map. This will clean up and release all resources associated with the map (DOM elements, event bindings, web workers, and WebGL resources).
-Note: Afterwards, you must not call any other methods on the map.
 
 ```javascript
 map.remove();
 ```
+
+<Callout warning>
+After calling `.remove()`, you must not call any other methods on the map.
+</Callout>
+
+<Callout warning>
+
+**For React users:**
+
+Not removing the map on component unmount can cause memory leaks in your application. It's recommended to call `.remove()` in either the return function of a React `useEffect` hook or the `componentWillUnmount` lifecycle hook of a class component.
+
+</Callout>
 
 ## Add map to html website
 
--- a/src/fragments/lib/geo/js/maps.mdx
+++ b/src/fragments/lib/geo/js/maps.mdx
@@ -16,9 +16,13 @@ Add the dependencies to your app:
 npm install maplibre-gl@1 maplibre-gl-js-amplify
 ```
 
-> **Note:** Make sure that `maplibre-gl-js-amplify` version `1.1.0` or above is installed.
+<Callout>
 
-> **Note:** Ensure that your package bundler (webpack, rollup, etc) is configured to handle css files. Check out the webpack documentation [here](https://webpack.js.org/loaders/css-loader/).
+Verify the following:
+- `maplibre-gl-js-amplify` version `1.1.0` or above is installed
+- Any package bundlers (webpack, rollup, etc) are configured to handle css files. Check out the webpack documentation [here](https://webpack.js.org/loaders/css-loader/).
+
+</Callout>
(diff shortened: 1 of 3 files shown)


thaddmt (approved):

src/fragments/lib/geo/js/maps.mdx
```diff
@@ -160,6 +162,15 @@ map.setStyle("myAmplifyGeoEsriTopographicMap"); // map name received from getAva
 map.resize(); // forces the map to re-render
 ```
 
+## Removing a map from the DOM
+
+When it's time to remove the map from the DOM, you can use the `.remove` method of the generated map. This will clean up and release all resources associated with the map (DOM elements, event bindings, web workers, and WebGL resources).
+Note: Afterwards, you must not call any other methods on the map.
```
siegerts, line 168: Will something destructive happen if a method is called? Or, is it just that the method will not work? Just trying to gauge if this should be bolded or called out more as a `warning`.

aspittel: Yeah, I think it's important to call out that if you don't do this in a framework like React you'll have a memory leak

aspittel:
Hey, one piece of feedback throughout is to add more explanations before code snippets rather than just adding comments to the code, often they scroll off the page. I think this is a good first step though.

sammartinez:
> Hey, one piece of feedback throughout is to add more explanations before code snippets rather than just adding comments to the code, often they scroll off the page. I think this is a good first step though.

Im in agreement with this with @aspittel. We should be give as much data as we can for our customers on this.

src/fragments/lib/geo/js/maps.mdx
TreTuna: > Will something destructive happen if a method is called? Or, is it just that the method will not work? Just trying to gauge if this should be bolded or called out more as a `warning`.

I'm actually not positive _what_ happens here, that's all the information MapLibre gives in their documentation about `remove()`, would need to do some manual messing around to know more atm.

Happy to update this more as a warning and add something specific calling out React for memory leaks.

https://maplibre.org/maplibre-gl-js-docs/api/map/#map#remove

image

TreTuna:
> Hey, one piece of feedback throughout is to add more explanations before code snippets rather than just adding comments to the code, often they scroll off the page. I think this is a good first step though.

I understand this, but I believe it's also a balance upon doing this and also staying consistent with all the other categories in the documentation. 

In this case, I'm guessing the comment is mainly in reference with the "Top level configuration" portion, which is staying consistent by using comments as all of the other configuration objects on that page.

TreTuna:
Made several updates around `.remove` and for React users:

The initial snippet to create a map now has a React section:
image
image

And the section on `.remove` now uses `<Callout warning>` for both:
- The problem of calling methods after removing the map
- Warning React users about memory leaks if they do not remove the map

src/fragments/lib/geo/js/maps.mdx
TreTuna: Added updates you can see below around this.

src/fragments/lib/geo/js/maps.mdx
```diff
@@ -160,6 +214,26 @@ map.setStyle("myAmplifyGeoEsriTopographicMap"); // map name received from getAva
...
+## Removing a map from the DOM
+
+When it's time to remove the map from the DOM, you can use the `.remove` method of the generated map. This will clean up and release all resources associated with the map (DOM elements, event bindings, web workers, and WebGL resources).
+
+```javascript
+map.remove();
+```
+
+<Callout warning>
+After calling `.remove()`, you must not call any other methods on the map.
```
siegerts, line 226:
```suggestion

After calling `.remove()`, you must not call any other methods on the map.

```

siegerts: Just so that md renders as code

siegerts (approved):
small format on the callout for mdx to recognize the code hightlight. LGTM 🗺️

PR #3728, opened 2021-11-02
remove Analytics.record() from code snippet

chrisbonifacio:
_Issue #, if available:_
https://github.com/aws-amplify/amplify-js/issues/9138

_Description of changes:_
Removing the line `Amplify.record()` from the example code snippet because it can cause confusion with the usage of the `record` method by making it seem like it can be called without arguments when it [expects either a string or object ](https://github.com/aws-amplify/amplify-js/blob/49da354140141a0dc88d6a3922ce79728857630e/packages/analytics/src/Analytics.ts#L232) as the first argument.

It will throw an error if called as is in the code snippet.


sammartinez (approved):
LGTM 🌮

offlineprogrammer (approved):

Issue #3737, opened 2021-11-05
[Feedback]FEEDBACK_TITLE_HERE


mbreslow:
**Page**: [`/guides/location-service/setting-up-your-app/q/platform/[platform]`](https://docs.amplify.aws/guides/location-service/setting-up-your-app/q/platform/js/#creating-a-new-react-app)

**Feedback**:

There seems to be an `amplify init` step missing ahead of `amplify add auth`

siegerts:
Hi @mbreslow, thanks for opening.  The `init` command has been updated. Also, the referenced page in the Geo docs explains that the CLI will prompt for Auth configuration if needed during the `amplify add geo` configuration.

https://docs.amplify.aws/cli/geo/maps/

PR #3740, opened 2021-11-07
Added the missing step

nilisha-jais:
Issue : #3737 

_Description of changes:_
Added the missing step of  `amplify init` 

### DIFF
--- a/src/fragments/guides/location-service/setting-up-your-app-js.mdx
+++ b/src/fragments/guides/location-service/setting-up-your-app-js.mdx
@@ -47,7 +47,7 @@ In the following procedure, you’ll add authentication to your React app using 
 
 **To add authentication to your React app**
 
-1. To do so, we'll initialize an Amplify project and add authentication. From your projects directory, run the following command, and answer the prompted questions as indicated.
+1. Initialize an Amplify project and add authentication. From your project's directory, run the following commands, and answer the prompted questions as indicated.
 
 ```bash
 amplify init
--- a/src/fragments/guides/location-service/setting-up-your-app-js.mdx
+++ b/src/fragments/guides/location-service/setting-up-your-app-js.mdx
@@ -47,9 +47,13 @@ In the following procedure, you’ll add authentication to your React app using 
 
 **To add authentication to your React app**
 
-1. From your projects directory, run the following command, and answer the prompted questions as indicated.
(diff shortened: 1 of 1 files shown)


siegerts (approved):

src/fragments/guides/location-service/setting-up-your-app-js.mdx
```diff
@@ -47,9 +47,13 @@ In the following procedure, you’ll add authentication to your React app using
 
 **To add authentication to your React app**
 
-1. From your projects directory, run the following command, and answer the prompted questions as indicated.
+1. To do so, we'll initialize an Amplify project and add authentication. From your projects directory, run the following command, and answer the prompted questions as indicated.
```
aspittel, line 50: projects should be possessive here - 'project's'

aspittel: I'd cut "To do so, we'll " and just go with "Initialize an Amplify project and add authentication"

src/fragments/guides/location-service/setting-up-your-app-js.mdx
```diff
@@ -47,9 +47,13 @@ In the following procedure, you’ll add authentication to your React app using
...
 
 ```bash
+amplify init
+# Follow the steps to give the project a name, environment name, and set the default text editor.
```
aspittel, line 54: I would show each prompt and the recommended answer here

aspittel (changes requested):

ashika01 (approved):

src/fragments/guides/location-service/setting-up-your-app-js.mdx
ashika01: @nilisha-jais could we have this added or can you speak with @aspittel to see if we could go without this. I am okay without since the config for init might change based on user's config.

aspittel: Some guides have `init` some don't - most of the good getting started guides have a walk through of `init` ex: https://docs.amplify.aws/lib/datastore/getting-started/q/platform/js/#prerequisites

If you don't want to do this, I would instead have "make sure to have amplify initialized in your project" rather than listing it as a step without more explanation. Or link to further docs on amplify init.

siegerts:
Closing since this information has been restructured/updated - https://github.com/aws-amplify/docs/issues/3737#issuecomment-1067091604

Thanks for the PR, @nilisha-jais. Apologies on the delay.

PR #3784, opened 2021-11-18
Multi-region copy

jakeburden:
_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/src/plugins/code-block.tsx
+++ b/src/plugins/code-block.tsx
@@ -46,7 +46,9 @@ const supportedLanguages = [
 
 loadLanguages(supportedLanguages);
 
-const highlight = (code, language) => {
+let lineCountOffSet = 0;
+
+const highlight = (code, language, regions, idx) => {
   language = language.replace('language-', '');
   let highlighted = '';
   const languageIsSet = !!(language && language.trim().length > 0);
@@ -68,7 +70,13 @@ const highlight = (code, language) => {
     languageIsSet ? `-${language}` : ''
   }">${highlighted}</div>`;
 
-  const lineCount = html.split(/\r\n|\r|\n/).length;
+  let lineCount = html.split(/\r\n|\r|\n/).length;
+
+  if (regions > 1 && idx > 0) {
+    lineCountOffSet = lineCountOffSet + lineCount;
+  } else {
+    lineCountOffSet = 0;
+  }
 
   // const noCopy = code.indexOf('###BEGIN_COPY###') > -1;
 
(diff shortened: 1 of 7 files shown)


offlineprogrammer (approved):

jakeburden:
https://user-images.githubusercontent.com/2081699/142953591-b8cfdfe4-2c12-4e2f-a377-2bc26c64faca.mov

src/plugins/code-block.tsx
```diff
@@ -65,61 +67,82 @@ const highlight = (code, language) => {
...
-    versions.ANDROID_KOTLIN_VERSION,
+    versions.ANDROID_KOTLIN_VERSION
   );
   code = code.replace(/ANDROID_GEO_VERSION/g, versions.ANDROID_GEO_VERSION);
+
+  return code;
+};
+
+const removeCopy = (code) => {
+  code = code.split('###BEGIN_COPY###');
```
tjleing, line 123: so no support for `###END_COPY###`?  not quite sure the best way to add that to this implementation, but I'd imagine you could first split by `###BEGIN_COPY###`, and then loop over the list and see which pieces have `###END_COPY###`...  and also maintain some data keeping track of which array segments ended with an `###END_COPY###`, so you can pass it to each `CodeBlock`.

src/plugins/code-block.tsx
```diff
@@ -65,61 +67,82 @@ const highlight = (code, language) => {
...
   return code;
 };
 
+const reducer = (previousValue, currentValue) =>
```
tjleing, line 127: personally I think I'd define this within the `visit` function, just because that's the only place it's used.  up to you though!

src/plugins/code-block.tsx
```diff
@@ -65,61 +67,82 @@ const highlight = (code, language) => {
...
   // eslint-disable-next-line @typescript-eslint/no-var-requires
-  const visit = require("unist-util-visit");
+  const visit = require('unist-util-visit');
 
   visit(tree, (node) => {
-    if (node.tagName === "code") {
-      const code = addVersions(node.children[0].value);
+    if (node.tagName === 'code') {
+      let code = addVersions(node.children[0].value);
+      code = removeCopy(code);
```
tjleing, line 137:
```suggestion
      const code = addVersions(node.children[0].value);
      const codeSections = removeCopy(code);
```
naming is hard, but I personally prefer this?  (github didn't let me select a few lines past this, but you'd obviously need to change `node.children = code` to `node.children = codeSections` too)

tjleing (commented):
This is a great feature, I hope you had fun with `unist`!  A few nits -- also, could you add a description of how to use these code blocks to the README?  I'm guessing that this isn't on a staging branch because there's no mdx file that uses it yet.

PR #3839, opened 2021-11-30
[Android] Add CLI steps for geo

raphkim:
Addressed issues that caused minor friction while following the documentation for geo setup guide.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -909,6 +909,7 @@
     "PinpointTargeting",
     "pipenv",
     "pipenv's",
+    "placeindex",
     "pluggable",
     "png",
     "Podfile",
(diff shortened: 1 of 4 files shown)


offlineprogrammer (approved):

drochetti:
I am confused on why the CLI instructions are in place only for Android and not the other platforms. The setup process is the same for all platforms we support, why only the Android docs will have it? Any thoughts @hdworld11 @TreTuna @ameter

raphkim:
i can also move the CLI instruction to native-common so that both iOS and Android users will get the same instructions. For now, I only addressed it in Android because list of frictions presented to me were specifically for the Android docs.

hdworld11 (approved):

PR #3875, opened 2021-12-15
fix(flutter): handles setting platform and framework to flutter

haverchuck:
_Description of changes:_
Docs site was linking the user to the js getting started section from the flutter UI components page. The cause was that we weren't handling flutter cases in the `withFilterOverrides` function. This PR attempts to solve that issue. The overall  function could possibly use some refactoring in order to scale with more frameworks, but I think this solves the immediate issue.


offlineprogrammer (approved):

Jordan-Nelson (approved):
Tested briefly locally. lgtm

tjleing (approved):
Definitely would like a refactor after this PR lands

jakeburden (approved):

PR #3910, opened 2022-01-05
Updated custom resolver info

Rafcin:
Akshbhu recommended using cdk.FN.ref in https://github.com/aws-amplify/amplify-cli/issues/9391
This is the right approach and should be updated.

Migration info for users using going from ES to OS was added as well, the data source is renamed automatically when migrating and resolvers need to be updated as such.

_Issue #, if available:_
https://github.com/aws-amplify/amplify-cli/issues/9424

_Description of changes:_
Updated resolver template to use cdk.FN.ref and added note about going from Elastic to OpenSearch

### DIFF
--- a/src/pages/cli/graphql/custom-business-logic.mdx
+++ b/src/pages/cli/graphql/custom-business-logic.mdx
@@ -526,7 +526,11 @@ export class cdkStack extends cdk.Stack {
     );
 
     const resolver = new appsync.CfnResolver(this, "custom-resolver", {
-      apiId: retVal.api.new.GraphQLAPIIdOutput,
+      // apiId: retVal.api.new.GraphQLAPIIdOutput,
+      // https://github.com/aws-amplify/amplify-cli/issues/9391#event-5843293887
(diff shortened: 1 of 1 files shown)


src/pages/cli/graphql/custom-business-logic.mdx
```diff
@@ -526,7 +526,11 @@ export class cdkStack extends cdk.Stack {
     );
 
     const resolver = new appsync.CfnResolver(this, "custom-resolver", {
-      apiId: retVal.api.new.GraphQLAPIIdOutput,
+      // apiId: retVal.api.new.GraphQLAPIIdOutput,
+      // https://github.com/aws-amplify/amplify-cli/issues/9391#event-5843293887
+      // If you use Amplify you can access the paramater via Ref since it's a CDK paramater passed from the root stack.
```
siegerts, line 531:
```suggestion
      // If you use Amplify you can access the parameter via Ref since it's a CDK parameter passed from the root stack.
```

edwardfoyle (approved):

Rafcin:
Sorry didn't see the email for this. I'll change in a bit!

siegerts (approved):

PR #3934, opened 2022-01-13
auth/delete_user

haverchuck:
**DO NOT MERGE BEFORE FEATURE IS LAUNCHED**

_Description of changes:_
Adds deleteUser docs for flutter (and creates skeleton for other platforms)


offlineprogrammer (approved):

src/fragments/lib/auth/native_common/delete_user/common.mdx
```diff
@@ -0,0 +1,15 @@
+Invoke the `deleteUser` API to delete a user from the Auth category. This action will also sign your user out.
+
+If your application is using the default Amazon Cognito plugin, this action will delete the user from the Amazon Cognito User Pool. It will have no effect if you are federating with a Cognito Identity Pool alone.
```
palpatim, line 3: What other plugin might a customer be using? It would more directly address the issue by saying something like "If your application is using the default configuration for the Amazon Cognito plugin..." Which is I think what you're trying to get to. That is, the default config sets up a user pool & identity pool, but the customer can override to just use federated identities, in which case the deleteUser API has no effect. Is that about what you're trying to say?

PR #3938, opened 2022-01-14
Update: add details to GraphQL docs

marcvberg:
_Issue #, if available:_

_Description of changes:_
Adding in some additional details for GraphQL documentation


src/pages/cli/graphql/search-and-result-aggregations.mdx
```diff
@@ -160,11 +160,13 @@ query MyQuery {
 }
 ```
 
-In the example above, the response's `total` field contains the total search result count for "Students" whose name ends with "Brandel". 
+In the example above, the response's `total` field contains the total search result count for "Students" whose name ends with "Brandel".
+This total will be calculated based on all records, in the case of pagination even if not all records can be returned at once the total will still be representative of all matching results.
```
renebrandel, line 164: do we need to explicitly call this out? Isn't this the intuitive answer? I recognize we didn't do this behavior in the past, therefore needed a note. Maybe more concise:

```suggestion
Note: `total` is calculated based on all records, irrespective of pagination configurations.
```

src/pages/cli/graphql/search-and-result-aggregations.mdx
```diff
@@ -160,11 +160,13 @@ query MyQuery {
...
 
 ## Aggregate values for search result (minimum, maximum, average, sum, terms)
 
-Use the `aggregates` parameter to get aggregate values such as "minimum", "maximum", "average", and "sum" returned in the `aggregateItems` field. 
+Use the `aggregates` parameter to get aggregate values such as "minimum", "maximum", "average", and "sum" returned in the `aggregateItems` field.
+The same as [total](#total-count-of-search-results), these aggregates will be calculated based on all matching records, aggregate results are not based on pagination
```
renebrandel, line 169: Same nit as above.

src/pages/cli/graphql/data-modeling.mdx
```diff
@@ -500,7 +503,7 @@ The `@model` directive will generate:
 - Filter input objects that allow you to filter objects in list queries and relationship fields.
 - For list queries the default number of objects returned is 100. You can override this behavior by setting the limit argument.
 
-Type definition of the `@model` directive:
+### Type definition of the `@model` directive
```
renebrandel, line 506: Suggestions: Elevate all pages' "How it works" into H2s and then make he definitions H3s

src/pages/cli/graphql/data-modeling.mdx
```diff
@@ -460,6 +461,8 @@ type Todo @model(queries: { get: "queryFor" }, mutations: null, subscriptions: n
 
 In the example above, you'll be able to run a `queryForTodo` query to get a single Todo element. Mutations (meaning create, update, and delete) as well as subscriptions are disabled for the Todo model above.
 
+The types for queries, mutations, and subscriptions are [found here](#type-definition-of-the-`@model`-directive) at the bottom of the page
```
renebrandel, line 464:
```suggestion
For the type definitions of queries, mutations, and subscriptions, see [Type Definitions of the `@model` Directive](#type-definition-of-the-`@model`-directive).
```

src/pages/cli/graphql/data-modeling.mdx
```diff
@@ -445,6 +445,7 @@ type Todo @model {
 ```
 
 If you create a new Todo and don't supply a `content` input, Amplify will ensure that `My new Todo` is auto populated as a value. 
+Because the `@default` directive provides a value when none is provided on creation, a non-null assertion (i.e. `String!`) is disregarded by the transformer.
```
renebrandel, line 448:
```suggestion
When `@default` is applied, non-null assertions using `!` are disregarded. For example, `String!` is treated the same as `String`.
```

src/pages/cli/graphql/authorization-rules.mdx
```diff
@@ -260,7 +260,7 @@ In the example above:
 
 ## Field-level authorization rules
 
-When an authorization rule is added to a field, it'll strictly define the authorization rules applied on the field. Field-level authorization rules **do not** inherit model-level authorization rules. Meaning, only the specified rule is always applied. 
+When an authorization rule is added to a field, it'll strictly define the authorization rules applied on the field. Field-level authorization rules **do not** inherit model-level authorization rules. Meaning, only the specified field-level auth rule is applied.
```
renebrandel, line 263:
```suggestion
When an authorization rule is added to a field, it'll strictly define the authorization rules applied on the field. Field-level authorization rules **do not** inherit model-level authorization rules. Meaning, only the specified field-level authorization rule is applied. 
```

src/pages/cli/graphql/client-code-generation.mdx
```diff
@@ -21,6 +21,8 @@ You’ll see questions as before, but now it will also automatically ask you if
 
 When you deploy your GraphQL API to the cloud, you are prompted to configure codegen. When a project is configured to generate code with codegen, it stores all the configuration `.graphqlconfig.yml` file in the root folder of your project.
 
+Models will not be automatically generated by codegen on an `amplify push`.
```
renebrandel, line 24:
```suggestion
Models are not automatically generated by codegen on an `amplify push`.
```

src/pages/cli/graphql/custom-business-logic.mdx
```diff
@@ -28,6 +28,9 @@ type Query {
     - [HTTP resolver](#http-resolver): call an HTTP endpoint upon a query or mutation
     - [VTL resolver](#vtl-resolver) (most advanced): use VTL mapping templates to customize the query and mutation logic
 
+3. Optionally, secure your custom query or mutation with [field-level auth](/cli/graphql/authorization-rules/)]
```
renebrandel, line 31: Given deny-by-default:
```suggestion
3. Secure your custom query or mutation with [field-level authorization rules](/cli/graphql/authorization-rules/)]
```

src/pages/cli/graphql/custom-business-logic.mdx
```diff
@@ -28,6 +28,9 @@ type Query {
...
+    - `@auth`on a custom query or mutation only supports static groups, iam, and apiKey. Owner and dynamic groups are currently mapped to the model generated operations since those dynamic fields are handled directly in the resolvers.
```
renebrandel, line 32:
```suggestion
    - Note: Only API Key, IAM, and static group authorization rules are supported on a custom query or mutation.
```

src/pages/cli/graphql/search-and-result-aggregations.mdx
marcvberg: It can be intuitive for some, but for someone without experience I could easily see an argument being made for either. I could imagine someone thinking their result is for the records they have in hand, and then using it as something it isn't

renebrandel (approved):

src/pages/cli/graphql/client-code-generation.mdx
```diff
@@ -21,6 +21,8 @@ You’ll see questions as before, but now it will also automatically ask you if
 
 When you deploy your GraphQL API to the cloud, you are prompted to configure codegen. When a project is configured to generate code with codegen, it stores all the configuration `.graphqlconfig.yml` file in the root folder of your project.
```
undefobj, line 22: We should explicitly state that customers should not modify this file if we're calling it out.

src/pages/cli/graphql/client-code-generation.mdx
```diff
@@ -21,6 +21,8 @@ You’ll see questions as before, but now it will also automatically ask you if
...
 
+Models are not automatically generated by codegen on an `amplify push`.
```
undefobj, line 24: Why aren't we triggering modelgen? If it's necessary for their app and they've configured it upon prompt then we should do it automatically.

src/pages/cli/graphql/custom-business-logic.mdx
```diff
@@ -28,6 +28,9 @@ type Query {
     - [HTTP resolver](#http-resolver): call an HTTP endpoint upon a query or mutation
     - [VTL resolver](#vtl-resolver) (most advanced): use VTL mapping templates to customize the query and mutation logic
 
+3. Secure your custom query or mutation with [field-level authorization rules](/cli/graphql/authorization-rules/)]
+    - Note: Only API Key, IAM, custom, and static group authorization rules are supported on a custom query or mutation.
```
undefobj, line 32: So what is not supported? Dynamic auth? If that's the known thing that is unsupported it would be better to explicitly call that out.

src/pages/cli/graphql/client-code-generation.mdx
marcvberg: It doesn't seem like anyone has clear information on this after investigating, so I'm removing this and there will be a follow up discussion/change

src/pages/cli/graphql/custom-business-logic.mdx
marcvberg: Updated

src/pages/cli/graphql/client-code-generation.mdx
marcvberg: Updated

undefobj (approved):

PR #3974, opened 2022-01-27
add docs for flutter authenticator custom ui, unauthenticated routes

Jordan-Nelson:
_Issue #, if available:_

_Description of changes:_
- add docs for flutter authenticator custom ui, unauthenticated routes


dnys1 (approved):
Looks good, thanks!

src/fragments/ui/auth/flutter/authenticator.mdx
```diff
@@ -159,6 +181,110 @@ return Authenticator(
...
+
+This is done by providing a builder to the Authenticator's `authenticatorBuilder` value. If you are familiar with widgets like `StreamBuilder`, `FutureBuilder`, or `LayoutBuilder`, this may feel familiar. `authenticatorBuilder` is a builder method that takes arguments for the current `BuildContext` and `AuthenticatorState`, which is a representation of the state being managed by the authenticator.
+
+Below is an example of how you can use `authenticatorBuilder` to add a logo to both the sign up and sign in screen, as well as change the layout to remove the default tab structure.
+
+This example uses the `SignInForm()` and `SignUpForm()` widgets from the amplify_authenticator package. In addition to the available prebuilt forms, form fields (ex: `SignInFormField.username()` or `SignUpFormField.email()`) and form buttons (ex: `SignInButton()`) are also available. All of the prebuilt widgets are fully integrated into the authenticator's state.
```
haverchuck, line 192: Do we need both of these examples? It makes the code sample kind of verbose.

haverchuck (approved):
I think the customization code snipped could be cut down a bit for brevity, but other than that LGTM.

src/fragments/ui/auth/flutter/authenticator.mdx
Jordan-Nelson: I could change it to the following. I wanted to include a could examples since there is no list of all the available components (which I think we should include in the docs before GA, but I did think was needed now).

```suggestion
This example uses the `SignInForm()` and `SignUpForm()` widgets from the amplify_authenticator package. In addition to the available prebuilt forms, form fields (ex: `SignInFormField.username()`) and form buttons (ex: `SignInButton()`) are also available. All of the prebuilt widgets are fully integrated into the authenticator's state. 
```

offlineprogrammer (approved):

PR #3981, opened 2022-01-31
chore(getting started docs): update to accurate filename

evcodes:
I believe the latest xcode/swift ui filename is `ContentView.swift`

_Description of changes:_

### DIFF
--- a/src/fragments/start/getting-started/ios/integrate.mdx
+++ b/src/fragments/start/getting-started/ios/integrate.mdx
@@ -4,7 +4,7 @@ Next you'll use the generated model to create, update, query, and delete data. I
 
 First, we'll add the DataStore plugin and configure Amplify.
 
-1. Open the main file of the application - `AppDelegate.swift` or `TodoApp.swift` depending on the Life Cycle previously chosen - and **add the following** import statements at the top of the file:
+1. Open the main file of the application - `ContentView.swift` or `TodoApp.swift` depending on the Life Cycle previously chosen - and **add the following** import statements at the top of the file:
   <BlockSwitcher>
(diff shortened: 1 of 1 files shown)


offlineprogrammer (approved):

src/fragments/start/getting-started/ios/integrate.mdx
```diff
@@ -4,7 +4,7 @@ Next you'll use the generated model to create, update, query, and delete data. I
 
 First, we'll add the DataStore plugin and configure Amplify.
 
-1. Open the main file of the application - `AppDelegate.swift` or `TodoApp.swift` depending on the Life Cycle previously chosen - and **add the following** import statements at the top of the file:
+1. Open the main file of the application - `ContentView.swift` or `TodoApp.swift` depending on the Life Cycle previously chosen - and **add the following** import statements at the top of the file:
```
atierian, line 7: It's correct that the generated main view file is `ContentView.swift`, however this step refers to the app's lifecycle management file. In Xcode 12, you could choose the desired lifecycle, which would generate either `<ProjectName>App.swift` or `AppDelegate.swift` depending on the choice. However, as of Xcode 13 that option no longer exists. Considering the screen shots in the Getting Started guide show Xcode 13, we should drop the `AppDelegate.swift` option all together.

```suggestion
1. Open the main file of the application - `TodoApp.swift` and **add the following** import statements at the top of the file:
```

evcodes: I think that this is the right thing to do.

atierian (approved):

PR #3994, opened 2022-02-02
Adding tests for docs react components

timngyn:
_Description of changes:_ 

- Added `@testing-library/react` to test react components. 
- Added tests for `Block`,`BlockSwicher`, and `Callout`

More info on testing-library/react: https://testing-library.com/docs/react-testing-library/intro/

To test:
Run `yarn test`

### DIFF
--- a/src/components/Callout/__tests__/styles.test.tsx
+++ b/src/components/Callout/__tests__/styles.test.tsx
@@ -1,0 +1,30 @@
+import Callout from '../styles';
+import { render, screen } from '@testing-library/react';
+
+describe('Callout', () => {
+  test('renders info style callout when no warning prop is passed', () => {
+    const {container} = render(<Callout>Text for callout</Callout>);
+
+    const callout = container.children[0];
+
+    expect(callout).toHaveStyle(`
+      display: block;
+      padding-left: 0.75rem;
+      margin-bottom: 1rem;
+      background-color: var(--color-ink-hv);
+    `);
+  });
+
+  test('renders warning style callout when no warning prop is passed', () => {
+    const {container} = render(<Callout warning><div>Warning text for callout</div></Callout>);
+
+    const callout = container.children[0];
+
+    expect(callout).toHaveStyle(`
+      display: block;
+      padding-left: 0.75rem;
+      margin-bottom: 1rem;
+      background-color: var(--color-orange-hv);
+    `);
+  });
+});
--- a/src/components/Callout/__tests__/styles.test.tsx
+++ b/src/components/Callout/__tests__/styles.test.tsx
@@ -1,9 +1,24 @@
 import Callout from '../styles';
 import { render, screen } from '@testing-library/react';
+import createCache from '@emotion/cache';
+import { CacheProvider } from '@emotion/react';
+
+function renderWithProvider(ui) {
+  const myCache = createCache({ key: 'my-cache-key' });
+  myCache.compat = true;
+
+  const Wrapper = ({ children }) => (
+    <CacheProvider value={myCache}>{children}</CacheProvider>
+  );
+
+  return render(ui, { wrapper: Wrapper });
+}
 
 describe('Callout', () => {
   test('renders info style callout when no warning prop is passed', () => {
-    const {container} = render(<Callout>Text for callout</Callout>);
+    const { container } = renderWithProvider(
+      <Callout>Text for callout</Callout>
+    );
 
     const callout = container.children[0];
 
@@ -16,7 +31,11 @@ describe('Callout', () => {
   });
 
   test('renders warning style callout when no warning prop is passed', () => {
-    const {container} = render(<Callout warning><div>Warning text for callout</div></Callout>);
+    const { container } = renderWithProvider(
+      <Callout warning>
+        <div>Warning text for callout</div>
+      </Callout>
+    );
 
     const callout = container.children[0];
 
--- a/src/components/Callout/styles.tsx
+++ b/src/components/Callout/styles.tsx
@@ -18,7 +18,7 @@ export const CalloutOuter = styled.div<CalloutProps>((props) => {
     margin-bottom: 0;
     margin-top: 1rem;
 
-    &:first-child {
+    &:first-of-type {
       margin-top: 0;
     }
   }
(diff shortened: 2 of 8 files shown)


src/components/CodeBlockProvider/index.tsx
```diff
@@ -79,5 +79,10 @@ export default function CodeBlockProvider({children}) {
 
 export function useCodeBlockContext() {
   const context = useContext(CodeBlockContext);
+
+  if (!context) {
```
timngyn, line 83: When writing tests for this component, I ran into some issues getting the content of the Block to show up. Didn't realize that I needed to use it under the `CodeBlockProvider` and so I thought adding a guard here would be helpful for developers.

src/components/Callout/styles.tsx
```diff
@@ -18,7 +18,7 @@ export const CalloutOuter = styled.div<CalloutProps>((props) => {
     margin-bottom: 0;
     margin-top: 1rem;
 
-    &:first-child {
+    &:first-of-type {
```
timngyn, line 21: Tests were complaining about using `:first-child` when using server side rendering and suggested to use `:first-of-type` instead.

tjleing:
Just FYI, the GitHub build is failing because you're using a more recent build of one of the packages (`prism` from 1.2.3 to 1.2.6) based off of your `yarn.lock`.  I looked into it a while ago and couldn't remember if this was fixable by just changing the code block to use `text` or `markdown` or something similar, or if it meant that the package was just impossible to upgrade.

src/components/Callout/styles.tsx
tjleing: Theoretically, an HTML tag inside a `<Callout>` could be affected by this, although the only real consequence I see here is imagining having headings inside.  I don't really understand why `:first-of-type` is preferred, since I'm imagining that the issue with server-side rendering is that the order of elements in the DOM isn't guaranteed, right?  So in theory the two would have the same issue...  But if it makes the error go away, I'm tentatively in favor.

timngyn:
> Just FYI, the GitHub build is failing because you're using a more recent build of one of the packages (`prism` from 1.2.3 to 1.2.6) based off of your `yarn.lock`. I looked into it a while ago and couldn't remember if this was fixable by just changing the code block to use `text` or `markdown` or something similar, or if it meant that the package was just impossible to upgrade.

Thanks for the info Thomas. From your past research, how did you know the issue was with prism? I saw the error was coming from the `sync.mdx` file but didn't see anything obvious

tjleing:
> > Just FYI, the GitHub build is failing because you're using a more recent build of one of the packages (`prism` from 1.2.3 to 1.2.6) based off of your `yarn.lock`. I looked into it a while ago and couldn't remember if this was fixable by just changing the code block to use `text` or `markdown` or something similar, or if it meant that the package was just impossible to upgrade.
> 
> Thanks for the info Thomas. From your past research, how did you know the issue was with prism? I saw the error was coming from the `sync.mdx` file but didn't see anything obvious

If you search the codebase for "Contact Amplify JS team to request support" you'll see it comes from the syntax highlighting plugin -- there's a list of supported languages that work with `prism`, and it turns out that one of the recent updates deprecated that plain text option.

src/components/Callout/styles.tsx
timngyn: Yes, I think the error is coming from `emotion`. From what I've read it looks like they are injecting the actual `<style>` tags as the first child which breaks `:first-child`. More info from here: https://github.com/emotion-js/emotion/issues/1178

Still looking into what to do about it though. I don't think there were any warnings for this before, but I'm only getting it during the tests so it could be related to Jest

timngyn: Adding @jakeburden too

It looks like there are some options to get around this issue here: https://github.com/emotion-js/emotion/issues/1105#issuecomment-557726922

I've tried option 1 and option 2 and they make the tests pass while still keeping the css for the component the same.   

When navigating to a page with a `Callout` on my localhost, I saw that the same error regarding the `:first-child` issue is also printed in the console, so it looks like this might've been an issue for a while but hasn't broken anything I think. Have either of you seen this before?

Regarding option 2, I was able to make the changes to tests and it looks like this:
```typescript
import Callout from '../styles';
import { render, screen } from '@testing-library/react';
import createCache from '@emotion/cache';
import { CacheProvider } from '@emotion/react';

function renderWithProvider(ui) {
  const myCache = createCache({ key: 'my-cache-key' });
  myCache.compat = true;

  const Wrapper = ({ children }) => (
    <CacheProvider value={myCache}>{children}</CacheProvider>
  );

  return render(ui, { wrapper: Wrapper });
}

describe('Callout', () => {
  test('renders info style callout when no warning prop is passed', () => {
    const { container } = renderWithProvider(
      <Callout>Text for callout</Callout>
    );

    const callout = container.children[0];

    expect(callout).toHaveStyle(`
      display: block;
      padding-left: 0.75rem;
      margin-bottom: 1rem;
      background-color: var(--color-ink-hv);
    `);
  });

  test('renders warning style callout when no warning prop is passed', () => {
    const { container } = renderWithProvider(
      <Callout warning>
        <div>Warning text for callout</div>
      </Callout>
    );

    const callout = container.children[0];

    expect(callout).toHaveStyle(`
      display: block;
      padding-left: 0.75rem;
      margin-bottom: 1rem;
      background-color: var(--color-orange-hv);
    `);
  });
});
```
From my understanding of option 2, the `CacheProvider` and the `compat` flag store the styles in a cache and so they aren't inserted into the style elements. That's why the issue with `:first-child` isn't thrown anymore. (source: https://github.com/emotion-js/emotion/issues/2049#issuecomment-713868664)

src/components/Callout/__tests__/styles.test.tsx
```diff
@@ -1,9 +1,24 @@
 import Callout from '../styles';
 import { render, screen } from '@testing-library/react';
+import createCache from '@emotion/cache';
+import { CacheProvider } from '@emotion/react';
+
+function renderWithProvider(ui) {
```
timngyn, line 6: Wrapping the components with `CacheProvider` allows us to keep using `first-child`.

mauerbac:
Is this still in-progress?

timngyn:
Hi @mauerbac, this PR is currently on pause for now while I'm working on the feedback component.

timngyn:
Closing in favor of #4788

PR #3997, opened 2022-02-02
Update Amplify getting started example to use ui-react v2

joebuono:
_Issue #, if available:_

_Description of changes:_

Addressing this ticket: https://github.com/aws-amplify/amplify-ui/issues/1105

Currently the Amplify Docs site uses ui-react@v1.x.x. This PR updates the docs to use ui-react@v2 (without including any formatting changes). 

I confirmed that the getting started flow still works, and also updated a few other pages across the site.


src/fragments/lib/auth/js/getting-started.mdx
```diff
@@ -250,4 +250,4 @@ To implement authentication flows using Amplify you can either use the Amplify U
 
 `Auth` has over 30 methods including [`signUp`](/lib/auth/emailpassword#sign-up), [`signIn`](/lib/auth/emailpassword#sign-in), [`forgotPassword`](/lib/auth/manageusers#forgot-password), and [`signOut`](/lib/auth/emailpassword#sign-out) that allow you full control over all aspects of the user authentication flow.
 
-Check out the complete API [here](https://aws-amplify.github.io/amplify-js/api/classes/authclass.html).
+Check out the complete API and docs [here](https://ui.docs.amplify.aws/components/authenticator).
```
reesscot, line 253: This change isn't needed, as "Option 2" is where customers don't use the UI component, but wire up the Auth class themselves.

src/fragments/start/getting-started/react/auth.mdx
```diff
@@ -32,16 +32,17 @@ Now that we have our authentication service deployed to AWS, it's time to add au
 
 Open __src/App.js__ and make the following changes:
 
-1. Import the `withAuthenticator` component:
+1. Import the `withAuthenticator` component and styles:
```
reesscot, line 35: I'd prefer to update this to encourage use of the `Authenticator` component instead. It should match up with this page:
https://ui.docs.amplify.aws/components/authenticator

src/fragments/start/getting-started/react/auth.mdx
```diff
@@ -32,16 +32,17 @@ Now that we have our authentication service deployed to AWS, it's time to add au
...
 
 ```javascript
-import { withAuthenticator } from '@aws-amplify/ui-react'
+import { withAuthenticator } from '@aws-amplify/ui-react';
```
reesscot, line 38:
```suggestion
import { Authenticator } from '@aws-amplify/ui-react';
```

src/fragments/start/getting-started/react/auth.mdx
```diff
@@ -58,6 +59,6 @@ You can also customize this component to add or remove fields, update styling, o
 
 In addition to the `withAuthenticator` you can build custom authentication flows using the `Auth` class.
 
-`Auth` has over 30 methods including `signUp`, `signIn`, `forgotPassword`, and `signOut` that allow you full control over all aspects of the user authentication flow. Check out the complete API [here](https://aws-amplify.github.io/amplify-js/api/classes/authclass.html)
+`Auth` has over 30 methods including `signUp`, `signIn`, `forgotPassword`, and `signOut` that allow you full control over all aspects of the user authentication flow. Check out the complete API and docs [here](https://ui.docs.amplify.aws/components/authenticator)
```
reesscot, line 62: Depending on the answer to the above, this section made need to be removed entirely

src/fragments/start/getting-started/react/auth.mdx
```diff
@@ -58,6 +59,6 @@ You can also customize this component to add or remove fields, update styling, o
 
 In addition to the `withAuthenticator` you can build custom authentication flows using the `Auth` class.
```
reesscot, line 60: I don't think this is accurate, I don't think we expose the Auth class to allow this. Please check with @ErikCH and @wlee221

src/fragments/lib/auth/js/customui.mdx
```diff
@@ -1,5 +1,19 @@
 ### Customize UI theme
 
+<Callout warning={true}>
```
reesscot, line 3: Should we put this in a separate fragment like https://github.com/aws-amplify/docs/blob/2e9afd481cd22f2c22bade93b61e850c1c18f592/src/fragments/ui/auth/developer-preview-callout-js.mdx

reesscot (commented):
Looking good, just need some changes to the default recommended use of Authenticator

src/fragments/start/getting-started/react/auth.mdx
ErikCH: @joebuono I believe you mean the `Auth` js library class? We don't expose this, but the JS library does. Maybe you can make that more explicit? Also the link will need to be updated in the next paragraph to here -> https://aws-amplify.github.io/amplify-js/api/classes/authclass.html

src/fragments/start/getting-started/react/auth.mdx
wlee221: Yep, even if we keep it, this should still link to the JS API docs that was there previously.

IMO, since this is a comprehensive intro to Amplify JS *and* UI, we can keep it here for those interested in more fine-tuned Auth.

src/fragments/start/getting-started/react/auth.mdx
ErikCH: I would recommend also clarifying that they can use the Auth JS library, however it would be in replacement of the authenticator, except in the case of using overide function calls.

src/fragments/start/getting-started/react/auth.mdx
```diff
@@ -32,16 +32,44 @@ Now that we have our authentication service deployed to AWS, it's time to add au
...
+        todos.map((todo, index) => (
+          <div key={todo.id ? todo.id : index} style={styles.todo}>
+            <p style={styles.todoName}>{todo.name}</p>
+            <p style={styles.todoDescription}>{todo.description}</p>
+          </div>
+        ))
+      }
+    </div>
+  </Authenticator>
+);
```
ErikCH, line 72: I was testing this out, and I was receiving an error after logging in.
You have to wrap all children in a function, otherwise it will give you an error.

So something like this

```
 {({signOut}) => (
          <>
            <!-- Logged in Code goes here -->
           <button onClick={signOut}>Sign out</button>          
          </>
        )}
```

ErikCH (changes requested):
Just one more small issue, otherwise looks good!

src/fragments/start/getting-started/react/auth.mdx
joebuono: Ahh good catch!

ErikCH (approved):
👍

wlee221 (approved):

PR #4030, opened 2022-02-10
docs(geo): add docs for Geo `autocomplete` functionality

TreTuna:
_Description of changes:_
Add documentation about the new Geo Autocomplete functionality

image

### DIFF
--- a/src/fragments/lib/geo/js/search.mdx
+++ b/src/fragments/lib/geo/js/search.mdx
@@ -50,6 +50,82 @@ initializeMap();
 
 ![A search box on the top right corner of a map](/images/geocoder-search-box-map.png)
 
+### Autocomplete
+
+In order to help users find their query quickly, we provide several ways to accomplish autocomplete.
+
+By default, the Amplify search box will show search term suggestions while typing that the user can select from. These can be points of interest or addresses based on partial or misspelled free-form text.
+
+You can also show full location search results while typing by setting the `showResultsWhileTyping` option to `true`.
+
+```javascript
+createMaplibreGeocoder({
+    showResultsWhileTyping: true,
+});
+```
+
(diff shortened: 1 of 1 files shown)


src/fragments/lib/geo/js/search.mdx
```diff
@@ -50,6 +50,82 @@ initializeMap();
 
 ![A search box on the top right corner of a map](/images/geocoder-search-box-map.png)
 
+### Autocomplete
+
+In order to help users find their query quickly, we provide several ways to accomplish autocomplete.
+
+By default, the Amplify search box will show search term suggestions while typing that the user can select from. These can be points of interest or addresses based on partial or misspelled free-form text.
+
+You can also show full location search results while typing by setting the `showResultsWhileTyping` option to `true`.
```
Amplifiyer, line 59: I think this is complicated/confusing (suggestions being different than results. I know we understand it, most developers won't be able to in their first attempt making this API use confusing). 

I believe we should only expose one flag `suggestions` on or off OR `autocomplete` on or off etc. If the flag is on, we combine the results in the output, if it's off, only the location results are displayed on enter key press.

Any further customizations are done through `processResults` interceptor.


>Reduce decisions customers need to make. This is also known as "sensible defaults" as part of the "Convention over Configuration" umbrella.
When asking "What if the customer wants to do X if Y happens" the answer should almost always be to automatically do X for them and not expose it as a configuration option. Some customers might want to do this themselves, however we assert with Amplify they will be in the minority and we can help those customers via escape hatches or if feedback is large then we add in optional config later.

>Rule of Least Power - When possible, provide less options rather than more. This will promote architectural best practices via code & config, customers that wish to go against this can leverage escape hatches

TreTuna: After a UX review this morning, we modified this behavior and the docs have been updated to reflect that.

offlineprogrammer (approved):

PR #4052, opened 2022-02-18
fix(flutter): add reevaluate syncExpression section

HuiSF:
_Issue #, if available:_

https://github.com/aws-amplify/amplify-flutter/issues/1315

_Description of changes:_

Add reevaluate selective sync expression document for amplify-flutter.

Preview:

image


abdallahshaban557 (approved):
LGTM - Thank you, Hui!

cshfang (approved):

cshfang:
I think much of the wording has been copied over from other platforms so it's probably a change that needs to happen across the board but some of it seems to strongly imply (at least to me) that sync expressions will filter the data that is synced from the cloud to local stores rather than just being a client filter :/

PR #4054, opened 2022-02-21
`updateEndpoint` description improvement - optional fields & `attributes` fields type

BartoszKlonowski:
_Issue #, if available:_
closes aws-amplify/amplify-js#9608

_Description of changes:_
This pull request emphasizes about some of rules that apply to `updateEndpoint` method of aws-amplify/amplify-js library.
Especially the fact that all fields in the `endpoint.attributes` should be given as an array, otherwise the `updateEndpoint` method will fail with:
`AWSPinpointProvider - updateEndpoint failed [TypeError: undefined is not a function]`

It took me some decent amount of time to investigate on that error, so I believe that having these notes in the documentation should prove helpful to somebody.

There are also other notes given about the `endpoint` object structure, which can be confusing.
In original documentation it is unclear whether fields being optional or required comparing to other examples where optional fields are marked with `OPTIONAL` in the comments, while required are left without any note.

---

### DIFF
--- a/src/fragments/lib/analytics/js/update-endpoint.mdx
+++ b/src/fragments/lib/analytics/js/update-endpoint.mdx
@@ -3,11 +3,12 @@ An endpoint uniquely identifies your app within Pinpoint. In order to update you
 ```javascript
 import { Analytics } from 'aws-amplify';
 
+// NOTE: All fields are OPTIONAL
 await Analytics.updateEndpoint({
   address: 'xxxxxxx', // The unique identifier for the recipient. For example, an address could be a device token, email address, or mobile phone number.
   attributes: {
     // Custom attributes that your app reports to Amazon Pinpoint. You can use these attributes as selection criteria when you create a segment.
-    hobbies: ['piano', 'hiking']
+    hobbies: ['piano', 'hiking'] // MUST be an array, even single value
   },
   channelType: 'APNS', // The channel type. Valid values: APNS, GCM
   demographic: {


src/fragments/lib/analytics/js/getting-started.mdx
```diff
@@ -171,15 +171,19 @@ Amazon Pinpoint service requires an IAM policy in order to use the `record` API:
...
 import { Analytics } from 'aws-amplify';
 
+// NOTE: All fields are OPTIONAL
 Analytics.updateEndpoint({
     address: 'xxxxxxx', // The unique identifier for the recipient. For example, an address could be a device token, email address, or mobile phone number.
     attributes: {
         // Custom attributes that your app reports to Amazon Pinpoint. You can use these attributes as selection criteria when you create a segment.
         hobbies: ['piano', 'hiking'],
+        // NOTE: all fields in the attributes needs to be given as an array, even a single value:
+        age: [25],
```
david-mcafee, line 186:
```suggestion
```
Suggest removing in favor of single comment above

src/fragments/lib/analytics/js/getting-started.mdx
```diff
@@ -171,15 +171,19 @@ Amazon Pinpoint service requires an IAM policy in order to use the `record` API:
 ## Update Endpoint
 
 An endpoint uniquely identifies your app within Pinpoint. In order to update your <a href="https://docs.aws.amazon.com/pinpoint/latest/apireference/rest-api-endpoints.html" target="_blank">endpoint</a> use the `updateEndpoint()` method:
+Know that `updateEndpoint` method takes the object containing the configuration of `Analytics.AWSPinpoint.endpoint`, not the whole configuration object used to [set up the existing analytics backend](#set-up-existing-analytics-backend).
```
david-mcafee, line 174:
```suggestion
```
Though I definitely understand adding this, I think that we accomplish the same thing below by adding your "optional" note below. All of the possible attributes are also documented below.

david-mcafee (changes requested):
Thanks for creating this PR, @
BartoszKlonowski! I've left a few suggestions, which can be directly committed if you agree with them. Thank you, again!

src/fragments/lib/analytics/js/getting-started.mdx
BartoszKlonowski: Most important:
Am I able to use `updateEndpoint` to update for example: `Auth.region` or `Analytics.autoSessionRecord`?

If yes then right - we can remove this note, but if not, and if only the `Analytics.AWSPinpoint.endpoint` can be updated, then I would keep this note to reduce the confusion as much as possible.

BartoszKlonowski: @david-mcafee  Just a kind reminder in case my question went unnoticed ⬆️

BartoszKlonowski: @david-mcafee What is your opinion about my concerns?

BartoszKlonowski:
@david-mcafee Just a kind reminder about my response to your comment.

BartoszKlonowski:
@david-mcafee @renebrandel @siegerts Just a kind reminder about this PR

david-mcafee (changes requested):
Responded to @BartoszKlonowski's comments - I believe if we remove the comment from the intro paragraph, this PR is good to go!

david-mcafee (dismissed):
LGTM @BartoszKlonowski!! Thanks for this PR!

katiegoines (approved):
LGTM! Thank you!

BartoszKlonowski:
@david-mcafee @katiegoines I rebased the changes resolving conflicts and applied all the agreed changes (wasn't much) to the proper file - please recheck.

katiegoines (approved):

david-mcafee (approved):
LGTM!

PR #4073, opened 2022-02-25
Timeout fetch auth session flutter

abdallahshaban557:
_Description of changes:_

Added `timeout` function to `fetchAuthSession` for Flutter

### DIFF
--- a/src/fragments/lib/auth/flutter/access_credentials/10_fetchAuthSession.mdx
+++ b/src/fragments/lib/auth/flutter/access_credentials/10_fetchAuthSession.mdx
@@ -27,5 +27,6 @@ try {
 } on TimeoutException {
   print('fetchAuthSession timed out');
 }
+
 ```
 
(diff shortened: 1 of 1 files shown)


src/fragments/lib/auth/flutter/access_credentials/10_fetchAuthSession.mdx
```diff
@@ -13,3 +13,20 @@
 ```
 
 If the `getAWSCredentials` option is true, the result will contain AWS credentials and tokens.  If it is set to false, the result will contain a simple `isSignedIn` flag. 
+
+### Setting a timeout for fetchAuthSession
+
+`fetchAuthSession` timeout can be controlled using the `timeout` function as shown in the below example.
```
dnys1, line 19:
```suggestion
On spotty networks, the `fetchAuthSession` call can take upwards of a minute to either complete or fail due to internal retries. If this is too long, consider adding a custom timeout using the `timeout` function as shown in the below example.
```

dnys1 (approved):
Thanks!

siegerts (approved):

PR #4075, opened 2022-02-25
fix(docs): CLI - split supported resolver slots by operation

josefaidt:
_Issue #, if available:_

https://github.com/aws-amplify/amplify-cli/issues/9856

_Description of changes:_

- Splits single table into three separate tables by GraphQL operation
	- [Query](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts#L46-L47) table inherits original table
	- [Mutation](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts#L66-L67) table changes pre/postDataLoad with pre/postUpdate
	- [Subscription](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts#L85) table to gain the `preSubscribe` slot using the description from `preDataLoad`

### DIFF
--- a/src/pages/cli/graphql/custom-business-logic.mdx
+++ b/src/pages/cli/graphql/custom-business-logic.mdx
@@ -612,12 +612,36 @@ For example, the a resolver function file named `Mutation.createTodo.postAuth.2.
 
 ### Supported resolver slots
 
-|Sequence|Slot name|Description|
-|-|-|-|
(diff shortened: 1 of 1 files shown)


sachscode (approved):
LGTM

siegerts (approved):

PR #4100, opened 2022-03-03
Create a custom Algolia search for docs

jakeburden:
_Issue #, if available:_

_Description of changes:_
**Work in Progress**

This PR is a replacement for docsearch and builds a custom Algolia index on each deploy.
It utilities the Alogila search API to deliver custom query results.


tasks/build-algolia-search.mjs
```diff
@@ -0,0 +1,228 @@
...
+              }
+            });
+          }
+        });
+      }
+    });
+
+    // derive URL path
+    filename = filename.split('[');
+    filename[1] = platform;
```
tjleing, line 164: Isn't this going to mean that `platform` will look like `framework].mdx` or something?  not sure

tjleing: or the opposite I mean

tjleing: oh okay nvm

src/plugins/code-block.tsx
```diff
@@ -1,61 +1,61 @@
...
+const prism = require('prismjs');
 // eslint-disable-next-line @typescript-eslint/no-var-requires
-const loadLanguages = require("prismjs/components/");
+const loadLanguages = require('prismjs/components/');
 // eslint-disable-next-line @typescript-eslint/no-var-requires
-const Html5Entities = require("html-entities");
+const Html5Entities = require('html-entities');
 // eslint-disable-next-line @typescript-eslint/no-var-requires
-const unified = require("unified");
+const unified = import('unified').then((mod) => mod.default);
```
jakeburden, line 8: @timngyn we had to make this change because `unified` has updated to be an ESM module. We should make sure this works with the static site build as well as ensure we haven't missed any other deps that need to be switched from `require` to `import`.

calavera (approved):

timngyn (approved):

PR #4184, opened 2022-04-05
Reacting to updates for esm packages (custom aloglia search)

timngyn:
These changes are reacting to the `unified` and `unist-util-visit` package upgrades from the #4100 PR.

These packages converted to use `esm` modules and as a result, references to these modules need to be updated to use `import` statements instead of `require`.

_Description of changes:_
- Convert the plugins using `unist-util-visit` to be async modules so that we can `await import('unist-util-visit')` instead of using `require`.
- Update `next.js` to version 12.1.4 so that `next.config.js` can use async functions and await the plugins mentioned above.
- Reacting to changes in `next` that made `patch-next-scrolling.ts` no longer work. Instead use `history.replaceState` in conjunction with next `router.replace()` to handle updating the heading hashes during scroll.

Link to amplify app: https://timngyn-custom-algolia-search.d1oosfztpocl9c.amplifyapp.com/
To test:
1. Go to https://timngyn-custom-algolia-search.d1oosfztpocl9c.amplifyapp.com/lib/datastore/getting-started/q/platform/js/
2. Scroll around the page and verify that the page is not skipping around.
3. Click through the docs site and verify that back and forward browser buttons are working with the hashes


calavera (approved):
nice!

calavera (changes requested):
@timngyn you have to resolve conflicts

src/components/TableOfContents/index.tsx
```diff
@@ -82,36 +82,35 @@ export default function TableOfContents({children, title}) {
...
+                aTag.classList.add('active');
               });
             if (activeElement) {
-              router.replace(
-                window.location.href.split("#")[0],
-                window.location.href.split("#")[0] + "#" + activeElement,
-                {shallow: true, scroll: false},
-              );
+              // Update the url hash without updating the history.
+              history.replaceState(history.state, null, '#' + activeElement);
```
timngyn, line 97: Even though Next.js v12 added [changes for checking whether or not `scrollToHash()` should be called](https://github.com/vercel/next.js/pull/31921/files#diff-ad584158cf978a8bf737eec493b9e3b16100bec3e42e89954fd628a8e03f6295), I found that the scrolling was still broken.

This might've been because the there was still a `scrollToHash()` call inside the `componentDidUpdate()`: https://github.com/vercel/next.js/blob/b1a7b888906307d93b1a3b342a96d7b609606ec8/packages/next/client/index.tsx#L149-L151

Running the patch and viewing the output of `yarn build` still had janky scrolling where the page would jump around though. This made me think that there might've been some other changes in Next.js that needed to be fixed.

Instead of using next `router.replace()` here, I am using `history.replaceState()` because it's able to update the url with the updated hash without creating new state: https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState

calavera (approved):

tjleing:
DM summary:

> tiny bug: go to [this page](https://timngyn-custom-algolia-search.d1oosfztpocl9c.amplifyapp.com/start/getting-started/data-model/q/integration/js/), scroll down to the first header (create …) so that the hash in the url updates, click to different left sidebar link and hit the back button — url is back to non-hash

> `package.json` changes already dealt with by @jakeburden 

When that bug in the first quote is fixed, just let me know and I'll approve it!

timngyn:
@tjleing I think I have a fix for your bug. It should work as expected following your steps. Thank you again for taking the time to look at it!

calavera:
> @tjleing I think I have a fix for your bug. It should work as expected following your steps with the staging link. Thank you again for taking the time to look at it!

I've verified that it works as expected now 🎊

tjleing (approved):

timngyn:
Resolving conflicts

timngyn:
Closing PR because #4100 includes the changes now!

PR #4198, opened 2022-04-12
updated create IAM user Gif

Gksenthilarun:
# Descripiton

The Gif which provides for creating an IAM user be like a user needs the` AdministratorAccess` policy, but actually, the requirement is [AdministratorAccess-Amplify](https://docs.aws.amazon.com/amplify/latest/userguide/security-iam-awsmanpol.html). So which is updated.

### DIFF
--- a/src/fragments/lib/project-setup/native_common/prereq/common_body.mdx
+++ b/src/fragments/lib/project-setup/native_common/prereq/common_body.mdx
@@ -50,9 +50,9 @@ Specify the username of the new IAM user:
 Complete the user creation using the AWS console
 ```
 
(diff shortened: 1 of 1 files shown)


josefaidt (changes requested):
Hey @Gksenthilarun :wave: thank you for taking the time to file this, and apologies for the delay getting this resolved! Would you mind replacing the [gif in this repo](https://github.com/aws-amplify/docs/blob/main/public/images/user-creation.gif) with the one you've created?

Gksenthilarun:
Hey, I noticed that this is already committed via https://github.com/aws-amplify/docs/pull/4668/commits/ec506d4ccf698a36bf886f9f61f38d7ed835465c. So closing this.

PR #4210, opened 2022-04-14
Move indexes instead of deleting objects

calavera:
This operation is atomic, so customers don't find situations
when there are no objects in the index.

This change also makes the index name configurable per environment,
so we can have different indexes in production and staging.

@jakeburden test this in staging before merging 🙏 


timngyn (approved):

jakeburden (approved):

jakeburden:
Staging link:
https://move-indexes.d1oosfztpocl9c.amplifyapp.com/

PR #4310, opened 2022-05-17
Update node/npm versions

abdallahshaban557:
_Description of changes:_

Update node/npm versions, and removed callouts for CRA NPM/Node version numbers.


josefaidt (approved):

david-mcafee (commented):
Have we verified that the latest version of CRA works with these versions of Node / NPM? If not, we'll want to keep the callout and update the versions. Otherwise, this looks good! Thanks for doing this!

david-mcafee (approved):
@abdallahshaban557 tested the latest version of CRA, so we're good to go with these changes. Than you, @abdallahshaban557!!

Issue #4161, opened 2022-03-26
Add a last updated label on the doc pages.


Fanna1119:
with speed amplify changes, it is frustrating jumping around the docs, not knowing how up to date it is.

timngyn:
Hi @Fanna1119, thank you for opening this issue! [docs.amplify.aws](https://docs.amplify.aws/) now has when a page was last updated. It was merged in from this PR: #4341

It's located in the left menu and looks like this:
![image](https://user-images.githubusercontent.com/54393192/219113376-42457b97-e8b0-4e05-9573-45bedaf182ee.png)

PR #4341, opened 2022-05-27
Add when a page was last updated into the Docs site

timngyn:
_Issue #, if available:_ #4161, Add a last updated label on the doc pages.

_Description of changes:_

- Add a script that runs during the `preBuild` phase on Amplify to programmatically add frontmatter to each mdx. This frontmatter contains a `lastUpdated` property that contains the date of the last commit for the file.
- Displays the last updated date inside the `Menu` component


src/plugins/code-block.tsx
```diff
@@ -45,7 +45,8 @@ module.exports = (async () => {
     'groovy',
     'dart',
     'text',
-    'plain'
+    'plain',
+    'sh'
```
timngyn, line 49: Adding this because the build kept complaining about now having the `sh` language

src/components/Menu/index.tsx
```diff
@@ -22,102 +29,125 @@ type MenuProps = {
...
-    }
-    return <MenuOpenButton openMenu={this.openMenu} />;
+        </div>
+      </MenuStyle>
+    );
+  }
+  return <MenuOpenButton openMenu={openMenu} />;
+}
+
+export default forwardRef(Menu);
```
timngyn, line 135: Have to use `forwardRef` since the `ref` property being used when the `Menu` component is created inside the `Page` component.

The `ref` is used so that the `SidebarLayoutToggle` component can open and close the menu on mobile screen sizes

src/components/Menu/index.tsx
```diff
@@ -22,102 +29,125 @@ type MenuProps = {
...
+  }, []);
 
-  closeMenu = () => {
-    this.setState({
-      isOpen: false
-    });
+  useImperativeHandle(ref, () => ({
+    closeMenu: () => closeMenu(),
+    openMenu: () => openMenu()
+  }));
```
timngyn, line 47: Since we're using `useForwardRef` on this component, we need to expose the functions that we're trying to share. `SidebarLayout` uses the `closeMenu` and `openMenu` functions

src/components/Page/index.tsx
```diff
@@ -92,7 +106,10 @@ export default function Page({
       />
     );
   }
-  const [menuIsOpen, setMenuIsOpen] = useState(false);
+
+  if (filterKey !== prevFilterKey) {
+    setPrevFilterKey(filterKey);
```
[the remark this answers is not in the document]
timngyn: Ah thanks for calling that out, I think I forgot to remove it when I refactored away from my object implementation. Removing it now

jacoblogan (approved):

katiegoines (approved):

PR #4362, opened 2022-06-07
Feedback merge fixes

timngyn:
_Issue #, if available:_

_Description of changes:_ 

- Noticed on prod that the feedback component is delayed after click because I'm awaiting an Amplify API call. Set the feedback state right after click to make it more "responsive" to click
- Forked PRs are not able to build on github workflow because they don't have access to environment variables in the Docs repo. Create a github environment to share secrets only for testing PRs
- Added script `create-awsconfig.sh` to create a blank `aws-exports.js` file so that contributors can run `yarn dev` without needed to do an `amplify pull`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,7 @@
 name: Test
-on: pull_request
+on: 
+  pull_request_target:
+    branches: [main]
 permissions:
   id-token: write
   contents: read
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -7,6 +7,7 @@ jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
+    environment: test-pr
```
timngyn, line 10: Created an Github environment called `test-pr` to have different secrets pointing to dummy app so that fork PRs can access them (still need to test this though).

PR #4372, opened 2022-06-08
Workaround for awsexports

timngyn:
_Issue #, if available:_

_Description of changes:_

- Noticed on prod that the feedback component is delayed after click because I'm awaiting an Amplify API call. Set the feedback state right after click to make it more "responsive" to click
- Added script `create-awsconfig.sh` to create a blank `aws-exports.js` file so that contributors can run `yarn dev` without needed to do an `amplify pull`

### DIFF
--- a/build_support/create-awsconfig.sh
+++ b/build_support/create-awsconfig.sh
@@ -1,0 +1,5 @@
+FILE=aws-exports.js
+if [ ! -f "./src/$FILE" ]; then
+  echo "$FILE does not exist in ./src/, creating blank aws-exports.js"
+  touch "./src/$FILE"
+fi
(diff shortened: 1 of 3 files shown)


calavera (approved):

build_support/create-awsconfig.sh
```diff
@@ -0,0 +1,6 @@
+FILE=aws-exports.js
+if [ ! -f "./src/$FILE" ]; then
+  echo "$FILE does not exist in ./src/, creating aws-exports.js"
+  touch "./src/$FILE"
+  echo "export default {}" > "./src/$FILE"
```
timngyn, line 5: Add this line recommended by @josefaidt. With this addition it gets rid of a warning inside code editor saying that `aws-exports` wasn't a module

PR #4393, opened 2022-06-15
doc: add filtered list query

danielleadams:
_Description of changes:_

Adds an example of changes to a a list query with filter when looking for both owner formats.

### DIFF
--- a/src/pages/cli/migration/identity-claim-changes.mdx
+++ b/src/pages/cli/migration/identity-claim-changes.mdx
@@ -13,7 +13,7 @@ In the Amplify CLI >= v8.0.3 owner-based `@auth` rule uses `"username"` as the d
 
 There are no breaking changes to your GraphQL API when using the new default identity claim. The resolvers will store these values in your DynamoDB tables in the format of `<sub>::<username>`, and return `username` to the client code by default.
 
-While the other directives will work as they currently expected to function (i.e. `@searchable`, `@function`), using custom queries to your databases may need to be changed. For example, if you are performing a query with the `filter` parameter, you will need to change it from using the `filter` query with just 1 argument to using an `or` conditional statement like the following:
+While the other directives will work as they are currently expected to function (i.e. `@searchable`, `@function`), using custom queries to your databases may need to be changed. For example, if you are performing a query with the `filter` parameter, you will need to change it from using the `filter` query with just 1 argument to using an `or` conditional statement like the following:
 
 ```graphql
 query MyQuery {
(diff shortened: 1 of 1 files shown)


src/pages/cli/migration/identity-claim-changes.mdx
```diff
@@ -13,13 +13,31 @@ In the Amplify CLI >= v8.0.3 owner-based `@auth` rule uses `"username"` as the d
 
 There are no breaking changes to your GraphQL API when using the new default identity claim. The resolvers will store these values in your DynamoDB tables in the format of `<sub>::<username>`, and return `username` to the client code by default.
 
-While the other directives will work as they currently work (ie. `@searchable`, `@function`), using custom queries to your databases, such as OpenSearch, may need to be changed. For example, if you are using OpenSearch for a custom query like the one below, you will need your queries to account for the format of the stored owner field with a matching operation (https://docs.amplify.aws/cli/graphql/search-and-result-aggregations/#supported-search-operations). Here's an example of using the wildcard operation:
+While the other directives will work as they currently expected to function (ie. `@searchable`, `@function`), using custom queries to your databases may need to be changed. For example, if you are performing a query with the `filter` parameter, you will need to change it from using the `filter` query with just 1 argument to using an `or` conditional statement like the following:
```
josefaidt, line 16:
```suggestion
While the other directives will work as they currently expected to function (i.e. `@searchable`, `@function`), using custom queries to your databases may need to be changed. For example, if you are performing a query with the `filter` parameter, you will need to change it from using the `filter` query with just 1 argument to using an `or` conditional statement like the following:
```

josefaidt (approved):

src/pages/cli/migration/identity-claim-changes.mdx
```diff
@@ -13,13 +13,31 @@ In the Amplify CLI >= v8.0.3 owner-based `@auth` rule uses `"username"` as the d
 
 There are no breaking changes to your GraphQL API when using the new default identity claim. The resolvers will store these values in your DynamoDB tables in the format of `<sub>::<username>`, and return `username` to the client code by default.
 
-While the other directives will work as they currently work (ie. `@searchable`, `@function`), using custom queries to your databases, such as OpenSearch, may need to be changed. For example, if you are using OpenSearch for a custom query like the one below, you will need your queries to account for the format of the stored owner field with a matching operation (https://docs.amplify.aws/cli/graphql/search-and-result-aggregations/#supported-search-operations). Here's an example of using the wildcard operation:
+While the other directives will work as they are currently expected to function (i.e. `@searchable`, `@function`), using custom queries to your databases may need to be changed. For example, if you are performing a query with the `filter` parameter, you will need to change it from using the `filter` query with just 1 argument to using an `or` conditional statement like the following:
+
```
sachscode, line 17: Would it be helpful here to show both the old behavior ( list query with single parameter ) and the new behavior ( like this one ) so that the reader could understand the difference ?

sachscode (commented):
LGTM but just one suggestion

src/pages/cli/migration/identity-claim-changes.mdx
danielleadams: Yep! That works for me.

danielleadams: @sachscode I think less words is easier to understand. I don't know if we should add the old snippet because it might be confusing to show the before and after.

sachscode (approved):

arundna (approved):

PR #4428, opened 2022-06-29
expanded warnings and examples about working with stale data

svidgen:
_Issue #, if available:_

_Description of changes:_

Expands on the warnings about working with stale data in DataStore, updates the save examples to reflect recommended usage, and adds examples of addressing with `observeQuery`.

### DIFF
--- a/src/fragments/lib/datastore/flutter/data-access/observe-update-snippet.mdx
+++ b/src/fragments/lib/datastore/flutter/data-access/observe-update-snippet.mdx
@@ -1,0 +1,55 @@
+```dart
+void main() {
+  runApp(MyApp());
+}
+
+class MyApp extends StatefulWidget {
+  const MyApp({Key? key}) : super(key: key);
+
+  @override
+  State<MyApp> createState() => _MyAppState();
+}
+
+class _MyAppState extends State<MyApp> {
+  StreamSubscription<QuerySnapshot<Post>>? _stream;
+
+  // A place to keep the post we care about
+  Post? _post;
+
+  // Initialize a boolean indicating if the sync process has completed
+  bool _isSynced = false;
+
+  // Initialize the libraries
+  ...
+
+  void observeQuery() {
+    _stream = Amplify.DataStore.observeQuery(
+      Post.classType,
+      where: Post.ID.eq('123')
+    ).listen((QuerySnapshot<Post> snapshot) {
+      setState(() {
+        _post = snapshot.items.first;
+        _isSynced = snapshot.isSynced;
+      });
+    });
+  }
+
+  Future<void> updatePost(newTitle) async {
+      final updatedPost = _post.copyWith(id: _post.id, title: "Updated Title");
+      await Amplify.DataStore.save(updatedPost);
+
+      // we do not need to explicitly set _post here; observeQuery will see
+      // the update and set the variable.
+  }
+
+  // Build function and UI elements
+  ...
+
+  @override
+  void dispose() {
+    _stream?.cancel();
+    super.dispose();
+  }
+}
+
+```
--- a/src/fragments/lib/datastore/flutter/data-access/observe-update-snippet.mdx
+++ b/src/fragments/lib/datastore/flutter/data-access/observe-update-snippet.mdx
@@ -12,7 +12,16 @@ class _MyAppState extends State<MyApp> {
   // A reference to the retrieved post
   late Post _post;
 
-  // Initialize the Amplify libraries
+  @override
+  void initState() {
+    super.initState();
+    _configure();
+  }
+  
+  // Initialize the Amplify libraries and call `observeQuery`
+  Future<void> _configure() async {
+    // ...
+  }
   ...
 
   void observeQuery() {
@@ -26,8 +35,8 @@ class _MyAppState extends State<MyApp> {
     });
   }
 
-  Future<void> updatePost(newTitle) async {
-      final updatedPost = _post.copyWith(id: _post.id, title: 'Updated Title');
+  Future<void> updatePost(String newTitle) async {
+      final updatedPost = _post.copyWith(title: newTitle);
       await Amplify.DataStore.save(updatedPost);
 
       // we do not need to explicitly set _post here; observeQuery will see
--- a/src/fragments/lib/datastore/flutter/data-access/observe-update-snippet.mdx
+++ b/src/fragments/lib/datastore/flutter/data-access/observe-update-snippet.mdx
@@ -1,8 +1,4 @@
 ```dart
-void main() {
-  runApp(MyApp());
-}
-
 class MyApp extends StatefulWidget {
   const MyApp({Key? key}) : super(key: key);
 
@@ -13,13 +9,10 @@ class MyApp extends StatefulWidget {
 class _MyAppState extends State<MyApp> {
   StreamSubscription<QuerySnapshot<Post>>? _stream;
 
-  // A place to keep the post we care about
-  Post? _post;
+  // A reference to the retrieved post
+  late Post _post;
 
-  // Initialize a boolean indicating if the sync process has completed
-  bool _isSynced = false;
-
-  // Initialize the libraries
+  // Initialize the Amplify libraries
   ...
 
   void observeQuery() {
@@ -29,13 +22,12 @@ class _MyAppState extends State<MyApp> {
     ).listen((QuerySnapshot<Post> snapshot) {
       setState(() {
         _post = snapshot.items.first;
-        _isSynced = snapshot.isSynced;
       });
     });
   }
 
   Future<void> updatePost(newTitle) async {
-      final updatedPost = _post.copyWith(id: _post.id, title: "Updated Title");
+      final updatedPost = _post.copyWith(id: _post.id, title: 'Updated Title');
       await Amplify.DataStore.save(updatedPost);
 
       // we do not need to explicitly set _post here; observeQuery will see
--- a/src/fragments/lib/datastore/js/data-access/observe-update-snippet.mdx
+++ b/src/fragments/lib/datastore/js/data-access/observe-update-snippet.mdx
@@ -5,16 +5,16 @@
 // other frameworks.
 
 function App() {
-  const [todo, setTodo] = useState<Todo>();
+  const [post, setPost] = useState<Post>();
 
   useEffect(() => {
     /**
      * This keeps `todo` fresh.
      */
-    const sub = DataStore.observeQuery(Todo, (c) =>
+    const sub = DataStore.observeQuery(Post, (c) =>
       c.id("eq", "e4dd1dc5-e85c-4566-8aaa-54a801396456")
     ).subscribe(({ items }) => {
-      setTodo(items[0]);
+      setPost(items[0]);
     });
 
     return () => {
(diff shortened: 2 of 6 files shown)


src/fragments/lib/datastore/js/data-access/observe-update-snippet.mdx
```diff
@@ -0,0 +1,55 @@
+```ts
+// Example showing how to observe the model and keep state updated before
+// performing a save. This uses the useEffect React hook, but you can
+// substitute for a similar mechanism in your application lifecycle with
+// other frameworks.
```
lawmicha, line 5: This seems like a very complex example on this page. we're also using `observeQuery` before the "Real Time" section, but understandably, the paragraph before this example does reference that section. 

I'm thinking what the ideal would be... 
- "Getting started" section 1 remains the same
- "Manipulating data" section 2, under "**Avoid working with stale data!** points to section 4
- "Real time" - move this up from section 7 to section 3
- "Putting it all together" - section 4, talk about the two options - "observe and save" and "query then save" . When talking about "observe and save", reference back to section 3 "real time" and then show the `observe-update-snippet` example. then a brief explanation that if you don't subscribe to data and update the models in memory, then you'll have to perform the "query then save" option. This is so that we provide the recommended way is "observe and save" from the indepth example code here, but explain if they don't subscribe to data, they should query-merge-save their models.

src/fragments/lib/datastore/native_common/data-access.mdx
```diff
@@ -44,10 +44,26 @@ import flutter8 from "/src/fragments/lib/datastore/flutter/data-access/update-sn
 
   **Avoid working with stale data!**
 
-  Model instances become stale when the original records are overwritten. Saving a record from a copy of a stale model instance can produce errors and/or unexpected states. Fetch records and make fresh copies as close to saving as possible. This can be aided by [observing](/lib/datastore/real-time) a model instance for changes.
+  Model instances which store values, such as those from the results of a `DataStore.Query` operation, can become stale and outdated when their properties are updated. This can be the result of a manual change or even a side effect of [real time data](/lib/datastore/sync#distributed-data) being received by the application. In order to ensure you are performing mutations on the latest committed state to the system, either perform a query directly before the `DataStore.save()` operation or observe the model to keep the state updated at all times and perform mutations on that latest state referencing the model instance. The preceding example demonstrates one approach. The following example demonstrates the `observeQuery` approach.
```
lawmicha, line 47: this makes sense, if the local data in memory is up-to-date because the app is subscribed to the local store via observe or observeQuery API then the local data will always be the latest before DataStore.save().

src/fragments/lib/datastore/js/data-access/update-snippet.mdx
```diff
@@ -1,11 +1,12 @@
...
-  })
-);
+async function updatePost(id, newTitle) {
+  const original = await DataStore.query(Post, id);
+  await DataStore.save(
+    Post.copyOf(original, updated => {
+      updated.title = newTitle
+    })
+  );
+}
```
lawmicha, line 9: If the app is not subscribing to real time events, and had performed a query for a list of models, and end-user update one of the models, app calls DataStore.save(), then it should get the latest local model and replace fields that is being updated, title in this case

lawmicha: would this example conflict with the following example of performing `observeQuery` and maintaining the latest local data in memory? Since with the real time APIs, they don't need to query before saving

HuiSF (approved):

src/fragments/lib/datastore/flutter/data-access/observe-update-snippet.mdx
```diff
@@ -0,0 +1,55 @@
+```dart
+void main() {
+  runApp(MyApp());
+}
```
salihgueler, line 4: If we do not care about MaterialApp and any other elements, why do we need this main function in the first place?

src/fragments/lib/datastore/flutter/data-access/observe-update-snippet.mdx
```diff
@@ -0,0 +1,55 @@
...
+      // we do not need to explicitly set _post here; observeQuery will see
+      // the update and set the variable.
+  }
+
+  // Build function and UI elements
+  ...
+
+  @override
+  void dispose() {
+    _stream?.cancel();
```
salihgueler, line 50: Great touch on cancelling!

src/fragments/lib/datastore/flutter/data-access/observe-update-snippet.mdx
```diff
@@ -0,0 +1,55 @@
...
+  const MyApp({Key? key}) : super(key: key);
+
+  @override
+  State<MyApp> createState() => _MyAppState();
+}
+
+class _MyAppState extends State<MyApp> {
+  StreamSubscription<QuerySnapshot<Post>>? _stream;
+
+  // A place to keep the post we care about
```
salihgueler, line 16:
```suggestion
  // A reference to the retrieved post
```

src/fragments/lib/datastore/flutter/data-access/observe-update-snippet.mdx
```diff
@@ -0,0 +1,55 @@
...
+    ).listen((QuerySnapshot<Post> snapshot) {
+      setState(() {
+        _post = snapshot.items.first;
+        _isSynced = snapshot.isSynced;
+      });
+    });
+  }
+
+  Future<void> updatePost(newTitle) async {
+      final updatedPost = _post.copyWith(id: _post.id, title: "Updated Title");
```
salihgueler, line 38: Wouldn't `_post` want a question mark. Also please try to use single quotes with Strings in Dart and Flutter
```suggestion
      final updatedPost = _post.copyWith(id: _post.id, title: 'Updated Title');
```

src/fragments/lib/datastore/flutter/data-access/observe-update-snippet.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+  @override
+  State<MyApp> createState() => _MyAppState();
+}
+
+class _MyAppState extends State<MyApp> {
+  StreamSubscription<QuerySnapshot<Post>>? _stream;
+
+  // A place to keep the post we care about
+  Post? _post;
```
salihgueler, line 17: If we want to leverage from the _isFirstSync kind of variable, then we can turn this into a non-nullable type with late keyword.

src/fragments/lib/datastore/flutter/data-access/observe-update-snippet.mdx
```diff
@@ -0,0 +1,47 @@
...
+      where: Post.ID.eq('123')
+    ).listen((QuerySnapshot<Post> snapshot) {
+      setState(() {
+        _post = snapshot.items.first;
+      });
+    });
+  }
+
+  Future<void> updatePost(newTitle) async {
+      final updatedPost = _post.copyWith(id: _post.id, title: 'Updated Title');
```
dnys1, line 30: `copyWith` will preserve non-overwritten fields, so ID doesn't need to be manually included.

```suggestion
  Future<void> updatePost(String newTitle) async {
      final updatedPost = _post.copyWith(title: newTitle);
```

src/fragments/lib/datastore/flutter/data-access/observe-update-snippet.mdx
dnys1: +1

src/fragments/lib/datastore/flutter/data-access/observe-update-snippet.mdx
```diff
@@ -0,0 +1,47 @@
...
+  State<MyApp> createState() => _MyAppState();
+}
+
+class _MyAppState extends State<MyApp> {
+  StreamSubscription<QuerySnapshot<Post>>? _stream;
+
+  // A reference to the retrieved post
+  late Post _post;
+
+  // Initialize the Amplify libraries
```
dnys1, line 15: It may not be immediately obvious where this should happen and where `observeQuery` should be called.

```suggestion
  @override
  void initState() {
    super.initState();
    _configure();
  }
  
  // Initialize the Amplify libraries and call `observeQuery`
  Future<void> _configure() async {
    // ...
  }
```

dnys1 (approved):
Couple comments from a Flutter folk. Looks great - thanks!

src/fragments/lib/datastore/flutter/data-access/observe-update-snippet.mdx
svidgen: Thanks!

src/fragments/lib/datastore/flutter/data-access/observe-update-snippet.mdx
svidgen: Thanks again!

src/fragments/lib/datastore/js/data-access/update-snippet.mdx
svidgen: Yeah, I'm not so sure. Not every use-case will call for live data. The aim is just to cover both use cases.

src/fragments/lib/datastore/js/data-access/observe-update-snippet.mdx
svidgen: Yeah, the linearity issue makes sense. But, it also seems a bit out of place to put a "Putting it all together" link in the middle of the list. It's more like, "putting the things we've introduced so far together." 😛 Having a discussion about new pages and menu ordering makes me concerned that we'll rabbit hole on information architecture, whereas I think it's more pressing to get the example code out in a _reasonable_ spot soon.

That said, I could add the examples to the realtime data page instead in a new, third page section and just provide a link here. It may improve linearity. And, it doesn't _seem_ like an IA rabbit hole to me.

What do you think?

(Then again, maybe I'm already *in* an IA rabbit hole! 🐇 ⛳)

salihgueler (approved):

src/fragments/lib/datastore/js/data-access/observe-update-snippet.mdx
lawmicha: I agree we shouldn't rabbit hole on where the information is, better to decouple that from the example itself, and also placing it on the realtime data page may improve linearity, so i'll leave it for you to decide, either keep it here and let it sit in production to get feedback, or attempt the move to realtime data page now

src/fragments/lib/datastore/js/data-access/observe-update-snippet.mdx
```diff
@@ -0,0 +1,55 @@
...
+  const [todo, setTodo] = useState<Todo>();
+
+  useEffect(() => {
+    /**
+     * This keeps `todo` fresh.
+     */
+    const sub = DataStore.observeQuery(Todo, (c) =>
+      c.id("eq", "e4dd1dc5-e85c-4566-8aaa-54a801396456")
+    ).subscribe(({ items }) => {
+      setTodo(items[0]);
```
lawmicha, line 17: I see the other Android/Flutter example using a Post, any reason this is diverging here?

svidgen: Done!

src/fragments/lib/datastore/flutter/data-access/observe-update-snippet.mdx
svidgen: TY!

lawmicha (approved):

svidgen:
Because this is above the approval threshold, I'm moving this back to draft mode to prevent merging while awaiting Android review.

undefobj (approved):

offlineprogrammer (approved):

svidgen:
Thanks for the Android corrections @mikepschneider!

mikepschneider (approved):

PR #4429, opened 2022-06-29
docs(hub): add hub remove code snippet example

helgabalashova:
_Issue #, if available:_   [#amplify-js/8807](https://github.com/aws-amplify/amplify-js/issues/8807)

_Description of changes:_ It is not clear from Hub docs how listening event is removed. Code snippet is added to clarify how it can be properly done.
Screen Shot 2022-06-29 at 1 26 02 PM

### DIFF
--- a/src/fragments/lib/utilities/js/hub.mdx
+++ b/src/fragments/lib/utilities/js/hub.mdx
@@ -73,7 +73,18 @@ The `event` field is recommended to be a small string without spaces such as `si
 
 ### Stop Listening
 
-Hub provides a way to stop listening for messages with `Hub.remove(channel: string | RegExp, listener: callback)`. This may be useful if you no longer need to receive messages in your application flow, as well as to avoid any memory leaks on low powered devices when you are sending large amounts of data through Hub on multiple channels.
+Hub provides a way to stop listening for messages with `Hub.remove(channel: string | RegExp, listener: callback)`:
+```javascript
+const callback = (data) => {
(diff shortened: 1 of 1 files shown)


src/fragments/lib/utilities/js/hub.mdx
```diff
@@ -73,7 +73,18 @@ The `event` field is recommended to be a small string without spaces such as `si
...
+const callback = (data) => {
+  console.log('Listening for all messages: ', data.payload.data);
+}
+Hub.listen(/.*/, callback);
+
+/* later */
+Hub.remove(/.*/, callback);
+```
+
+This may be useful if you no longer need to receive messages in your application flow, as well as to avoid any memory leaks on low powered devices when you are sending large amounts of data through Hub on multiple channels.
```
abdallahshaban557, line 87: what is meant by channels here? I think that should be clarified a bit.

abdallahshaban557 (commented):
just 1 comment

src/fragments/lib/utilities/js/hub.mdx
manueliglesias: @abdallahshaban557  The next paragraph talks about channels, do you think it would be better to swap the order of the "Stop Listening" and "Channels" sections?

rachelnabors: @helgabalashova I approve this change. Would you mind going ahead with it?

PR #4464, opened 2022-07-14
fix(cli): add callout for connected private models

phani-srikar:
_Issue #, if available:_

_Description of changes:_
Added a callout for developers using the `private` level access control on their `@model`s about the behavior with connected models 


josefaidt (approved):

phani-srikar:
> Just have one question for the context. If the `Task` does not have the same owner as `Todo`, the `Todo` user can still get these tasks right? (where the `private` rule is applied)

Yes Right

PR #4465, opened 2022-07-14
feat(cli): add custom primary key with relation model section

AaronZyLee:
_Issue #, if available:_

_Description of changes:_
Added custom primary key with relation model section in `CLI/API(GraphQL)/Data modeling`

### DIFF
--- a/src/pages/cli/graphql/data-modeling.mdx
+++ b/src/pages/cli/graphql/data-modeling.mdx
@@ -549,7 +549,494 @@ type Address @model {
 ```
 
 ### Relational models with custom primary keys
-TODO
+You can define custom primary keys for relational models. The transformer will handle the foreign key fields needed by queries and mutations, which allows you to retrieve data across different relations. 
+
+You can choose the default relations with **implicit** fields. This will result in auto generation of additional foreign key fields refering its counterpart primary key and sort key fields.
+
+<Callout>
+In "has many" relations, the foreign key fields are only generated in child type.
+</Callout>
+
+<BlockSwitcher>
+
+<Block name='Implicit Bi-directional "has one"'>
+
+```graphql
+type Project @model {
+  projectId: ID! @primaryKey(sortKeyFields:["name"])
+  name: String!
+  team: Team @hasOne
+}
+
+type Team @model {
+  teamId: ID! @primaryKey(sortKeyFields:["name"])
+  name: String!
+  project: Project @belongsTo
+}
+```
+
+This generates additional foreign key fields in both parent and child types.
+
+Model fields:
+```graphql
+type Project {
+  projectId: ID!
+  name: String!
+  team: Team
+  createdAt: AWSDateTime!
+  updatedAt: AWSDateTime!
+  projectTeamTeamId: ID # auto generated child primary key reference
+  projectTeamName: String # auto generated child sort key reference
+}
+
+type Team {
+  teamId: ID!
+  name: String!
+  project: Project
+  createdAt: AWSDateTime!
+  updatedAt: AWSDateTime!
+  teamProjectProjectId: ID # auto generated parent primary key reference
+  teamProjectName: String # auto generated parent sort key reference
+}
+```
+Input fields:
+```graphql
+input CreateProjectInput {
+  projectId: ID!
+  name: String!
+  projectTeamTeamId: ID # auto generated child primary key reference
+  projectTeamName: String # auto generated child sort key reference
+}
+
+input UpdateProjectInput {
+  projectId: ID!
+  name: String!
+  projectTeamTeamId: ID # auto generated child primary key reference
+  projectTeamName: String # auto generated child sort key reference
+}
+input CreateTeamInput {
+  teamId: ID!
+  name: String!
+  teamProjectProjectId: ID # auto generated parent primary key reference
+  teamProjectName: String # auto generated parent sort key reference
+}
+
+input UpdateTeamInput {
+  teamId: ID!
+  name: String!
+  teamProjectProjectId: ID # auto generated parent primary key reference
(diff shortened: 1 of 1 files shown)


src/pages/cli/graphql/data-modeling.mdx
```diff
@@ -549,7 +549,494 @@ type Address @model {
 ```
 
 ### Relational models with custom primary keys
-TODO
+You can define custom primary keys for relational models. The transformer will handle the foreign key fields needed by queries and mutations, which allows you to retrieve data across different relations.
```
HuiSF, line 552: "transformer" this word is not used previously in the documentation,  and "The transformer will handle the foreign key fields needed by queries and mutations" sounds more like an implementation detail. Should we expose it here?

src/pages/cli/graphql/data-modeling.mdx
```diff
@@ -549,7 +549,494 @@ type Address @model {
 ```
 
 ### Relational models with custom primary keys
-TODO
+You can define custom primary keys for relational models. The transformer will handle the foreign key fields needed by queries and mutations, which allows you to retrieve data across different relations. 
+
+You can choose the default relations with **implicit** fields. This will result in auto generation of additional foreign key fields refering its counterpart primary key and sort key fields.
```
HuiSF, line 554: Similar here "implicit fields" is a new term in the documentation. 
Would be better to use existing terms
e.g.

The fields arguments of `@belongsTo` directive.
`@belongTo` can be used without the `fields` argument. (this is the implicit I believe)

src/pages/cli/graphql/data-modeling.mdx
```diff
@@ -549,7 +549,494 @@ type Address @model {
...
+
+<Callout>
+In "has many" relations, the foreign key fields are only generated in child type.
```
HuiSF, line 557: This is always true regardless using CPK or not, correct? 
If this is not new, I'd recommend to omit it.

src/pages/cli/graphql/data-modeling.mdx
```diff
@@ -549,7 +549,494 @@ type Address @model {
...
+</Callout>
+
+<BlockSwitcher>
+
+<Block name='Implicit Bi-directional "has one"'>
```
HuiSF, line 562: Bidirectional one-to-one could be more precise. Ditto "implicit" term.

src/pages/cli/graphql/data-modeling.mdx
```diff
@@ -549,7 +549,494 @@ type Address @model {
...
+input UpdateTeamInput {
+  teamId: ID!
+  name: String!
+  teamProjectProjectId: ID # auto generated parent primary key reference
+  teamProjectName: String # auto generated parent sort key reference
+}
+```
+</Block>
+
+<Block name='Implicit Uni-directional "has one"'>
```
HuiSF, line 633: Could be just `has one`. "implicit" term ditto.

src/pages/cli/graphql/data-modeling.mdx
```diff
@@ -549,7 +549,494 @@ type Address @model {
...
+  content: String!
+  postCommentsPostId: ID # auto generated parent primary key reference
+  postCommentsTitle: String # auto generated parent sort key reference
+}
+```
+</Block>
+
+</BlockSwitcher>
+
+You can also customize the foreign key fields instead of the auto-generated ones in the schema by **explicitly** defining the `fields` parameter in relation directives.
```
HuiSF, line 774: Ditto better to use existing terms. and -

`fields` parameter => `fields` argument

src/pages/cli/graphql/data-modeling.mdx
```diff
@@ -549,7 +549,494 @@ type Address @model {
...
+
+<Callout>
+The order in "fields" parameter matters. The first element should refer the primary key field and the rest are sort key fields.
```
HuiSF, line 777:
```suggestion
The order of the elements in the value of "fields" argument matters. The first element should always be the primary key of the associated model; and remaining elements should match the `sortKeyFields` specified in the `@primaryKey` directive of t he associated model.
```

HuiSF (commented):
Thanks for adding the doc, overall it makes sense to me.
Added some suggestions.

src/pages/cli/graphql/data-modeling.mdx
AaronZyLee: This is true for `hasOne` and `belongsTo`. But for `hasMany` it is different. When defining `@hasMany(fields:)`, We are passing the PK and SK fields from the parent and associate it with the child secondary index. So I think the `associated model` here is not accurate enough to cover all cases.

src/pages/cli/graphql/data-modeling.mdx
AaronZyLee: The "transformer" term is removed

src/pages/cli/graphql/data-modeling.mdx
AaronZyLee: I am using `implicit` and `explicit` terms in the block header. So I add additional explanations on these two words

src/pages/cli/graphql/data-modeling.mdx
AaronZyLee: Yes this is not related to CPK but the general `hasMany` cases.

src/pages/cli/graphql/data-modeling.mdx
AaronZyLee: The `Bi-directional "has one"` is derived from the previous section

src/pages/cli/graphql/data-modeling.mdx
AaronZyLee: I temporarily replace the word with `relational`

PR #4467, opened 2022-07-15
Update husky to fix dependabot issue 25

timngyn:
_Issue #, if available:_ https://github.com/aws-amplify/docs/security/dependabot/25

_Description of changes:_
- Update husky to `^8.0.1` so there isn't a dependency for `semver-regex`
- Update cspell to so that it works with this version of husky 

_To test:_
1. Clone branch
2. Run `yarn install` in `/docs` directory
3. Edit an MDX file with a typo
4. Stage the file and try to commit it
   - Verify that commit is unsuccessful because there is a typo

image


cspell.json
```diff
@@ -318,6 +318,7 @@
     "baseQueryResultHandler",
     "baseRefreshIntervalInSeconds",
     "bbox",
+    "bcfa",
```
timngyn, line 321: Added these words because they came up during `yarn spellcheck` as unknown words. These words were being used as IDs and so I added them in here to make the spellcheck happy

timngyn:
TODO: Make sure that the pre-commit also works on windows because of the single quotes don't expand the glob

offlineprogrammer (approved):

PR #4485, opened 2022-07-27
Adding additional info on contenttype and type for s3

ahilashsasidharan:
_Issue #, if available:_
#10053 in aws/amplify-js repo
https://github.com/aws-amplify/amplify-js/issues/10053

_Description of changes:_
**Change:** Added a callout element with a note explaining the difference between Type and contentType in terms of uploading objects into S3. I also added an example, so users understand the implications of the note when uploading and accessing S3 files. 
**Background:** There can be confusion when users upload a file to S3 because the Type displayed in the S3 console is based on the extension of the 'key' of the uploaded file, while the contentType which determines how S3 treats the object is stored in the metadata. This means when a developer uploads a file, it could lead to confusion since Type in the S3 console may not match the contentType passed depending on the 'key's file extension. For example, this confusion could manifest as a developer thinking that a file that is uploaded with a .jpeg extension in the 'key' will be treated as a jpeg file by S3 since 'Type' will be set as 'jpeg', but this will not be the case since it will default to a contentType of 'binary/octet-stream' which means that if Storage.get was used to get a pre-signed URL for it clicking the URL would lead to an instant download instead of having the link displaying an image in the browser which should happen for files with contentType 'image/jpeg'. 
**Note:** More background information in my comment on the referenced issue and linked google docs.

### DIFF
--- a/src/fragments/lib/storage/js/upload.mdx
+++ b/src/fragments/lib/storage/js/upload.mdx
@@ -145,6 +145,14 @@ async function onChange(e) {
 <input type="file" onChange={onChange} />;
 ```
 
+<Callout>
+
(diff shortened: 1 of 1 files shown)


arundna (approved):

abdallahshaban557 (approved):
LGTM!

ashika01 (approved):

PR #4509, opened 2022-08-04
fix(Authentication): Keep social sigin-in page uniform across all pla…

ashwinkumar6:
_Description of changes:_
1. Keep social sigin page uniform across all platforms
```
`Authentication -> Social sign-in (OAuth)` is only available for Js
`Authentication -> Social sign in with web UI` is only available for iOS, Android and Flutter

Merge these under a common title
`Authentication -> Social sign-in (OAuth)` is available for all platforms (JS, iOS, Android and Flutter)

The following URLs are replaced
- FROM `lib/auth/social_signin_web_ui/q/platform/ios/` TO  `lib/auth/social/q/platform/ios/`
- FROM `lib/auth/social_signin_web_ui/q/platform/android/` TO  `lib/auth/social/q/platform/android/`
- FROM `lib/auth/social_signin_web_ui/q/platform/flutter/` TO  `lib/auth/social/q/platform/flutter/`
```
2. Add more detailed description to  `Social sign-in (OAuth)` using JS page
Detail out the Amplify CLI  `amplify add auth` steps 

3. Mention the possibility of `amplify update auth` for iOS, Android, Flutter pages


abdallahshaban557 (approved):
LGTM! Thank you!

stocaaro (approved):

arundna (approved):

src/directory/directory.js
```diff
@@ -289,8 +289,8 @@ const directory = {
             filters: ['android', 'flutter', 'ios']
           },
           {
-            title: 'Social sign in with web UI',
```
elorzafe, line 292: I thought this was because the way is defined by mobile platforms. I would check with iOS and Android teams about this

src/fragments/lib/auth/android/social_signin_web_ui/10_cli_setup.mdx
```diff
@@ -1,4 +1,10 @@
-In terminal, navigate to your project, run `amplify add auth`, and choose the following options (the last steps are specific to Facebook here but are similar for other providers):
+Once you have the social provider configured, run the following in your project’s root folder
+
+```bash
```
elorzafe, line 3: does it look different `bash` than `terminal`

elorzafe (commented):
Thanks @ashwinkumar6 

I am a little concerned we have different formats I am seeing `console`, `bash` and `terminal` @abdallahshaban557 what do you think? Maybe is not a problem?

src/directory/directory.js
ashwinkumar6: Maybe inside the Android / iOS / Flutter pages we could mention the `web UI` key word if this is required.
Will confirm with the receptive teams for the same

src/fragments/lib/auth/android/social_signin_web_ui/10_cli_setup.mdx
ashwinkumar6: We were [previously using](https://github.com/aws-amplify/docs/blob/main/src/fragments/lib/auth/js/social.mdx?plain=1#L15) `bash` and `terminal` on the Social sign-in (OAuth) Js page, so maintained the same.

Currently looks (bash, console)
Screen Shot 2022-08-08 at 7 53 42 PM

 If both are console
Screen Shot 2022-08-08 at 7 55 39 PM
 
 If both are bash
Screen Shot 2022-08-08 at 7 58 30 PM

abdallahshaban557: I think we should just use what it previously was - since it seems like the whole docs site right now uses terminal.

src/directory/directory.js
ashwinkumar6: Checked, this shouldn't be a concern

PR #4452, opened 2022-07-09
feat(data): add DataStore 'Advanced Workflows' section; add Custom Primary Key docs

david-mcafee:
# NOTE: to view changes locally:
- `gh pr checkout 4452`
- `yarn && yarn dev`
- Navigate to Libraries > DataStore > Advanced Workflows

_Description of changes:_
feat(data): add DataStore 'Advanced Workflows' section; add Custom Primary Key docs

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4557, opened 2022-08-31
chore(api): callout delete warning in api similar to datastore

sundersc:
Amplify transformers do not support cascading delete on many to many relations and this is currently called out in datastore docs.

With this PR, adding the same warning to API category as well.

### DIFF
--- a/src/fragments/lib/graphqlapi/js/mutate-data.mdx
+++ b/src/fragments/lib/graphqlapi/js/mutate-data.mdx
@@ -63,7 +63,7 @@ const deletedTodo = await API.graphql({ query: mutations.deleteTodo, variables: 
 Only an `id` is needed.
 
 <Callout>
-In a many-to-many relationship, the children are not removed and you must explicitly delete them.
+Join table records must be deleted prior to deleting the associated records. For example, for a many-to-many relationship between Posts and Tags, delete the PostTags join record prior to deleting a Post or Tag.
 </Callout>
 
 ### Custom authorization mode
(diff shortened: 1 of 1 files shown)


src/fragments/lib/graphqlapi/js/mutate-data.mdx
```diff
@@ -62,6 +62,10 @@ const deletedTodo = await API.graphql({ query: mutations.deleteTodo, variables:
 
 Only an `id` is needed.
 
+<Callout>
```
david-mcafee, line 65: We currently plan to add this in the CPK docs, wondering if we should align the phrasing here?

https://github.com/aws-amplify/docs/pull/4452/files#diff-4d957a94c2c999b5fb4f2dcb9748be613e4a2849b5893d54d7489199f4de504fR85-R90

david-mcafee (changes requested):
Left a single comment

src/fragments/lib/graphqlapi/js/mutate-data.mdx
```diff
@@ -62,6 +62,10 @@ const deletedTodo = await API.graphql({ query: mutations.deleteTodo, variables:
...
+In a many-to-many relationship, the children are not removed and you must explicitly delete them.
```
alharris-at, line 66: Is this the exact wording we use in the datastore docs?

```suggestion
In a many-to-many relationship, the related items, and join records are not removed and you must explicitly delete them.
```

src/fragments/lib/graphqlapi/js/mutate-data.mdx
sundersc: Updated the wordings to match CPK docs PR.

src/fragments/lib/graphqlapi/js/mutate-data.mdx
sundersc: Yes, it is from datastore docs. Anyways I've updated the wordings to match CPK docs PR.

alharris-at (approved):

david-mcafee (approved):
LGTM! 🎉

PR #4595, opened 2022-09-19
fix: make LibVersionSwitcher links with '/platform' filter

5d:
_Issue #, if available:_

_Description of changes:_

Fix the generated links in `LibVersionSwitcher` with `platform` filter.
- Links with `/lib` [is with kind](https://github.com/aws-amplify/docs/blob/dev-preview-ios/generatePathMap.cjs.js#L129) `platform` in the URL. 


hdworld11 (approved):

5d:
Might be better to have version info as a parameter inside the query. 
Something like `lib/project-setup/prereq/q/platform/[platform]/version/[version]`. We can redirect to latest version when a `version` for a `platform` doesn't exist. Not sure whether this is possible in Nextjs.

phantumcode (approved):

PR #4579, opened 2022-09-12
update iOS v2 auth code snippets

phantumcode:
_Issue #, if available:_

_Description of changes:_
* Update Amplify iOS Library V2 documentation with new async/await code examples

Please checkout working branch and build `yarn dev` in order to test and view changes locally in a web browser.

Sample screenshot:
Amplify iOS V2 Auth sample

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4606, opened 2022-09-21
chore(api): Update API docs for Amplify iOS v2

thisisabhash:
_Issue #, if available:_ None

_Description of changes:_ Update GraphQL and REST API documentation for Amplify iOS v2.

### DIFF
--- a/src/fragments/lib/graphqlapi/ios/advanced-workflows/10_example.mdx
+++ b/src/fragments/lib/graphqlapi/ios/advanced-workflows/10_example.mdx
@@ -1,4 +1,4 @@
 ```swift
 let todo = Todo(name: "my first todo", description: "todo description")
-Amplify.API.mutate(request: .create(todo))
+await Amplify.API.mutate(request: .create(todo))
 ```
--- a/src/fragments/lib/graphqlapi/ios/advanced-workflows/10_example.mdx
+++ b/src/fragments/lib/graphqlapi/ios/advanced-workflows/10_example.mdx
@@ -1,4 +1,4 @@
 ```swift
 let todo = Todo(name: "my first todo", description: "todo description")
-await Amplify.API.mutate(request: .create(todo))
+try await Amplify.API.mutate(request: .create(todo))
 ```
--- a/src/fragments/lib/graphqlapi/ios/advanced-workflows/40_multiple.mdx
+++ b/src/fragments/lib/graphqlapi/ios/advanced-workflows/40_multiple.mdx
@@ -26,27 +26,25 @@ Notice here that `JSONValue` is used as the `responseType`. `JSONValue` is utili
 Once you have the response data in a `JSONValue`, you can access each object in the JSON structure by encoding it back to Data and decoding it to the expected Model.
 
 ```swift
-Amplify.API.query(request: .get(byPostId: "[POST_ID]", todoId: "[TODO_ID]")) { result in
-    switch result {
-    case .success(let response):
-        switch response {
-        case .success(let data):
-            if let todoJSON = data.value(at: "getTodo"),
-                let todoData = try? JSONEncoder().encode(todoJSON),
-                let todo = try? JSONDecoder().decode(Todo.self, from: todoData) {
-                print(todo)
-            }
-            if let postJSON = data.value(at: "getPost"),
-                let postData = try? JSONEncoder().encode(postJSON),
-                let post = try? JSONDecoder().decode(Post.self, from: postData) {
-                print(post)
-            }
-        case .failure(let errorResponse):
-            print("Response contained errors: \(errorResponse)")
+do {
+    let response = try await Amplify.API.query(request: .get(byPostId: "[POST_ID]", todoId: "[TODO_ID]"))
+    switch response {
+    case .success(let data):
+        if let todoJSON = data.value(at: "getTodo"),
+            let todoData = try? JSONEncoder().encode(todoJSON),
+            let todo = try? JSONDecoder().decode(Todo.self, from: todoData) {
+            print(todo)
         }
-    case .failure(let apiError):
-        print("Failed with error: \(apiError)")
+        if let postJSON = data.value(at: "getPost"),
+            let postData = try? JSONEncoder().encode(postJSON),
+            let post = try? JSONDecoder().decode(Post.self, from: postData) {
+            print(post)
+        }
+    case .failure(let errorResponse):
+        print("Response contained errors: \(errorResponse)")
     }
+} catch (let error) {
+    print("Failed with error: \(error)")
 }
 ```
 
--- a/src/fragments/lib/graphqlapi/ios/advanced-workflows/40_multiple.mdx
+++ b/src/fragments/lib/graphqlapi/ios/advanced-workflows/40_multiple.mdx
@@ -43,8 +43,10 @@ do {
     case .failure(let errorResponse):
         print("Response contained errors: \(errorResponse)")
     }
-} catch (let error) {
+} catch (let error as APIError) {
     print("Failed with error: \(error)")
+} catch {
+    print("Unexpected error: \(error)")
 }
 ```
 
--- a/src/fragments/lib/graphqlapi/ios/authz/30_multi.mdx
+++ b/src/fragments/lib/graphqlapi/ios/authz/30_multi.mdx
@@ -2,5 +2,5 @@ When you have configured multiple APIs, you can specify the name of the API as a
 
 ```swift
 let request = GraphQLRequest(apiName: "[FRIENDLY-NAME-API-WITH-API-KEY]", ...)
-await Amplify.API.mutate(request: request)
+try await Amplify.API.mutate(request: request)
 ```
--- a/src/fragments/lib/graphqlapi/ios/authz/30_multi.mdx
+++ b/src/fragments/lib/graphqlapi/ios/authz/30_multi.mdx
@@ -2,5 +2,5 @@ When you have configured multiple APIs, you can specify the name of the API as a
 
 ```swift
 let request = GraphQLRequest(apiName: "[FRIENDLY-NAME-API-WITH-API-KEY]", ...)
-Amplify.API.mutate(request: request, listener: ...)
+await Amplify.API.mutate(request: request)
 ```
--- a/src/fragments/lib/graphqlapi/ios/getting-started/50_createtodo.mdx
+++ b/src/fragments/lib/graphqlapi/ios/getting-started/50_createtodo.mdx
@@ -1,58 +1,17 @@
 
-<BlockSwitcher>
-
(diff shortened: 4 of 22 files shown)


src/fragments/lib/graphqlapi/ios/advanced-workflows/10_example.mdx
```diff
@@ -1,4 +1,4 @@
 ```swift
 let todo = Todo(name: "my first todo", description: "todo description")
-Amplify.API.mutate(request: .create(todo))
+await Amplify.API.mutate(request: .create(todo))
```
lawmicha, line 3: should these be `try await`?

src/fragments/lib/graphqlapi/ios/authz/30_multi.mdx
```diff
@@ -2,5 +2,5 @@ When you have configured multiple APIs, you can specify the name of the API as a
 
 ```swift
 let request = GraphQLRequest(apiName: "[FRIENDLY-NAME-API-WITH-API-KEY]", ...)
-Amplify.API.mutate(request: request, listener: ...)
+await Amplify.API.mutate(request: request)
```
lawmicha, line 5: `try await` ?

src/fragments/lib/graphqlapi/ios/getting-started/50_createtodo.mdx
```diff
@@ -1,58 +1,17 @@
...
-        case .failure(let apiError):
-            print("Failed to create a todo", apiError)
-        }
-    }
-}
-```
-
-</Block>
-
-<Block name="Combine">
```
lawmicha, line 27: I don't think we're removing the Combine versions of the API, see Auth for example https://github.com/aws-amplify/docs/pull/4579 there's "Async/Await" that replaces "Listener (iOS 11+)" and then Combine is still there. Every APi can be wrapped with `Amplify.Publisher.create` to create the Combine publisher

src/fragments/lib/graphqlapi/ios/advanced-workflows/40_multiple.mdx
```diff
@@ -26,27 +26,25 @@ Notice here that `JSONValue` is used as the `responseType`. `JSONValue` is utili
...
+        if let postJSON = data.value(at: "getPost"),
+            let postData = try? JSONEncoder().encode(postJSON),
+            let post = try? JSONDecoder().decode(Post.self, from: postData) {
+            print(post)
+        }
+    case .failure(let errorResponse):
+        print("Response contained errors: \(errorResponse)")
     }
+} catch (let error) {
+    print("Failed with error: \(error)")
```
lawmicha, line 47: should we be expanding this to two catch blocks? 
```
} catch let error as APIError  {
    print("Failed with error: \(error)")
} catch {
   print("Unexpected error: \(error)"
}
```

src/fragments/lib/graphqlapi/ios/getting-started/50_createtodo.mdx
```diff
@@ -1,58 +1,17 @@
...
-
-Make sure you have the additional import at the top of your file
-```swift
-import Combine
-```
-
 ```swift
-func createTodo() -> AnyCancellable {
+func createTodo() async {
```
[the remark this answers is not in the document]
thisisabhash: yes, in this case it is wrapped in a do-try catch at line 6 below

src/fragments/lib/graphqlapi/ios/mutate-data.mdx
```diff
@@ -48,21 +48,23 @@ func updateTodo() -> AnyCancellable {
     // Retrieve your Todo using Amplify.API.query
     var todo = Todo(name: "my first todo", description: "todo description")
     todo.description = "updated description"
-    let sink = Amplify.API.mutate(request: .update(todo))
-        .resultPublisher
-        .sink {
-            if case let .failure(error) = $0 {
-                print("Got failed event with error \(error)")
-            }
+    let todoUpdated = todo
```
lawmicha, line 51: is there a reason to create a new variable here? how about this where we clarify the "retrievedTodo" and then a `var updatedTodo` that we modify?
```
let retrievedTodo = Todo(name: "my first todo", description: "todo description")
    todo.description = "updated description")
var updatedTodo = retrievedTodo
updatedTodo.description = "updatedDescription"
```

src/fragments/lib/graphqlapi/ios/query-data.mdx
```diff
@@ -125,31 +62,32 @@ func listTodos() -> AnyCancellable {
 
 </BlockSwitcher>
 
-</Block>
+## List Query
+
+You can get the list of items using `.paginatedList` with optional parameters `limit` and `where` to specify the page size and condition. By default, the page size is 1000.
```
lawmicha, line 67: `.list` now in V2

src/fragments/lib/graphqlapi/ios/query-data.mdx
```diff
@@ -162,8 +100,7 @@ func listTodos() {
 func listTodos() -> AnyCancellable {
     let todo = Todo.keys
     let predicate = todo.name == "my first todo" && todo.description == "todo description"
-    let request = GraphQLRequest<Todo>.list(Todo.self, where: predicate, limit: 1000)
-    let sink = Amplify.API.query(request: request)
+    let sink = Amplify.API.query(request: .paginatedList(Todo.self, where: predicate, limit: 1000))
```
lawmicha, line 103: `.list`

src/fragments/lib/graphqlapi/ios/subscribe-data.mdx
```diff
@@ -27,28 +27,33 @@ To listen to creation updates, you can use the following code sample:
...
-            case .success(let createdTodo):
-                print("Successfully got todo from subscription: \(createdTodo)")
-            case .failure(let error):
-                print("Got failed result with \(error.errorDescription)")
+    var subscription: AmplifyAsyncThrowingSequence<GraphQLSubscriptionEvent<Todo>>?
+    subscription = Amplify.API.subscribe(request: .subscription(of: Todo.self, type: .onCreate))
+    guard let subscription = subscription else {
+        print("Error getting subscription")
+        return
+    }
```
lawmicha, line 39: i don't think this will ever fail to get the sequence
```
let sequence = Amplify.API.subscribe(request: .subscription(of: Todo.self, type: .onCreate))

```

src/fragments/lib/graphqlapi/ios/subscribe-data.mdx
```diff
@@ -27,28 +27,33 @@ To listen to creation updates, you can use the following code sample:
...
+                    print("Subscription connect state is \(subscriptionConnectionState)")
+                case .data(let result):
+                    switch result {
+                    case .success(let createdTodo):
+                        print("Successfully got todo from subscription: \(createdTodo)")
+                    case .failure(let error):
+                        print("Got failed result with \(error.errorDescription)")
+                    }
+                }
             }
```
lawmicha, line 54: i believe you can store the task and then use that to cancel the subscription
```
self.subscription = Task  {
 ... 
}
```

src/fragments/lib/graphqlapi/ios/subscribe-data.mdx
```diff
@@ -60,7 +65,12 @@ func createSubscription() {
 ```swift
 func createSubscription() {
     subscription = Amplify.API.subscribe(request: .subscription(of: Todo.self, type: .onCreate))
-    dataSink = subscription?.subscriptionDataPublisher.sink {
+    guard let subscription = subscription else {
+        print("Error getting subscription")
+        return
+    }
+    dataSink = Amplify.Publisher.create(subscription)
```
lawmicha, line 72: same here, `let sequence = Amplify.API.subscribe..`

lawmicha: with the above, you can then keep the name of the AnyCancellable the same as before, `var subscription: AnyCancellable?`

src/fragments/lib/graphqlapi/ios/subscribe-data.mdx
```diff
@@ -103,12 +118,10 @@ Calling `cancel()` on the subscription will disconnect the subscription from the
```
lawmicha, line 118: i think we need to verify this statement
> Calling `cancel()` on the subscription will disconnect the subscription from the backend. Any downstream subscribers will also be cancelled. On the other hand, calling `cancel()` on the Combine subscriber (e.g., the `AnyCancellable` returned from `sink()` will cause that Combine subscriber to stop receiving updates, but any other attached subscribers will continue to receive subscription data. For example:

I wonder if this is still true, if the combine subscriber gets cancelled, i think it will cancel the sequence and other attached subscribers won't continue to receive subscriptin data anymore

feel free to create a new issue for this to verify this section of the doc separately

src/fragments/lib/graphqlapi/ios/mutate-data.mdx
thisisabhash: The `async/await` block doesn't allow `toDo` to be a `var` and gives the error `Reference to captured var in a concurrently executing code`

src/fragments/lib/graphqlapi/ios/subscribe-data.mdx
```diff
@@ -4,18 +4,18 @@ Because the lifetime of the subscription will last longer than the lifetime of a
 
 <BlockSwitcher>
 
-<Block name="Listener (iOS 11+)">
+<Block name="Async/Await">
 
 ```swift
-var subscription: GraphQLSubscriptionOperation<Todo>?
+var subscription: AmplifyAsyncThrowingSequence<GraphQLSubscriptionEvent<Todo>>
```
lawmicha, line 10: this looks correct, we hold onto the sequence and can use it to call `cancel()`

src/fragments/lib/graphqlapi/ios/subscribe-data.mdx
```diff
@@ -4,18 +4,18 @@ Because the lifetime of the subscription will last longer than the lifetime of a
...
 ```
 
 </Block>
 
 <Block name="Combine">
 
 ```swift
-var subscription: GraphQLSubscriptionOperation<Todo>?
+var subscription: AmplifyAsyncThrowingSequence<GraphQLSubscriptionEvent<Todo>>
 var dataSink: AnyCancellable?
```
lawmicha, line 19: on a high level, i don't think we need the sequence anymore, only need the AnyCancellable, since that will hold onto the sequence reference. Just having this i think is enough:
```
var subscription: AnyCancellable?
```

The calling code below can create a local sequence variable, and pass that into the Amplify.Publisher.create which returns the AnyCancellable to be held by this property here

lawmicha (approved):
LGTM, thanks @thisisabhash !

PR #4641, opened 2022-10-04
fix(iOS): update reference from AWS iOS SDK to AWS SDK for Swift

phantumcode:
_Issue #, if available:_

_Description of changes:_
* Updating reference from AWS iOS SDK to AWS SDK for Swift
* Add stability note

### DIFF
--- a/src/fragments/lib/ios.mdx
+++ b/src/fragments/lib/ios.mdx
@@ -4,9 +4,11 @@ If you already have existing resources to add to your application and want to by
 
 **Amplify libraries should be used for all new cloud connected applications.** If you are currently using the AWS SDK for Swift, you can access the documentation [here](https://docs.aws.amazon.com/sdk-for-swift/latest/developer-guide/getting-started.html).
 
-**Note:** While the Amplify Library for Swift is production ready, please note that the underlying AWS SDK for Swift is currently in Developer Preview, and is not yet intended for production workloads. [Here is additional reading material](https://github.com/awslabs/aws-sdk-swift/blob/main/docs/stability.md) on the stability of the SDK
-
 </Callout>
+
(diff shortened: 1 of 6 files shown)


src/fragments/lib/ios.mdx
```diff
@@ -2,13 +2,15 @@ If you already have existing resources to add to your application and want to by
 
 <Callout warning>
 
-**Amplify libraries should be used for all new cloud connected applications.** If you are currently using the AWS Mobile SDK for iOS, you can access the documentation [here](/sdk).
+**Amplify libraries should be used for all new cloud connected applications.** If you are currently using the AWS SDK for Swift, you can access the documentation [here](https://docs.aws.amazon.com/sdk-for-swift/latest/developer-guide/getting-started.html).
+
+**Note:** While the Amplify Library for Swift is production ready, please note that the underlying AWS SDK for Swift is currently in Developer Preview, and is not yet intended for production workloads. [Here is additional reading material](https://github.com/awslabs/aws-sdk-swift/blob/main/docs/stability.md) on the stability of the SDK
```
ruisebas, line 7: Would it make sense to define this as a new fragment and just import it on all the required pages ?
This would avoid having to update every one of them should this wording change.

phantumcode: no, the escape note is a subset and only contains the 2nd paragraph.  I think the only time it would be updated is when we remove it when Swift SDK is GA.

ruisebas: Yeah I meant the second paragraph. I was trying to avoid having copy-pasted text through the categories, but if you think it's fine I don't really mind.

phantumcode: On second thought, I moved it to a fragment like you suggested!

ruisebas (approved):

PR #4642, opened 2022-10-04
Updating Analytics docs

ruisebas:
_Description of changes:_
- Creating a fragment for the recording code, to avoid duplication
- Creating an iOS specific page of using existing resources so we can omit the `pinpointTargeting` section.
- Updated Swift UI instructions to use `init` instead of `UIApplicationDelegateAdaptor`
  - Also renamed the options to just **Swift UI** and **UIKit** and swapped them.
- Fixed a compilation issue in the **Identify User** code, and changing it to use the good syntax while at it 😌


phantumcode (approved):
LGTM

PR #4664, opened 2022-10-06
Fixing Analytics `unregisterGlobalProperties` example code

ruisebas:
_Description of changes:_
Adding explicit type, otherwise Xcode assumes an array and the code doesn't compile 😒


ameter:
Perhaps we should take an array and convert it to a set internally?

PR #4670, opened 2022-10-07
fix(ios): update auth signout examples

atierian:
_Issue #, if available:_
N/A

_Description of changes:_
Updates auth signout examples to include full error handling.

### DIFF
--- a/src/fragments/lib/auth/ios/signout/10_local_signout.mdx
+++ b/src/fragments/lib/auth/ios/signout/10_local_signout.mdx
@@ -38,36 +38,29 @@ func signOutLocally() async {
 func signOutLocally() -> AnyCancellable {
     Amplify.Publisher.create {
         await Amplify.Auth.signOut()
-    }.sink(
-        receiveCompletion: {
-            if case let .failure(error) = $0 {
-                print("Sign out failed with error \(error)")
-            }
-        },
-        receiveValue: { result in
-            guard let signOutResult = result as? AWSCognitoSignOutResult
-            else {
-                print("Signout failed")
-                return
-            }
-            print("Local signout successful: \(signOutResult.signedOutLocally)")
-            switch signOutResult {
-            case .complete:
-                print("SignOut completed")
-            case .failed(let error):
-                print("SignOut failed with \(error)")
-            case let .partial(revokeTokenError, globalSignOutError, hostedUIError):
-                print(
-                    """
-                    SignOut is partial.
-                    RevokeTokenError: \(String(describing: revokeTokenError))
-                    GlobalSignOutError: \(String(describing: globalSignOutError))
-                    HostedUIError: \(String(describing: hostedUIError))
-                    """
-                )
-            }
+    }.sink(receiveValue: { result in
+        guard let signOutResult = result as? AWSCognitoSignOutResult
+        else {
+            print("Signout failed")
+            return
         }
-    )
+        print("Local signout successful: \(signOutResult.signedOutLocally)")
+        switch signOutResult {
+        case .complete:
+            print("SignOut completed")
+        case .failed(let error):
+            print("SignOut failed with \(error)")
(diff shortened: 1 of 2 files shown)


src/fragments/lib/auth/ios/signout/10_local_signout.mdx
```diff
@@ -23,14 +34,33 @@ func signOutLocally() async {
...
+    }.sink(
+        receiveCompletion: {
+            if case let .failure(authError) = $0 {
+                print("Sign out failed with error \(authError)")
+            }
+        },
+        receiveValue: { result in
+            if let signOutResult = result as? AWSCognitoSignOutResult {
+                print("Local signout successful: \(signOutResult.signedOutLocally)")
+                switch signOutResult {
```
lawmicha, line 46: nit: should we use guard instead? for the Combine example, it's getting a few indentations deep. if we do, then should use it for the Async/Await example as well

lawmicha (approved):

src/fragments/lib/auth/ios/signout/10_local_signout.mdx
```diff
@@ -5,13 +5,24 @@
 ```swift
 func signOutLocally() async {
     let result = await Amplify.Auth.signOut()
-    guard let signOutResult = result as? AWSCognitoSignOutResult,
-          case .complete = signOutResult
-    else {
-        print("Sign out failed to complete")
-        return
+    if let signOutResult = result as? AWSCognitoSignOutResult {
+        print("Local signout successful: \(signOutResult.signedOutLocally)")
```
harsh62, line 9: I guess the print statement is incorrect.. The print statement won't be true when the `signOutResult` is failed. 

this is our implementation: 

```
public enum AWSCognitoSignOutResult: AuthSignOutResult {

    var signedOutLocally: Bool {
        if case .failed = self {
            return false
        }
        return true
    }
```

src/fragments/lib/auth/ios/signout/10_local_signout.mdx
```diff
@@ -23,14 +34,33 @@ func signOutLocally() async {
...
     Amplify.Publisher.create {
         await Amplify.Auth.signOut()
-    }.sink {
-        if case let .failure(authError) = $0 {
-            print("Sign out failed with error \(authError)")
+    }.sink(
+        receiveCompletion: {
+            if case let .failure(authError) = $0 {
+                print("Sign out failed with error \(authError)")
+            }
```
harsh62, line 41: wondering this is still valid? when would we get this failure?

src/fragments/lib/auth/ios/signout/10_local_signout.mdx
atierian: That would result in the following print statements, right?
```Local signout successful: true``` if it succeeds.
```Local signout successful: false``` if it fails.

That seems fine, but I'm open to other suggestions if you have one.

src/fragments/lib/auth/ios/signout/10_local_signout.mdx
atierian: good call. updated in [3cf480a](https://github.com/aws-amplify/docs/pull/4670/commits/3cf480a4e551f4d0aec6a611934e8e850c8fd07c)

src/fragments/lib/auth/ios/signout/10_local_signout.mdx
atierian: Good catch. It appears that the `receiveCompletion` block here is unnecessary, but required because `Amplify.Publisher.create` returns `AnyPublisher<_, Error>` even when a non-throwing function is provided as its argument. I updated the code snippets in 6af0cef and made a [PR in Amplify](https://github.com/aws-amplify/amplify-ios/pull/2414) to add a non-throwing overload of `create(_:)`.

PR #4708, opened 2022-10-14
quick fix amplify android getting-started

dengdan154:
_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/src/fragments/start/getting-started/android/setup.mdx
+++ b/src/fragments/start/getting-started/android/setup.mdx
@@ -52,6 +52,9 @@ Amplify for Android is distributed as an Apache Maven package. In this section, 
    dependencies {
        implementation 'com.amplifyframework:aws-api:ANDROID_VERSION'
        implementation 'com.amplifyframework:aws-datastore:ANDROID_VERSION'
+       def lifecycle_version = "2.4.0"
+       implementation "androidx.lifecycle:lifecycle-viewmodel:$lifecycle_version"
+       implementation "androidx.lifecycle:lifecycle-viewmodel-ktx:$lifecycle_version"
    }
    ```
 


src/fragments/start/getting-started/android/setup.mdx
```diff
@@ -52,6 +52,9 @@ Amplify for Android is distributed as an Apache Maven package. In this section,
    dependencies {
        implementation 'com.amplifyframework:aws-api:ANDROID_VERSION'
        implementation 'com.amplifyframework:aws-datastore:ANDROID_VERSION'
+       def lifecycle_version = "2.4.0"
```
sktimalsina, line 55: Do we need to fix customers to this particular lifecycle_version? if not, we can leave it as variable similar as `ANDROID_VERSION` above.

src/fragments/start/getting-started/android/setup.mdx
```diff
@@ -52,6 +52,9 @@ Amplify for Android is distributed as an Apache Maven package. In this section,
...
+       implementation "androidx.lifecycle:lifecycle-viewmodel:$lifecycle_version"
+       implementation "androidx.lifecycle:lifecycle-viewmodel-ktx:$lifecycle_version"
```
salihgueler, line 57: IMO putting the lifecycle version inside the libraries would be more consistent. People can create the version def at some other place like they do in their projects. 

In the Android Codelabs they use the approach below only when they define the root project external version numbers [here](https://developer.android.com/codelabs/android-room-with-a-view?hl=en#3):
`implementation "androidx.lifecycle:lifecycle-viewmodel:$rootProject.lifecycleVersion"`

```suggestion
       implementation "androidx.lifecycle:lifecycle-viewmodel:2.4.0"
       implementation "androidx.lifecycle:lifecycle-viewmodel-ktx:2.4.0"
```

src/fragments/start/getting-started/android/setup.mdx
dengdan154: I went ahead and committed @salihgueler suggested change, I don't think there's a variable for lifecycle_version similar to ANDROID_VERSION

salihgueler (dismissed):

div5yesh (dismissed):

tjleing (approved):

div5yesh (approved):

tjleing (approved):

PR #4723, opened 2022-10-20
Android v1/v2 docs

tjleing:
_Issue #, if available:_

_Description of changes:_


public/robots.txt
```diff
@@ -1,2 +1,2 @@
 User-agent: *
-Disallow:
+Disallow: /
```
timngyn, line 2: I think we could probably just remove this file since it's in our `.gitignore`. Since this file gets created during the build though, it should be okay to leave it in like this

Otherwise if you can, could we do a `git rm` of it?

timngyn (dismissed):
Looks good to me! Only checked out non `.mdx` files. Left a comment about the `robots.txt` file, but it's not blocking

timngyn (approved):

sdhuka (approved):

Issue #4729, opened 2022-10-20
? getting encoded as %3F


dengdan154:
**Describe the bug**
Getting a 404 page not found exception because the ? in URL link is getting encoded to %3F

**To Reproduce**

Go to https://docs.amplify.aws/
Select the Swift icon (Or any other icon) below “Get Started for Free” 
Now, select the platform picker and choose “Android” (Or any other icon)
It says page doesn’t exist

**Expected behavior**
the url should have ? instead of %3F

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #4730, opened 2022-10-20
fix: ignore queries while switching platform

5d:
_Issue #, if available:_

https://github.com/aws-amplify/docs/issues/4729

_Description of changes:_

- extract current pathname from `window.location`

### DIFF
--- a/src/components/Menu/FilterSelect/index.tsx
+++ b/src/components/Menu/FilterSelect/index.tsx
@@ -28,7 +28,7 @@ const getFirstPathSegment = (path: string): string | undefined => {
 const multiLibVersionPlatforms = ['ios']
 
 const convertToRouteHerf = (filter: FilterSelectProps, targetFilterKey: string) => {
-  let path = filter.url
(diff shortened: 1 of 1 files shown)


src/components/Menu/FilterSelect/index.tsx
```diff
@@ -28,7 +28,7 @@ const getFirstPathSegment = (path: string): string | undefined => {
 const multiLibVersionPlatforms = ['ios']
 
 const convertToRouteHerf = (filter: FilterSelectProps, targetFilterKey: string) => {
-  let path = filter.url
+  let path = window.location.pathname
```
jacoblogan, line 31: we can't use window with SSR, how about we just strip out the query parameters
```
// remove search query string when switching between platforms/frameworks
  let path =
    filter.url && filter.url.split ? filter.url.split('?')[0] : filter.url;
```

jacoblogan (approved):
LGTM

katiegoines (approved):

Issue #4732, opened 2022-10-20
Add secondary nav to docs home page


rachelnabors:
As a person new to Amplify or returning to the docs for reference material, I would like to be able to quickly navigate to Guides, CLI, etc from the landing page at docs.amplify.aws. But that page doesn't have the navigation bar to those things:

<img width="1209" alt="Screen Shot 2022-10-20 at 2 56 51 PM" src="https://user-images.githubusercontent.com/236306/197066960-4862733e-2d43-4ad8-adf7-e788ff34a122.png">

I have no choice but to click on one of the links on the page, then use that page's secondary navigation:
<img width="1215" alt="Screen Shot 2022-10-20 at 2 57 00 PM" src="https://user-images.githubusercontent.com/236306/197066903-048298a6-0456-4a47-acae-a18851e00be2.png">

This makes the docs hard to navigate. It would be better if the landing page had the secondary navigation present

oscarcao:
I'll look into this first issue.

PR #4761, opened 2022-10-28
Add secondary nav to docs home page

oscarcao:
Issue #4732

Description of changes: Add secondary nav to docs home page


rachelnabors (dismissed):
Thanks so much! I approve of this, but want to get @timngyn's review as well

timngyn (changes requested):
Hi @oscarcao, thank you so much for taking the time to do this! This looks really good so far!

One change I would like to see is for the `<Hero>` component to also have extra `margin-top` when the page is on mobile view. At the moment it looks like this when screen width is less than `975px`: 
image

At this width, the secondary nav moves the search bar into its own row and so the spacing between the nav and "Amplify Documentation" gets covered up.

My recommendation would be to add a new media query inside the `Hero` styled component from the`/components/Hero/styles.tsx` file so that there is extra margin when the max-width is `975px`. If you have any questions about this, please feel free to ask

src/pages/index.tsx
```diff
@@ -78,8 +78,8 @@ const Page = () => {
         socialLinks={SOCIAL_LINKS as NavMenuItem[]}
         currentSite={'Docs'}
       />
-      <UniversalNav blend={true} />
-      <Hero>
+      <SecondaryNav />
+      <Hero className="margin-top-lg">
```
timngyn, line 82: Hi @oscarcao, I'm not sure if this is the best way to do it as now we're adding extra margin to this component on all screen sizes. I think it would be better if we add a media query to the `Hero` styled component instead. The width when things start to change is `975px`. Is this something you are okay with changing or did you want me to do it?

timngyn (changes requested):

src/pages/index.tsx
rachelnabors: @oscarcao if you could fix this up, we can accept it!

oscarcao: Hi @timngyn and @rachelnabors

Have you guys had a look at how it looks on desktop with the added spacing at the top? I think you guys would prefer it with the extra spacing.

Cheers
Oscar

timngyn: Hi @oscarcao, thanks for the response, we did take a look and really appreciate your work on this! 

The issue I was trying to point out is that the spacing is inconsistent when the navbar switches to a mobile view and the desktop view. For instance, here in the video, I'm adjusting the width down and we can see that the "Amplify Documentation" title is jumping up and down even though the height of the navbar and the secondary nav are still the same.

https://user-images.githubusercontent.com/54393192/199626608-a49b2393-5341-4be7-acdd-e176f2e8b532.mov

I am okay with the extra spacing on desktop, but in my opinion it should be consistent between the different screen sizes. If you want to keep this extra margin then I think we can still add a media query to include this extra spacing like I mentioned earlier

oscarcao: No worries. Have updated implementation so the hero section does not jump around when break points change.

timngyn (approved):
LGTM!

rachelnabors (approved):
Thanks so much!!

PR #4765, opened 2022-10-28
fix(data): correct spelling errors

david-mcafee:
_Issue #, if available:_

_Description of changes:_


rachelnabors (commented):
Turns out "full-stack" isn't as popular a search term as "fullstack." I suspect that while the hyphen is technically correct, developers are too dang busy for that extra keystroke! We defer to popular convention. 

Screen Shot 2022-10-28 at 3 52 27 PM

PR #4766, opened 2022-10-28
Removing hyphens from 'full-stack' 

rachelnabors:
People search on "fullstack" more than "full-stack." Let's be sure to stick to the popular convention. I'll add it to our style guides, but this PR also flags it in CSPELL.

Could be _even better_ if CSPELL could flag "full stack" but I don't know if it's possible for it to look at word pairs


katiegoines (approved):
Thanks!

renebrandel (approved):

PR #4739, opened 2022-10-21
chore(auth): Add authflow type documentation for ios

royjit:
_Issue #, if available:_

_Description of changes:_ Add description about different auth flow types supported by iOS Auth Plugin

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4768, opened 2022-10-31
feat(Auth): Adding documentation for federation and switching auth flow type

harsh62:
_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/src/directory/directory.js
+++ b/src/directory/directory.js
@@ -289,6 +289,11 @@ const directory = {
             filters: ['android', 'flutter', 'ios']
           },
           {
+            title: 'Switching authentication flows',
+            route: '/lib/auth/switch-auth',
+            filters: ['ios']
+          },
+          {
             title: 'Sign in with custom flow',
             route: '/lib/auth/signin_with_custom_flow',
             filters: ['ios', 'flutter']
@@ -362,6 +367,11 @@ const directory = {
             title: 'Escape hatch',
             route: '/lib/auth/escapehatch',
             filters: ['android', 'ios']
+          },
+          {
+            title: 'Advanced workflows',
+            route: '/lib/auth/advanced',
+            filters: ['ios']
           },
           {
             title: 'Under the hood',
--- a/src/fragments/lib/auth/ios/advanced/advanced.mdx
+++ b/src/fragments/lib/auth/ios/advanced/advanced.mdx
@@ -1,0 +1,87 @@
+## Subscribing Events
+
+You can take specific actions when users sign-in or sign-out by subscribing authentication events in your app. Please see our [Hub Module Developer Guide](/lib/utilities/hub) for more information.
+
+## Identity Pool Federation
+
+You can alternatively use `federateToIdentityPool` to get AWS credentials directly from Cognito Federated Identities and not use User Pool federation. If you have logged in with `Auth.signIn` you **can not** call `federateToIdentityPool` as Amplify will perform this federation automatically for you in the background. In general, you should only call `Auth.federatedSignIn()` when using OAuth flows. 
+
+You can use the escape hatch API `federateToIdentityPool` with a valid token from other social providers.
+
+```swift
+func federateToIdentityPools() async throws {
+    guard let authCognitoPlugin = try Amplify.Auth.getPlugin(
+        for: "awsCognitoAuthPlugin") as? AWSCognitoAuthPlugin else {
+        fatalError("Unable to get the Auth plugin")
+    }
+    do {
+        let result = try await authCognitoPlugin.federateToIdentityPool(
+            withProviderToken: "YOUR_TOKEN", for: .facebook)
+        print("Federation successful with result: \(result)")
+    } catch {
+        print("Failed to federate to identity pools with error: \(error)")
+    }
+}
+```
+
+<Callout>
+Note that when federated, API's such as Auth.getCurrentUser() will throw an error as the user is not authenticated with User Pools.
+</Callout>
+
+### Retrieve Session
+
+After federated login, you can retrieve session using the `Auth.fetchAuthSession` API.
+
+### Token Refresh
+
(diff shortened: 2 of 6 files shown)


src/directory/directory.js
```diff
@@ -288,6 +288,11 @@ const directory = {
             route: '/lib/auth/signin',
             filters: ['android', 'flutter', 'ios']
           },
+          {
```
abdallahshaban557, line 291: can you please use the existing Switching authentication flows directory entry? Rather than creating another one just for iOS?

src/directory/directory.js
```diff
@@ -363,6 +368,11 @@ const directory = {
             route: '/lib/auth/escapehatch',
             filters: ['android', 'ios']
           },
+          {
```
abdallahshaban557, line 371: Same thing here for advanced workflow?

src/fragments/lib/auth/ios/advanced/advanced.mdx
```diff
@@ -0,0 +1,87 @@
+## Subscribing Events
+
+You can take specific actions when users sign-in or sign-out by subscribing authentication events in your app. Please see our [Hub Module Developer Guide](/lib/utilities/hub) for more information.
+
+## Identity Pool Federation
+
+You can alternatively use `federateToIdentityPool` to get AWS credentials directly from Cognito Federated Identities and not use User Pool federation. If you have logged in with `Auth.signIn` you **can not** call `federateToIdentityPool` as Amplify will perform this federation automatically for you in the background. In general, you should only call `Auth.federatedSignIn()` when using OAuth flows.
```
abdallahshaban557, line 7: can you please explain the use case here? rather than the technical implementation details? Why would a customer want to use Identity pool federation?

src/fragments/lib/auth/ios/advanced/advanced.mdx
```diff
@@ -0,0 +1,87 @@
...
+### Retrieve Session
+
+After federated login, you can retrieve session using the `Auth.fetchAuthSession` API.
+
+### Token Refresh
+
+By default, Amplify will **NOT** automatically refresh the tokens from the federated providers. You will need to handle the token refresh logic and provide the new token to the `federateToIdentityPool` API.
+
+<Callout>
+NOTE: Automatic token refresh is supported when used with Cognito User pool.
```
abdallahshaban557, line 40: can you please move this callout right under the token refresh header? Also, can you change it to "Automatic token refresh is not automatically supported with federated sign in?" - or something like that

src/fragments/lib/auth/ios/switch-auth/switch-auth.mdx
```diff
@@ -0,0 +1,84 @@
...
+    /// Lambda if it doesn't find the user name in the user pool.
+    case userPassword
+}
+```
+
+> For more information about authentication flows, please visit [AWS Cognito developer documentation](https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-authentication-flow.html#amazon-cognito-user-pools-custom-authentication-flow)
+
+## USER_PASSWORD_AUTH flow
+
+A use case for the `USER_PASSWORD_AUTH` authentication flow is migrating users into Amazon Cognito
```
abdallahshaban557, line 42: can you elaborate on that?

abdallahshaban557 (changes requested):
Left some feedback

src/fragments/lib/auth/ios/switch-auth/switch-auth.mdx
```diff
@@ -0,0 +1,84 @@
+
+For client side authentication there are three different flows:
```
royjit, line 2: I actually had a different PR for this - https://github.com/aws-amplify/docs/pull/4739 . But yours have more information, I think I should close my PR.  Can you check if you can address Tim's comment in this PR as well.

src/fragments/lib/auth/ios/advanced/advanced.mdx
```diff
@@ -0,0 +1,87 @@
...
+        try await authCognitoPlugin.clearFederationToIdentityPool()
+        print("Federation cleared successful")
+    } catch {
+        print("Clear federation failed with error: \(error)")
+    }
+}
+```
+
+<Callout>
+Clear federation will clear the session from the device locally and will NOT clear the session from the Cognito Identity Pool or the federated provider.
```
royjit, line 63: "will NOT clear the session from the Cognito Identity Pool" - not sure what you meant here? Maybe word it like this: `clearFederationToIdentityPool` will only clear the session from local cache, developer need to handle signing out from the federated provider.

src/fragments/lib/auth/ios/switch-auth/switch-auth.mdx
```diff
@@ -0,0 +1,84 @@
...
+            password: password,
+            options: AuthSignInRequest.Options(pluginOptions: option))
+        print("Sign in succeeded with result: \(result)")
+    } catch {
+        print("Failed to sign in with error: \(error)")
+    }
+}
+```
+
+### Setup auth backend
```
royjit, line 60: We should add a section on how this works when we have auth type in amplifyconfiguration.json

src/directory/directory.js
harsh62: If I use the existing directory entry, it messes up the order.. For example: 

`Advance workflows` would show up before `Sign in` which we don't want.

src/directory/directory.js
harsh62: Similar is the case here.. If I use the existing one, Switch Auth would be the first option whereas we want it to come after sign in..

src/fragments/lib/auth/ios/switch-auth/switch-auth.mdx
harsh62: I can do that as a follow up

abdallahshaban557 (approved):
LGTM

royjit (approved):

abdallahshaban557 (approved):
LGTM

tjleing (approved):

PR #4772, opened 2022-11-01
New method of initializing Analytics

tjleing:
passing `(this)` was used before for auto session tracking -- no longer required because we get the `Application` from a different source (plugin initialization)


abdallahshaban557 (approved):
LGTM!

PR #4790, opened 2022-11-07
chore: Clean up lingering usage of deprecated default exports.

jimblanc:
_Issue #, if available:_

_Description of changes:_
This change cleans up lingering usage of deprecated default exports from the `amplify-js` repo, which will be fully removed for the upcoming v5 release.

### DIFF
--- a/src/fragments/ui-legacy/vue/overview.mdx
+++ b/src/fragments/ui-legacy/vue/overview.mdx
@@ -14,7 +14,8 @@ npm i aws-amplify-vue
 Then, alter main.js:
 
 ```javascript
-import Amplify, * as AmplifyModules from 'aws-amplify'
+import { Amplify } from 'aws-amplify'
+import * as AmplifyModules from 'aws-amplify'
 import { AmplifyPlugin } from 'aws-amplify-vue'
 import awsconfig from './aws-exports'
 Amplify.configure(awsconfig)
--- a/src/fragments/lib-v1/predictions/js/getting-started.mdx
+++ b/src/fragments/lib-v1/predictions/js/getting-started.mdx
@@ -26,7 +26,8 @@ Import and load the configuration file in your app. It's recommended you add the
 
 ```javascript
 import { Amplify } from 'aws-amplify';
-import Predictions, {
+import {
+  Predictions,
   AmazonAIPredictionsProvider
 } from '@aws-amplify/predictions';
 import awsconfig from './aws-exports';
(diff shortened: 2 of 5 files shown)


src/fragments/ui-legacy/vue/overview.mdx
```diff
@@ -14,7 +14,8 @@ npm i aws-amplify-vue
 Then, alter main.js:
 
 ```javascript
-import Amplify, * as AmplifyModules from 'aws-amplify'
+import { Amplify } from 'aws-amplify'
+import * as AmplifyModules from 'aws-amplify'
 import { AmplifyPlugin } from 'aws-amplify-vue'
 import awsconfig from './aws-exports'
 Amplify.configure(awsconfig)
```
cshfang, line 21: Nit
```suggestion
import * as AmplifyModules from 'aws-amplify'
import { AmplifyPlugin } from 'aws-amplify-vue'
import awsconfig from './aws-exports'

AmplifyModules.Amplify.configure(awsconfig)
```

cshfang (approved):

src/fragments/ui-legacy/vue/overview.mdx
jimblanc: Yeah I went back and forth on this one, seemed cleaner to just have two imports (to have the standard `Amplify.configure` call) but willing to change if you have strong opinions about it.

abdallahshaban557: why do we need both? what does the * as AmplifyModules accomplish here?

abdallahshaban557 (commented):
left a question

src/fragments/ui-legacy/vue/overview.mdx
jimblanc: So with default exports you're able to import the default and also star import the rest of the module, e.g. the original code: `import Amplify, * as AmplifyModules from 'aws-amplify'` 

This really isn't possible with named exports, so something like this is not valid for example: `import { Amplify }, * as AmplifyModules from 'aws-amplify'`. We can do what Chris suggested, but I personally thought having `AmplifyModules.Amplify.configure(awsconfig)` felt weirder than just have two imports so we can use the standard `Amplify.configure` form which is used everywhere else. Can change though if folks feel strongly about it.

abdallahshaban557 (approved):
LGTM!

tannerabread (approved):
LGTM

PR #4792, opened 2022-11-07
docs(storage): Update list pagination content to match the published …

stocaaro:
_Description of changes:_

Update Storage `lib-v1` to be inline with the `lib` updates. Fix all instances of `maxKeys` to be `pageSize`. Update `pageToken` to be `nextToken` to correspond with the API fix posted earlier.

### DIFF
--- a/src/fragments/lib-v1/storage/js/list.mdx
+++ b/src/fragments/lib-v1/storage/js/list.mdx
@@ -96,18 +96,44 @@ Storage.list('photos/', { level: 'private' })
     .catch(err => console.log(err));
 ```
 
-## Access all files 
+## Access all files
 
-To get a list of all files in your S3 bucket under a specific prefix, you can set maxKeys to ALL. The default value of maxKeys is 1000.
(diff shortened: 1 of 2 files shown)


AllanZhengYP (approved):

src/fragments/lib-v1/storage/js/list.mdx
```diff
@@ -96,18 +96,44 @@ Storage.list('photos/', { level: 'private' })
...
-.catch(err => console.log(err));;
+Storage.list('photos/', { pageSize : 'ALL' })
+  .then(({ results }) => console.log(result))
+  .catch((err) => console.log(err));
+
+```
+
+## Paginated file access
+
+If the the pageSize is set to lower than total file size, a single `Storage.list` call only returns a subset of all the files. To list all the files with multiple calls, users can use the `hasNextToken` and `nextToken` key:
```
jimblanc, line 113: `the the` -> `the`
`set to lower` -> `set lower`
`than total file size` -> `than the total file size`
`key` -> `keys`

jimblanc (commented):
Some grammar nit-picks, otherwise looks good

PR #4795, opened 2022-11-07
[DO NOT MERGE UNTIL V5 IS RELEASED] Remove legacy UI

abdallahshaban557:
_Issue #, if available:_

Remove legacy UI

_Description of changes:_


rachelnabors:
Nice! This should also include a modification to our redirect file to point
the old pages to the appropriate UI docs pages.

abdallahshaban557:
> Nice! This should also include a modification to our redirect file to point the old pages to the appropriate UI docs pages.

@rachelnabors I think the UI packages were actually hidden from search for quite a while. This PR cleans up all the old files that are just there that are not accessible anymore. How would you recommend the redirects be setup?

src/components/Menu/VersionSwitcher/index.tsx
```diff
@@ -2,30 +2,6 @@ import Link from "next/link";
 import {ActiveSwitchStyle, SwitchStyle} from "./styles";
 import directory from "../../../directory/directory";
 
-const ui = directory["ui"].items;
-const uiLegacy = directory["ui-legacy"].items;
-const uiLegacyPaths = [];
-const uiPaths = [];
```
jimblanc, line 8: Is this safe to remove? Used below on line 58

jimblanc (commented):
Thanks for cleaning these up!

src/components/Menu/VersionSwitcher/index.tsx
abdallahshaban557: let me check that

abdallahshaban557: @timngyn  - can you help me with checking what do I need to remove from this file specifically? This is looking more surgical to me.

timngyn: Just made a new PR here with the changes I think we should remove: https://github.com/aws-amplify/docs/pull/4855/files

Since we're removing the UI Library pages completely we can just remove the `VersionSwitcher` component because that was hardcoded to only the UI Library pages and was not being used anywhere else

abdallahshaban557: merged now - thank you, Tim!

Milan-Shah (approved):

jacoblogan:
Removing the references to `ui-legacy` pages within `generatePathMap.cjs.js` should unblock the build.

reesscot:
Please add redirects from legacy ui paths to https://ui.docs.amplify.aws/

katiegoines (approved):

abdallahshaban557:
LGTM! Thank you for getting this through!

dbanksdesign (approved):

PR #4801, opened 2022-11-08
feat: document enabled feature flag, and usage in secondary index section of data modeling page

alharris-at:
_Issue #, if available:_
N/A

_Description of changes:_

* Update docs for secondary indexes, feature flag default behavior, and provide examples of usage in data modeling page.
* Relevant CLI PR: https://github.com/aws-amplify/amplify-cli/pull/11348

### DIFF
--- a/src/pages/cli/graphql/data-modeling.mdx
+++ b/src/pages/cli/graphql/data-modeling.mdx
@@ -87,11 +87,36 @@ type Customer @model {
   id: ID!
   name: String!
   phoneNumber: String
-  accountRepresentativeID: ID! @index(name: "byRepresentative", queryField: "customerByRepresentative")
+  accountRepresentativeID: ID! @index
 }
 ```
 
-The example client query below allows you to query for "Customer" based on their `accountRepresentativeID`:
+The example client query below allows you to query for "Customer" records based on their `accountRepresentativeID`:
+
+```graphql
+query QueryCustomersForAccountRepresentative($accountRepresentativeID: ID!) {
+  customersByAccountRepresentativeID(accountRepresentativeID: $accountRepresentativeID) {
+    customers {
+      items {
+        id
+        name
+        phoneNumber
+      }
+    }
+  }
+}
+```
+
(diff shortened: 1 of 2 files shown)


src/pages/cli/graphql/data-modeling.mdx
```diff
@@ -132,6 +132,37 @@ query MyQuery {
 }
 ```
 
+Beginning in Amplify CLI version 8.5.0, there is a feature flag named `enableautoindexquerynames`. If enabled, this flag allows configuration of secondary indexes without any explicit naming.
```
renebrandel, line 135: We don't usually put CLI version information into the "guides" section of the docs. The right place for this is in: https://docs.amplify.aws/cli/migration/* path. Also isn't the feature flag enabled by default for new apps? For customers without the feature flag, we should have the client print a warning rather to explain the gap in the FF.

src/pages/cli/graphql/data-modeling.mdx
```diff
@@ -132,6 +132,37 @@ query MyQuery {
...
+
+For more information on feature flags, refer to the [feature flags reference](/cli/reference/feature-flags/).
```
renebrandel, line 137: Given that this is the base case moving forward, we should start the docs with this rather than end on it. Specifying the index name explicitly should be the "advanced" case and follow after. (See similar paradigm with relationships setup.)

src/pages/cli/graphql/data-modeling.mdx
alharris-at: That's correct, it will be once we merge in the related CLI PR. I'll open an AI to add the necessary validation check and warn in the transformer. Instead of adding a migration guide.

src/pages/cli/graphql/data-modeling.mdx
alharris-at: Will do!

josefaidt (approved):
This LGTM! 🚢

jhockett (dismissed):

jhockett (approved):

katiegoines (approved):
lgtm

Issue #4384, opened 2022-06-10
Algolia search isn't returning some results


timngyn:
**Describe the bug**
When searching for `CUSTOM_AUTH`, there are no results returning for this page: https://docs.amplify.aws/lib/auth/switch-auth/q/platform/js/#customauth-flow

**To Reproduce**
1. Go to https://docs.amplify.aws/
2. Click on the search box and type `CUSTOM_AUTH`
3. Verify that there are no results pointing to this page: https://docs.amplify.aws/lib/auth/switch-auth/q/platform/js/#customauth-flow

**Expected behavior**
After searching for `CUSTOM_AUTH`, I expect to see a search result pointing to the page https://docs.amplify.aws/lib/auth/switch-auth/q/platform/js/#customauth-flow

**Screenshots**
![image](https://user-images.githubusercontent.com/54393192/173134349-9b34e513-0852-405a-9ebf-8653cae2cca6.png)

timngyn:
There are some other pages that aren't being indexed as well. For example, when searching for `Hub`, we would expect to find this page: https://docs.amplify.aws/lib/utilities/hub/q/platform/js/

PR #4826, opened 2022-11-11
fix bug so that top level fragments are included in search index

jacoblogan:
_Issue:_ #4384

_Description of changes:_
This small PR fixes a bug that was stopping all top level fragments from being indexed.  Once live top level fragments will be indexed.  This issue can be seen when searching on the docs site for "use existing amazon cognito resources" which currently only shows 2 search results but omits all of the /lib resources 

Current Results
Screen Shot 2022-11-11 at 12 25 39 PM

Example of a missing result
Screen Shot 2022-11-11 at 12 26 41 PM

Another PR is on its way which will include the ability to index nested fragments

### DIFF
--- a/tasks/build-algolia-search.mjs
+++ b/tasks/build-algolia-search.mjs
@@ -199,7 +199,7 @@ async function tryParseImports(
     filename = filename.split('src/pages')[1];
     filename = filename.split('.mdx')[0];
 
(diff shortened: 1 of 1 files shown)


timngyn (approved):
Nice catch!! LGTM

katiegoines (approved):
LGTM

PR #4855, opened 2022-11-16
Remove version switcher that was used for the UI page

timngyn:
_Description of changes:_
- `VersionSwitcher` component was only used for the UI Library but since these pages are getting removed, we can just remove the component


abdallahshaban557 (approved):
LGTM

PR #4576, opened 2022-09-12
chore: remove obselete Node version update page

josefaidt:
_Issue #, if available:_

_Description of changes:_

removes outdated "Node version update" doc for Lambda functions.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4891, opened 2022-11-24
chore: move Lambda config page from guides to cli functions

ykethan:
_Issue #, if available:_ https://github.com/aws-amplify/amplify-cli/issues/10894

_Description of changes:_
Moves the `Configuring Lambda function settings` from Guides to Functions category. The page provides information on configuring a Lambda function options such as runtime and memory. 
Page will be the new point of reference as the older page has been removed on https://github.com/aws-amplify/docs/pull/4576 

### DIFF
--- a/src/directory/directory.js
+++ b/src/directory/directory.js
@@ -286,7 +286,7 @@ const directory = {
           {
             title: 'Sign in with custom flow',
             route: '/lib/auth/signin_with_custom_flow',
-            filters: ['ios', 'android','flutter']
+            filters: ['ios', 'android', 'flutter']
           },
           {
             title: 'Sign in with web UI',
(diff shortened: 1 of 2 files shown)


src/directory/directory.js
```diff
@@ -1788,6 +1788,11 @@ const directory = {
             title: 'Build options',
             route: '/cli/function/build-options',
             filters: []
+          },
+          {
+            title: 'Configuring Lambda function settings',
```
josefaidt, line 1793: This seems like a long title name, does this render properly? Is "Configure Options" too vague? 

Additionally, we should aim to keep the filename and titles similar

josefaidt: After I posted this I noticed `'Calling DynamoDB using AWS Cognito triggers'` below

josefaidt (approved):
LGTM!

PR #4911, opened 2022-12-02
feat: enable raw HTML strings in sidenav links

josefaidt:
_Issue #, if available:_

_Description of changes:_

This PR aims to enable the usage of raw HTML strings inside `directory.js` in order to render inline code blocks inside sidenav link items. 

It introduces a use of `dangerouslySetInnerHTML` in the DirectoryGroup class component, however it first validates whether the supplied string contains an inline code block.


josefaidt:
this causes unexpected behavior with meta titles, page titles, and prev/next button titles

PR #4919, opened 2022-12-05
fix: h4 styles in content, fix margin in sidenav

josefaidt:
_Issue #, if available:_

_Description of changes:_

it was found that h4's didn't have any margin applied, which was very noticeable in content pages. When I added margin, I noticed the left sidenav uses h4's which the new margin now impacted the display, which has also been corrected.

final change is to fix h4's used in a content page as they do not have the same styles as h2's and h3's

P.S. we should look into adjusting the sidenav to use an unordered list without headings


josefaidt:
closing in favor of #4928

PR #4928, opened 2022-12-07
fix: h4 styles in content, fix margin in sidenav

josefaidt:
_Issue #, if available:_

_Description of changes:_

it was found that h4's didn't have any margin applied, which was very noticeable in content pages. When I added margin, I noticed the left sidenav uses h4's which the new margin now impacted the display, which has also been corrected.

final change is to fix h4's used in a content page as they do not have the same styles as h2's and h3's

P.S. we should look into adjusting the sidenav to use an unordered list without headings

### DIFF
--- a/src/components/Menu/Directory/styles.tsx
+++ b/src/components/Menu/Directory/styles.tsx
@@ -14,6 +14,8 @@ export const DirectoryGroupHeaderStyle = styled.button`
 
   h4 {
     cursor: pointer;
+    margin-bottom: 0;
+    margin-top: 0;
   }
 
   &:hover {
--- a/src/components/NextPrevious/styles.tsx
+++ b/src/components/NextPrevious/styles.tsx
@@ -26,6 +26,7 @@ export const NextPreviousLinkStyle = styled.a`
   }
 
   h4 {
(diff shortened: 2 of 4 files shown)


timngyn:
@josefaidt, are there certain pages where I can see the differences with your branch and the prod site? So far I noticed that the right TOC have the extra margins since they are `h4` tags

I also noticed that the cards on the home page also have increased margins cause the cards use `h4` tags for the title too: https://fix-h4-margin-styles.d3kzzkiugubm89.amplifyapp.com/

Do you think we could move the margins from `elements.css` into `Page/styles.tsx` so it gets isolated to only the page content?

timngyn:
@josefaidt, thank you for refactoring it into the `Page/styles.tsx`. I was looking at this page, https://fix-h4-margin-styles.d3kzzkiugubm89.amplifyapp.com/guides/api-graphql/image-and-file-uploads/q/platform/js/#how-it-all-works, as I think it has some `h4` headings in the content. Should these have the hover effect applied to it or should it be for `h4` headings somewhere else?

katiegoines (approved):

PR #4939, opened 2022-12-09
fix: android getting started for react native push notifications

tannerabread:
_Issue #, if available:_

_Description of changes:_
This fix removes step 8 which explicitly pinned the gradle distribution at 6.3 which causes issues and is an unnecessary step. It also adds a callout about android 12+ versions.

### DIFF
--- a/src/fragments/lib/push-notifications/js/getting-started.mdx
+++ b/src/fragments/lib/push-notifications/js/getting-started.mdx
@@ -91,13 +91,13 @@ Setup instructions are provided for Android and iOS, and configuration for both 
     apply plugin: "com.google.gms.google-services"
     ``` 
 
-8. Open *android/gradle/wrapper/gradle-wrapper.properties* update the Gradle `distributionUrl`:
+8. Open *android/app/src/main/AndroidManifest.xml* file and add the following configuration into `application` element.
 
-    ```properties
-    distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-all.zip
-    ```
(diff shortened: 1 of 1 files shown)


abdallahshaban557 (dismissed):
LGTM

src/fragments/lib/push-notifications/js/getting-started.mdx
```diff
@@ -91,13 +91,13 @@ Setup instructions are provided for Android and iOS, and configuration for both
...
-8. Open *android/gradle/wrapper/gradle-wrapper.properties* update the Gradle `distributionUrl`:
+8. Open *android/app/src/main/AndroidManifest.xml* file and add the following configuration into `application` element.
 
-    ```properties
-    distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-all.zip
-    ```
+    <Callout>
 
-9. Open *android/app/src/main/AndroidManifest.xml* file and add the following configuration into `application` element.
+    When using Android 12+, make sure to add the property `android:exported="false"` to the service tags below.
```
cshfang, line 98: I think we might be able to just add that property to our snippet. Android 12 began enforcing that this property be explicitly specified but there should be no harm (and is perhaps better practice) to include it regardless.

tannerabread: Changed this to suggested

cshfang (approved):
Lgtm

abdallahshaban557 (approved):
LGTM

PR #4947, opened 2022-12-13
Add React example to JavaScript Storage download page

kevinold:
_Issue #, if available:_

_Description of changes:_

Add an example of the downloadBlob functionality in a React component.

### DIFF
--- a/src/fragments/lib/storage/js/download.mdx
+++ b/src/fragments/lib/storage/js/download.mdx
@@ -137,6 +137,42 @@ You can use `expires` option to limit the availability of your URLs. This config
 await Storage.get('filename.txt', { expires: 60 });
 ```
 
+### Example
+
+Below is an example of a React component with a Download link to a file:
+
+```javascript
+import { Storage } from "aws-amplify";
+import React from "react";
+
+function DownloadLinkExample({ fileKey }) {
+  function downloadBlob(blob, filename) {
+    const url = URL.createObjectURL(blob);
+    const a = document.createElement("a");
+    a.href = url;
+    a.download = filename || "download";
+    const clickHandler = () => {
+      setTimeout(() => {
+        URL.revokeObjectURL(url);
+        a.removeEventListener("click", clickHandler);
+      }, 150);
+    };
+    a.addEventListener("click", clickHandler, false);
+    a.click();
+    return a;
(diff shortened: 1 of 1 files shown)


rachelnabors (approved):

abdallahshaban557 (approved):
LGTM1

src/fragments/lib/storage/js/download.mdx
```diff
@@ -137,6 +137,42 @@ You can use `expires` option to limit the availability of your URLs. This config
 await Storage.get('filename.txt', { expires: 60 });
 ```
 
+### Example
+
+Below is an example of a React component with a Download link to a file:
+
+```javascript
```
hdworld11, line 144: Should this be titled 'React' instead?

kevinold: I just added an Example section in which this or other examples could live.  If we add additional examples, we could use subheaders or paragraphs for those

hdworld11: Makes sense, thank you!

hdworld11 (approved):

src/fragments/lib/storage/js/download.mdx
```diff
@@ -137,6 +137,42 @@ You can use `expires` option to limit the availability of your URLs. This config
 await Storage.get('filename.txt', { expires: 60 });
 ```
 
+### Example
+
+Below is an example of a React component with a Download link to a file:
```
stocaaro, line 142: In general, do we want to extend our js examples with framework specific versions? This is inconsistent with our other documentation and we should review this to ensure it aligns with our docs strategy.

This wraps the “[File download option](https://docs.amplify.aws/lib/storage/download/q/platform/js/#file-download-option)” example in React, extending it to show the user a download link. Should we recommend developers offer file download using the `download` flag?

If we're looking to provide an example of how to download a file using storage, wouldn't an example using an `<a href...>`  like the one detailed under "[Get](https://docs.amplify.aws/lib/storage/download/q/platform/js/#get)" be better for our customers so our their users are allowing the browser to handle the file download instead of pulling the file contents into memory before triggering the download?

stocaaro (changes requested):

PR #4930, opened 2022-12-07
feat: enable raw HTML strings in sidenav links

josefaidt:
_Description of changes:_

This PR aims to enable the usage of raw HTML strings inside `directory.js` in order to render inline code blocks inside sidenav link items. 

It introduces a use of `dangerouslySetInnerHTML` in the DirectoryGroup class component, however it first validates whether the supplied string contains an inline code block. 

image

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #2110, opened 2020-07-14
[Feedback] exhaustive reference for cli options


cyrfer:
**Page**: [`/cli`](https://docs.amplify.aws/cli)

**Feedback**:

Where is the exhaustive reference for options to commands like `amplify init` ?

I'm looking for something like:
https://docs.aws.amazon.com/cli/latest/reference/

sw-yx:
i like this idea but not sure it has a ton of depth - most `amplify ___` commands use prompts rather than CLI flags. still, worth documenting.

cyrfer:
I only care about coverage. I cannot find descriptions for options on several amplify commands.

Normally, commands support `help` or `-h` or `--help` that describe their options, but I've not been able to find info on options for commands, despite people mentioning those options in other github issues, e.g.

```bash
amplify init \
--amplify $AMPLIFY \
--providers $PROVIDERS \
--yes
```
https://github.com/aws-amplify/amplify-cli/issues/3292

Documention like this is built into some CLI frameworks, like [oclif](https://oclif.io/).

dabit3:
If we were to put this in the docs, where would it make sense for us to provide it?

cyrfer:
hey @dabit3 

I think you can put them anywhere. Some ideas (maybe all of them):

1. in the body of https://docs.amplify.aws/cli
   * specifically https://docs.amplify.aws/cli#reference
2. make the docs here more detailed: https://github.com/aws-amplify/amplify-cli#commands-summary
3. another item under `Reference`
![image](https://user-images.githubusercontent.com/4392990/88318058-09a00500-cccf-11ea-9888-0a006384ff94.png)

dsypniewski:
I just found this https://docs.amplify.aws/cli/usage/headless
Maybe someone here will find it useful as I just did, but still built in -h/--help and at least some reference to the headless page from already mentioned places like "Reference" section would be nice as well.

sw-yx:
fwiw I strongly agree that any professional quality CLI should support standard conventions like  - - help. it's just that the ask here is also for a web version. I don't mind putting one together.

pinging @renebrandel for this feedback on the cli

PR #4950, opened 2022-12-14
feat: add Amplify CLI command reference

josefaidt:
_Issue #, if available:_

#2110 
#5207 
#5292
#5311

_Description of changes:_

image

UPDATED

image

image

image

### Todo

- [x] ensure sidenav links are displayed as inline code blocks #4930  
- [ ] ~~ensure titles are displayed as the full command as an inline code block (e.g. `amplify env`)~~
- [x] ensure page SEO is working as expected
- [ ] ~~ensure "previous" and "next" buttons display commands as inline code blocks~~
- [ ] ~~remove ALL references to commands, their structure, and their usage, reference new pages instead~~
	- https://github.com/aws-amplify/docs/issues/5153
- [x] automate pulling commands from Amplify CLI


josefaidt:
adding do-not-merge to look into issue where data is getting passed around to command pages, where navigating away from and back to a specific command will carry subcommands, options, and flags back to the command page

package.json
```diff
@@ -24,6 +24,7 @@
     "@emotion/styled": "^11.3.0",
     "@mdx-js/mdx": "^2.0.0",
     "algoliasearch": "^4.12.1",
+    "amplify-cli-core": "3.6.2",
```
edwardfoyle, line 27: should this be carat versioned?

src/data/cli-commands.mjs
```diff
@@ -0,0 +1,49 @@
+import { commandsInfo } from 'amplify-cli-core/lib/help/commands-info.js';
```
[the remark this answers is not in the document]
josefaidt: oh that's a good question. If this is changed in the CLI the docs site will not build. Ideally if we automate this on updates to `amplify-cli-core` the PR associated with the version bump and lockfile updates will fail

package.json
josefaidt: I originally had this set to `latest`, which would only require a lockfile update when updating docs after a CLI release. It's pinned to be explicit, but I don't see any issues with adding a carat

sobolk: Also, please switch to `^4.0.0`

josefaidt: @sobolk I've bumped the version and changed the package name, though unfortunately I am unable to import the help information 
```ts
import { commandsInfo } from '@aws-amplify/amplify-cli-core';
```
image

Thankfully using yarn will allow us to still import the transient dependency, so it's not entirely necessary to update the import just yet

lazpavel (approved):

katiegoines (approved):

Issue #4512, opened 2022-08-05
Provide documentation for TOTP flow for JS


abdallahshaban557:
**Is your feature request related to a problem? Please describe.**
As an Amplify JS customer, I want to have a clear description and walkthrough of how to setup TOTP authentication using the Amplify JS library, so that I do not have to spend a lot of time in setting up my application.

**Describe the solution you'd like**
Add a documentation page to the Authentication category documentation on docs.amplify.aws

**Additional context**
There are [steps](https://github.com/aws-amplify/amplify-js/issues/5271#issuecomment-1205797880) mentioned here on how to make TOTP work in Amplify JS

cwomack:
Closing this after PR #4961 was merged in.

PR #4961, opened 2022-12-15
(chore) - TOTP setup refactor

cwomack:
_Issue #, if available:_

#4512 

_Description of changes:_

### DIFF
--- a/src/fragments/lib/auth/js/mfa.mdx
+++ b/src/fragments/lib/auth/js/mfa.mdx
@@ -17,38 +17,6 @@ You can setup TOTP for a user in your app:
 ```javascript
 import { Auth } from 'aws-amplify';
 
-const user = await Auth.signIn(username, password);
-
-userEffect(() => {
-  async function createCode() { 
-    try {
-      const codeSetUp = await Auth.setupTOTP(props.user);
-      setCodeTOTP(codeSetUp)
-    } catch (error) {
-      console.log(error)
-    }
-  };
-  if (props.user?.challengeName !== 'SOFTWARE_TOKEN_MFA') {
-    createCode();
-  }
-}, [props.user])
-
-const handleCode = async (event: any) => {
-  event.preventDefault()
-  try {
-    if (props.user?.challengeName !== 'SOFTWARE_TOKEN_MFA') {
-      await Auth.verifyTotpToken(props.user, code)
-    } else {
-      await Auth.confirmSignIn(props.user, code, 'SOFTWARE_TOKEN_MFA')
-    }
-    navigate('/dashboard', {replace: true})
-  } catch (error) {
-    console.log(error)
-  }
-}
-
-// PREVIOUS START ----------------------------------------------------------------
-
 // To setup TOTP, first you need to get a `authorization code` from Amazon Cognito
 // `user` is the current Authenticated user
 Auth.setupTOTP(user).then((code) => {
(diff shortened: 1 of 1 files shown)


cwomack:
Started with the original community example given within https://github.com/aws-amplify/amplify-js/issues/5271#issuecomment-1205797880 but stripped it down to only include the relevant methods and missing example of using the confirmSignIn when MFA is enabled.

src/fragments/lib/auth/js/mfa.mdx
```diff
@@ -31,6 +31,7 @@ Auth.setupTOTP(user).then((code) => {
 
 // Then you will have your TOTP account in your TOTP-generating app (like Google Authenticator)
 // Use the generated one-time password to verify the setup
+// This can also be used if user is already signed in for testing/validation
```
tannerabread, line 34: Didn't notice line 33 when we first spoke about this, maybe some combination of 33 + 34 would be better because 33 already mentions that this method is for verifying the setup, although that's probably a little vague for some users.
I don't think both lines would be needed though

src/fragments/lib/auth/js/mfa.mdx
```diff
@@ -40,6 +41,18 @@ Auth.verifyTotpToken(user, challengeAnswer)
   .catch((e) => {
     // Token is not verified
   });
+
+// If signing in with MFA is enabled, use confirmSignIn method to 
+// send a new MFA code to be used with Authenticator.
```
tannerabread, line 46: The previous 2 comments in this section start with `first` and `then`, maybe put a `finally` for this one or take out the sequence words completely.

Also, `confirmSignIn` doesn't send the new MFA code to the authenticator, instead you would have already set up your MFA authenticator program and it should be auto-refreshing every ~30 seconds or whatever the time is on the app. `confirmSignIn` is used to verify that you have the MFA code in addition to the username/password. So this method would be used following the `signIn` method which then would fail because it needs the 
SMS_MFA or SOFTWARE_TOKEN_MFA `user.challengeName`

src/fragments/lib/auth/js/mfa.mdx
```diff
@@ -40,6 +41,18 @@ Auth.verifyTotpToken(user, challengeAnswer)
...
+
+// If signing in with MFA is enabled, use confirmSignIn method to 
+// send a new MFA code to be used with Authenticator. 
+
+// ...
+
+if (
+  user.challengeName === "SMS_MFA" ||
+  user.challengeName === "SOFTWARE_TOKEN_MFA"
+) { const loggedUser = await Auth.confirmSignIn( user, challengeAnswer )}
```
tannerabread, line 53:
```suggestion
Auth.confirmSignIn(user, code, mfaType) // MFA Type e.g. SMS_MFA || SOFTWARE_TOKEN_MFA
```

tannerabread (commented):
Excuse the fact that some things are different from what we spoke about yesterday, once I saw it written it didn't look cohesive with the rest of the docs or where I think we are trying to go to

abdallahshaban557 (approved):
LGTM!

nadetastic (commented):
Should `setupTOTP` be included in this doc? 

https://aws-amplify.github.io/amplify-js/api/classes/authclass.html#setuptotp

nadetastic (approved):
Looks good to me!!

Issue #4943, opened 2022-12-09
Left vertical menu doesn't scroll to the very bottom


helgabalashova:
**Describe the bug**
When several menu items dropdowns are open, it is impossible to scroll to the very bottom of the left vertical menu. The ways to get there are to close dropdowns or to scroll all content on the right side to the bottom and then the menu will also be scrolled to the very bottom.

**To Reproduce**
1. Go to [amplify docs](https://docs.amplify.aws/lib/auth/getting-started/q/platform/js/)
2. Click on Authentication and DataStore dropdowns
3. Place cursor to the left menu and try to scroll down.
4. It will only scroll to a certain point and then content on the right will start scrolling
5. after scrolling on the right goes to the end, the menu on the left will also scroll down

**Expected behavior**
Should be able to scroll to the very bottom of the left nav bar

**Screenshots**

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser: Chrome
 - Version 108.0.5359.71 (Official Build) (arm64)

timngyn:
Hi @helgabalashova, sorry I forgot to let you know that this should be fixed now. Please feel free to open another issue if there is something wrong still. Closing this issue now. Thank you!

PR #4962, opened 2022-12-15
fix: add bottom padding to menu so there is more room to scroll

timngyn:
_Issue #, if available:_ #4943

_Description of changes:_
- Add extra `padding-bottom` to the Menu component so that there is more

_To test:_
1. Go to `/lib/auth/manageusers/q/platform/js/`
2. Expand some of the section collapsible headers in the left menu
3. Try to scroll down in the left menu
    - You should be able to scroll and reach the bottom of the menu without having to scroll the content page

### DIFF
--- a/src/components/Menu/styles.tsx
+++ b/src/components/Menu/styles.tsx
@@ -29,7 +29,6 @@ export const MenuStyle = styled.div`
       max-height: 100vh;
       overflow-y: auto; /* for Firefox */
       overflow-y: overlay; /* for Webkit browsers */
-      margin-bottom: 6rem;
     }
   }
 `;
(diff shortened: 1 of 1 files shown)


src/components/Menu/styles.tsx
```diff
@@ -37,6 +37,7 @@ export const MenuStyle = styled.div`
 export const MenuBodyStyle = styled.div`
   display: block;
   margin: 0 2.5rem;
+  padding-bottom: 200px;
```
jacoblogan, line 40: It looks like this is doing what line 32 is trying to do, can we also remove the `margin-bottom: 6rem` from line 32 of this file.

timngyn: Thanks so much Jacob!! I ended up changing it to `padding-bottom: 9rem` to give it extra space for viewing the page last updated text. Staging site should up in a bit

jacoblogan (approved):

katiegoines (approved):
lgtm

PR #4977, opened 2022-12-21
fix(auth): Removed the note about confirm reset password needing to be in conjuction with reset password

gpanshu:
_Issue #, if available:_

_Description of changes:_ Currently the documentation mentions that reset and confirm reset password have to be used in the same session but that is not correct for V2 for Android or V1/V2 for iOS.

### DIFF
--- a/src/fragments/lib-v1/auth/android/password_management/10_reset_password_note.mdx
+++ b/src/fragments/lib-v1/auth/android/password_management/10_reset_password_note.mdx
@@ -1,4 +1,3 @@
 <Callout>
 Note that you must call confirmResetPassword in the same app session as you call resetPassword. If you close the app, you'll need to call resetPassword again.
-As a result, for testing purposes, you'll at least need an input field where you can enter the code sent by the resetPassword api and feed it to confirmResetPassword.
 </Callout>
--- a/src/fragments/lib-v1/auth/android/password_management/10_reset_password_note.mdx
+++ b/src/fragments/lib-v1/auth/android/password_management/10_reset_password_note.mdx
@@ -1,0 +1,4 @@
+<Callout>
+Note that you must call confirmResetPassword in the same app session as you call resetPassword. If you close the app, you'll need to call resetPassword again.
+As a result, for testing purposes, you'll at least need an input field where you can enter the code sent by the resetPassword api and feed it to confirmResetPassword.
+</Callout>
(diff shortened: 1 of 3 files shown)


royjit (dismissed):

sdhuka (approved):

src/fragments/lib-v1/auth/android/password_management/10_reset_password_note.mdx
```diff
@@ -0,0 +1,4 @@
+<Callout>
+Note that you must call confirmResetPassword in the same app session as you call resetPassword. If you close the app, you'll need to call resetPassword again.
+As a result, for testing purposes, you'll at least need an input field where you can enter the code sent by the resetPassword api and feed it to confirmResetPassword.
```
abdallahshaban557, line 3: I feel like this second part is a bit overkill. Why do we need to add it specifically here?

gpanshu: I agree and wanted to make this change but this is how it is currently there in V1. I just moved it under Android only as it only pertains to Android. I can remove the last 2 lines if you would like?

abdallahshaban557: yes, please! I think that would make that message a lot more concise!

gpanshu: Done.

abdallahshaban557 (dismissed):
LGTM!

sdhuka (approved):

src/fragments/lib-v1/auth/android/password_management/10_reset_password_note.mdx
```diff
@@ -0,0 +1,3 @@
+<Callout>
+Note that you must call confirmResetPassword in the same app session as you call resetPassword. If you close the app, you'll need to call resetPassword again.
```
josefaidt, line 2: is `confirmResetPassword` a method? should this be represented as an inline code block?

```suggestion
Note that you must call `confirmResetPassword` in the same app session as you call `resetPassword`. If you close the app, you'll need to call `resetPassword` again.
```

rachelnabors (approved):

abdallahshaban557 (approved):
Approved

sdhuka (approved):

josefaidt (approved):

PR #5006, opened 2023-01-11
add heading hashes to algolia search results

jacoblogan:
_Description of changes:_
This pull request modifies the algolia search index script to include headings as part of the search link so search results will take users to the desired section within the page instead of just linking to the top of the page.


katiegoines (approved):

timngyn:
Noticed that on the staging site that the new search does navigate to the closest heading, but it gets covered by the secondary nav bar. We can look at it together if you want. Not too sure why it doesn't work at the moment since the link you created correctly scrolls to the section if we open it in a new tab

PR #5007, opened 2023-01-12
Add callout when using existing Cognito Resource

nadetastic:
_Issue #, if available:_
https://github.com/aws-amplify/amplify-js/issues/3526

_Description of changes:_
Add a callout for when using existing Cognito resources. If UserPool client doesn't have required attributes set to mutable, then login failure will happen with Social Sign In


src/fragments/lib/auth/js/start.mdx
```diff
@@ -127,6 +127,9 @@ Amplify.configure({
 // You can get the current config object
 const currentConfig = Auth.configure();
 ```
+<Callout warning={true}>
```
abdallahshaban557, line 130: Is there a link we can point customers to on how to do that? maybe from the Cognito docs?

nadetastic: Yes, this looks sufficient - https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-client-apps.html

I'll add and update

abdallahshaban557 (approved):
LGTM!

cwomack (approved):
LGTM with the addition of the reference link!

PR #5022, opened 2023-01-17
add github workflow to run the link tag check script on each PR

jacoblogan:
_Description of changes:_
This PR adds a github workflow to run the script that scans all of the mdx files for HTML A link tags.  We should always be using the markdown syntax for links in the mdx files and this script will block if PRs are found with html a link tags.

### DIFF
--- a/.github/workflows/tagcheck.yml
+++ b/.github/workflows/tagcheck.yml
@@ -1,0 +1,20 @@
+name: Linktagcheck
+on:
+  pull_request:
+    types: [opened, synchronize]
+permissions:
+  contents: read
+jobs:
+  Spellcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Setup Node.js 14.x
+        uses: actions/setup-node@main
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/tagcheck.yml
```diff
@@ -0,0 +1,20 @@
+name: Linktagcheck
+on:
+  pull_request:
+    types: [opened, synchronize]
+permissions:
+  contents: read
+jobs:
+  Linktagcheck:
```
timngyn, line 8: Is it okay if we use pascal case for the build name so it's `LinkTagCheck`?

```suggestion
  LinkTagCheck:
```

timngyn (approved):
Looks good to me! Just had one small nitpick, but it's not blocking

.github/workflows/tagcheck.yml
```diff
@@ -0,0 +1,20 @@
+name: Linktagcheck
```
timngyn, line 1:
```suggestion
name: LinkTagCheck
```

timngyn (approved):

katiegoines (approved):

PR #4935, opened 2022-12-08
remove React Native CLI steps from DataStore getting started guide for JS

chrisbonifacio:
_Issue #, if available:_

_Description of changes:_

The JS page for DataStore's Getting Started guide included some steps to initialize a React Native CLI app and configure the SQLite Adapter. These steps have been removed, only the Create React App init steps should be rendered.

Verified the getting started guide is still comprehensive by following the remaining steps locally.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5024, opened 2023-01-17
Fix/js datastore schema updates

chrisbonifacio:
_Issue #, if available:_

_Description of changes:_

Adding the path to the local GraphQL API schema that needs to be updated before re-deployment. This is to reflect the GraphQL API [documentation](https://docs.amplify.aws/lib/graphqlapi/getting-started/q/platform/js/#updating-your-graphql-schema) which clearly mentions it to the reader.


src/fragments/lib/datastore/js/getting-started/30_platformIntegration.mdx
```diff
@@ -14,68 +14,3 @@ npx create-react-app amplify-datastore --use-npm
 cd amplify-datastore
 npx amplify-app@latest
 ```
-
```
david-mcafee, line 17: @chrisbonifacio - was the removal of this intentional?

david-mcafee (changes requested):

src/fragments/lib/datastore/js/getting-started/30_platformIntegration.mdx
chrisbonifacio: Was not intentional in this commit, this change was already merged in another PR: https://github.com/aws-amplify/docs/pull/4935

Would merging the same change have any affect? I can try to make a clean commit if need be

PR #5037, opened 2023-01-19
fix: explicit Has Many definition

iartemiev:
_Issue #, if available:_
https://github.com/aws-amplify/amplify-js/issues/10864#issuecomment-1396971466

_Description of changes:_
We currently have an incorrect example in the docs for defining explicit Has Many relationships that is re-used across the CLI and client libraries (DataStore section of each platform)

We instruct customers to create an index with a composite key with two fields on the child that's associated with a single field on the parent. If using a composite key index, we would need to have the same number of associated fields on the parent as we do on the child.

This PR corrects this by removing the additional field from the index.

Current state in prod (see `sortKeyFields`):
```graphql
type Post @model {
  id: ID!
  title: String!
  comments: [Comment] @hasMany(indexName: "byPost", fields: ["id"])
}

type Comment @model {
  id: ID!
  postID: ID! @index(name: "byPost", sortKeyFields: ["content"])
  content: String!
}
```

Change in PR:
```graphql
type Post @model {
  id: ID!
  title: String!
  comments: [Comment] @hasMany(indexName: "byPost", fields: ["id"])
}

type Comment @model {
  id: ID!
  postID: ID! @index(name: "byPost")
  content: String!
}
```

### DIFF
--- a/src/fragments/lib-v1/datastore/android/relational/updated-schema.mdx
+++ b/src/fragments/lib-v1/datastore/android/relational/updated-schema.mdx
@@ -18,7 +18,7 @@ type Post @model {
 # New model
 type Comment @model {
   id: ID!
-  postID: ID! @index(name: "byPost", sortKeyFields: ["content"])
+  postID: ID! @index(name: "byPost")
   post: Post! @belongsTo(fields: ["postID"])
   content: String!
 }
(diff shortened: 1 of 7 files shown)


marcvberg (approved):
LGTM

alharris-at (approved):

iartemiev:
Converting to draft while I do some additional validation

david-mcafee (approved):
LGTM!

iartemiev:
We're going to treat this as a bug in DataStore instead and will add support for this type of explicit hasMany definition. Closing the PR

PR #5110, opened 2023-02-13
Add workflow to check for new files so CODEOWNERS can be updated

timngyn:
#### Description of changes:
- Add workflow to look for new files from the PR and check to see if its path is covered by CODEOWNERS. If it isn't then it adds a comment to the PR specifying which file wasn't covered by CODEOWNERS.

### Instructions

- [ ] amplify-cli
- [ ] amplify-ui
- [ ] amplify-studio
- [ ] amplify-hosting
- [ ] amplify-libraries

- [ ] JS
- [ ] iOS
- [ ] Android
- [ ] Flutter
- [ ] React Native

#### Checks

- [ ] Ready to merge


katiegoines (approved):
makes sense to me :)

katiegoines (approved):

.github/workflows/check_for_new_files.yml
```diff
@@ -0,0 +1,75 @@
...
+        uses: actions/github-script@v6.4.0
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const ignore = require('ignore');
+            const { NEW_FILES, CURRENT_BRANCH } = process.env
+
+            console.log('New files: ', NEW_FILES);
+
+            const response = await fetch(`https://raw.githubusercontent.com/aws-amplify/docs/${CURRENT_BRANCH}/.github/CODEOWNERS`);
```
timngyn, line 49: The goal of this workflow is that if a new file is added and it isn't covered by a CODEOWNER, then the current PR should make an update to specify it. By fetching the CODEOWNERS file from the current branch here we can keep track of when the files do get added to CODEOWNERS.

katiegoines (approved):

wlee221 (changes requested):
Instead of writing inline scripts, I'm wondering if we should move the scripts into its own bash or JS file.

Large inline scripts tend to be hard to maintain, because they don't get any syntax highlighting or type checks (for JS).

.github/workflows/scripts/check_for_new_files.js
```diff
@@ -0,0 +1,77 @@
...
+  },
+  validateCodeowners: async ({ github, context, fetch, ignore }) => {
+    const { NEW_FILES, CURRENT_BRANCH, CURRENT_REPO } = process.env;
+
+    const codeownersFile = `https://raw.githubusercontent.com/${CURRENT_REPO}/${CURRENT_BRANCH}/.github/CODEOWNERS`;
+
+    console.log('Fetching CODEOWNERS from: ', codeownersFile);
+
+    const response = await fetch(codeownersFile);
+    const body = await response.text();
```
wlee221, line 31: is `await` required if `response` is already awaited?

.github/workflows/scripts/check_for_new_files.js
```diff
@@ -0,0 +1,77 @@
...
+
+    // Filter out comments from CODEOWNERS file
+    const codeownersFilePatterns = body
+      .split('\n')
+      .filter((e) => !e.startsWith('#'))
+      .filter((e) => e.length > 1)
+      .map((e) => e.split(/\s+/)[0]);
+
+    console.log(codeownersFilePatterns);
```
wlee221, line 40: Did you mean to leave this console log? If so, let's make the log more descriptive. If not, let's remove it.

.github/workflows/scripts/check_for_new_files.js
```diff
@@ -0,0 +1,77 @@
+module.exports = {
+  getAddedFiles: ({ github, context, core }) => {
+    const {
+      issue: { number: issue_number },
+      repo: { owner, repo }
+    } = context;
+
+    // Use the Github API to query for the list of files from the PR
+    github
```
wlee221, line 9: Can we use `github.rest`? I think that is easier to read and is consistent with the other API call you make below. You'd have to make this function async tho, but I prefer `await`s then `.then`s.

.github/workflows/check_for_new_files.yml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Get count of added files
+        uses: actions/github-script@v6.4.0
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const { getAddedFiles } = require('./.github/workflows/scripts/check_for_new_files.js');
+            getAddedFiles({github, context, core});
```
wlee221, line 20: FYI, you can [return string data](https://github.com/actions/github-script#result-encoding) from `github-script` and reference it in later steps. 

I personally prefer that because it's more clear where the `NEW_FILES_COUNT` is coming from in later steps, but I'll leave the decision to you!

.github/workflows/check_for_new_files.yml
```diff
@@ -0,0 +1,35 @@
+name: Validate CODEOWNERS
+on:
+  pull_request:
+    branches: [main]
+jobs:
+  checkIfCodeownersUpdateNeeded:
+    permissions:
+      pull-requests: write
```
wlee221, line 8: Can you add an inline comment why this permission is needed for future readers?

.github/workflows/scripts/check_for_new_files.js
timngyn: I think we could, but I wanted to use `github.paginate` because it makes it easier to grab all the files and filter them inline since the list of files for a PR might require pagination: https://octokit.github.io/rest.js/v19#pagination

If you think changing it to `github.rest` would be easier than I can do that too, otherwise I could add a comment to explain why I'm using `paginate`

.github/workflows/scripts/check_for_new_files.js
timngyn: Yea, response is technically a `Promise` still so I need to await it to get the actual string: https://developer.mozilla.org/en-US/docs/Web/API/Response/text

.github/workflows/scripts/check_for_new_files.js
wlee221: That makes sense to me, I feel like code is self explanatory then

wlee221 (approved):
LGTM from technical perspective 👍

katiegoines (approved):

PR #5315, opened 2023-03-30
Upgrade to Next13

katiegoines:
#### Description of changes:

- upgraded Next to v13
- upgraded React to v18

### Instructions

- [ ] amplify-cli
- [ ] amplify-ui
- [ ] amplify-studio
- [ ] amplify-hosting
- [ ] amplify-libraries

- [ ] JS
- [ ] iOS
- [ ] Android
- [ ] Flutter
- [ ] React Native

#### Checks

- [ ] Ready to merge


josefaidt (dismissed):
Hey @katiegoines this looks great! I was able to poke around the preview and confirm things are working as expected. Although I have concerns about the `domLoaded` piece it will be a larger effort to fix the hydration issues, and I do not see any negative impact with this in place compared to the current setup.

package.json
```diff
@@ -31,17 +31,19 @@
     "copy-to-clipboard": "^3.2.1",
     "dotenv": "^16.0.0",
     "emotion": "^10.0.23",
+    "eslint-config-next": "^13.2.4",
     "extract-mdx-metadata": "^2.0.0",
     "gray-matter": "4.0.3",
     "html-entities": "^1.2.1",
     "instantsearch.js": "^4.39.1",
-    "next": "^12.1.4",
+    "module": "^1.2.5",
```
josefaidt, line 39: this is a node.js built-in https://nodejs.org/docs/latest-v16.x/api/module.html

```suggestion
```

src/components/InternalLink/index.tsx
```diff
@@ -28,6 +31,7 @@ export default function InternalLink({ href, children }) {
           href = `${hrefParts[0]}/q/${filterKind}/${filterKey}#${hrefParts[1]}`;
         } else {
           href += `/q/${filterKind}/${filterKey}`;
+          useEffect(() => setUrl(href), []);
```
josefaidt, line 34: does this need a `useEffect` to set the href properly? Can all of this conditional logic be inside of the `useEffect`? https://legacy.reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

parsing local storage (i.e. from `window`) should also reside inside the useEffect as `window` will not be available server-side  (during prerendering)

src/components/Layout/index.tsx
```diff
@@ -10,7 +10,7 @@ import {
   RIGHT_NAV_LINKS,
   SOCIAL_LINKS
 } from '../../utils/globalnav';
-import React from 'react';
+import React, { useEffect, useState } from 'react';
```
josefaidt, line 13: this isn't entirely necessary to change but the `React` import is no longer required for React 17+ https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

src/components/SearchBar/Autocomplete.tsx
```diff
@@ -98,8 +100,14 @@ export function Autocomplete(props) {
...
-        render(children, root);
-      },
+        if (!panelRootRef.current || rootRef.current !== root) {
+          rootRef.current = root;
+
+          panelRootRef.current?.unmount();
+          panelRootRef.current = createRoot(root);
+        }
+
+        panelRootRef.current.render(children);      },
```
josefaidt, line 110: what is this doing? Is it possible to use `React.lazy` with Suspense here? https://legacy.reactjs.org/docs/code-splitting.html#reactlazy

or maybe `next/dynamic`? https://nextjs.org/docs/advanced-features/dynamic-import

timngyn: I'm also a little confused by this as well. Should we add some comments for why we're doing this?

timngyn (commented):
It looks good so far! Thank you for doing this @katiegoines

I have similar questions to what @josefaidt asked as well with regards to the `useEffect` for calling `setUrl`

src/components/Menu/FilterSelect/index.tsx
```diff
@@ -31,30 +31,19 @@ const convertToRouteHerf = (
...
-  return {
-    pathname: path,
-    query: {
-      [filter.filterKind]: targetFilterKey,
-      firstPathSegment: multiLibVersionPlatforms.includes(targetFilterKey)
-        ? firstPathSegment
-        : firstPathSegment?.split('-')[0]
-    }
-  };
+  return path;
```
jacoblogan, line 46: It looks like here we are changing the return type of this function but not how we use it.  Is the expected output for this ever expected to be the object that was previously being returned?

src/components/Callout/styles.tsx
```diff
@@ -18,7 +18,7 @@ export const CalloutOuter = styled.div<CalloutProps>((props) => {
     margin-bottom: 0;
     margin-top: 1rem;
 
-    &:first-child {
+    &:first-of-type {
```
jacoblogan, line 21: These two selectors will often have the same effect but not always.  Just for my own understanding, why the switch to first-of-type?

src/components/SearchBar/Autocomplete.tsx
katiegoines: This is following the Algolia recommendation for upgrading to React18. Looking into the alternatives suggested.

src/components/InternalLink/index.tsx
katiegoines: This is hydration related. I missed removing it when separating out the hydration efforts. Will address this on that branch. Thanks

src/components/Menu/FilterSelect/index.tsx
```diff
@@ -106,7 +101,7 @@ export default class FilterSelect extends React.Component<
     }
 
     return (
-      <Link href={href} key={name}>
+      <Link href={href} key={name} legacyBehavior>
```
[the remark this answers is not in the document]
katiegoines: Yes, in some places it did not make sense to remove out the <a> when there is the workaround.

src/components/Callout/styles.tsx
katiegoines: We were getting a warning, but it's not related to the upgrade, so I've removed these and they can be addressed at another time.

josefaidt (approved):
LGTM 🚀 Great work on this, @katiegoines!

jacoblogan (approved):

jacoblogan (approved):

josefaidt (approved):

PR #5335, opened 2023-04-04
Adding link to flutter reference docs to top of navigation

rachelnabors:
#### Description of changes:

Adds a "flutter API references" index page to the docs that links out to all the flutter docs on pub.dev, visible only to people who have chosen Flutter as their platform, visible at **docs.amplify.aws/flutter-references**

image


abdallahshaban557:
@rachelnabors - the only thing I have noticed is that the platform picker disappears when you go to the API references page. Is that expected? Or would we want to add it there as well?

rachelnabors:
I'm guessing if you're on the flutter API page, since there is no alternative in other platforms and you're looking at Flutter docs, it isn't a very useful thing to have!

abdallahshaban557:
got it! approving!

abdallahshaban557 (dismissed):
LGTM

abdallahshaban557 (approved):
LGTM

PR #5372, opened 2023-04-11
feat(storage): add validateObjectExistence option in get api

kvramyasri7:
#### Description of changes:
This change is to enable the users to check the existence of the file.

### Instructions

- [ ] amplify-cli
- [ ] amplify-ui
- [ ] amplify-studio
- [ ] amplify-hosting
- [X] amplify-libraries

- [X] JS
- [ ] iOS
- [ ] Android
- [ ] Flutter
- [ ] React Native

#### Checks

- [X] Does this PR conform to [the styleguide](https://github.com/aws-amplify/docs/blob/main/STYLEGUIDE.md)?

- [X] Does this PR include filetypes other than markdown or images? Please add or update unit tests accordingly.

- [X] Are any files being deleted with this PR? If so, have the needed redirects been created?

- [X] Are all links in MDX files using the MDX link syntax rather than HTML link syntax? <br /> 

- [x] Ready to merge

### DIFF
--- a/src/fragments/lib/storage/js/download.mdx
+++ b/src/fragments/lib/storage/js/download.mdx
@@ -13,7 +13,7 @@ await Storage.get(key: string, config: {
 });
 ```
 
-`Storage.get` returns a signed URL `string` to your file, if `download` is false, which is the default. You can use this to create a download link for people to click on. This is our recommended option. Note that this method **does not check if the file actually exists by default**. But you can check for the existence of file by setting `validateObjectExistence` to `true`.
+`Storage.get` returns a signed URL `string` to your file if `download` is false, which is the default. You can use this to create a download link for users to click on. Note that this function **does not check if the file exists by default**. You can enable this check by setting `validateObjectExistence` to `true`.
 
 ```js
 // get the signed URL string
--- a/src/fragments/lib/storage/js/download.mdx
+++ b/src/fragments/lib/storage/js/download.mdx
@@ -143,7 +143,6 @@ await Storage.get('filename.txt', { expires: 60 });
 
 Users can run into unexpected issues, so we are giving you advance notice in documentation with links to open issues - please vote for what you need, to help the team prioritize.
 
-- Very common issue: Calling `Storage.get` for a nonexistent file, or incorrect credentials, **does not throw an error** because that would involve an extra API call. [Active issue here](https://github.com/aws-amplify/amplify-js/issues/1145). If you are having trouble accessing a file, make sure to check that you have the right filename, bucket, region, and auth configs. You can get a current list of files from `Storage.list`.
 - `Storage.get` is cached; if you have recently modified a file you may not get the most recent version right away. [There is an active issue for a new option to enable cache busting.](https://github.com/aws-amplify/amplify-js/issues/6413)
 - `Storage.get` only returns the latest cached version of the file; there is [not yet an API to view prior versions](https://github.com/aws-amplify/amplify-js/issues/2131).
 - You cannot [only get file metadata](https://github.com/aws-amplify/amplify-js/issues/6157) yet.
--- a/src/fragments/lib/storage/js/download.mdx
+++ b/src/fragments/lib/storage/js/download.mdx
@@ -139,6 +139,15 @@ You can use `expires` option to limit the availability of your URLs. This config
 await Storage.get('filename.txt', { expires: 60 });
 ```
 
+### Check existence of file
+
+You can check the existence of file in the storage using `validateObjectExistence` option. If the the file is existed the get call returns a pre-signed URL and if the file does not exist then return a 404 Error.
+
+```javascript
+// To check existence of file
+await Storage.get('filename.txt', { validateObjectExistence: true });
+```
+
 ### Frequently Asked Questions
 
 Users can run into unexpected issues, so we are giving you advance notice in documentation with links to open issues - please vote for what you need, to help the team prioritize.
--- a/src/fragments/lib/storage/js/download.mdx
+++ b/src/fragments/lib/storage/js/download.mdx
@@ -141,7 +141,7 @@ await Storage.get('filename.txt', { expires: 60 });
 
 ### Check for existence of a file
 
-You can check for the existence of a file in the storage category's `get` method using the `validateObjectExistence` option. When this flag is enabled a `get` call will return a pre-signed URL if the file exists and a `404` error if it does not.
+You can check for the existence of a file in the storage category's `get` API using the `validateObjectExistence` option. When this flag is enabled a `get` call will return a pre-signed URL if the file exists and raise a `404` error if it does not.
 
 ```javascript
 // To check for existence of a file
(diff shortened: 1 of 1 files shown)


src/fragments/lib/storage/js/download.mdx
```diff
@@ -9,17 +9,20 @@ await Storage.get(key: string, config: {
   download?: boolean, // defaults to false
   expires?: number, // validity of the URL, in seconds. defaults to 900 (15 minutes)
   contentType?: string // set return content type, eg "text/html"
+  validateObjectExistence?: boolean // defaults to false
 });
 ```
 
-`Storage.get` returns a signed URL `string` to your file, if `download` is false, which is the default. You can use this to create a download link for people to click on. This is our recommended option. Note that this method **does not check if the file actually exists** as that would involve an extra API call.
+`Storage.get` returns a signed URL `string` to your file, if `download` is false, which is the default. You can use this to create a download link for people to click on. This is our recommended option. Note that this method **does not check if the file actually exists by default**. But you can check for the existence of file by setting `validateObjectExistence` to `true`.
```
cshfang, line 16: Suggest updating method -> function for JS docs. Imo we can also just refer to the check as the check after talking about it in the penultimate sentence.
```suggestion
`Storage.get` returns a signed URL `string` to your file, if `download` is false, which is the default. You can use this to create a download link for people to click on. This is our recommended option. Note that this function **does not check if the file exists by default**. You can enable this check by setting `validateObjectExistence` to `true`.
```

abdallahshaban557, line 16: why is it our recommended option?

src/fragments/lib/storage/js/download.mdx
```diff
@@ -137,11 +139,19 @@ You can use `expires` option to limit the availability of your URLs. This config
 await Storage.get('filename.txt', { expires: 60 });
 ```
 
+### Check existence of file
```
cshfang, line 142: Nit: Arguable if this is better but it feels a bit more like treating the existence as a noun?
```suggestion
### Check for existence of a file
```

abdallahshaban557, line 142: Is there a reason why we are calling this out again in the guide?

src/fragments/lib/storage/js/download.mdx
```diff
@@ -137,11 +139,19 @@ You can use `expires` option to limit the availability of your URLs. This config
...
+
+You can check the existence of file in the storage using `validateObjectExistence` option. If the the file is existed the get call returns a pre-signed URL and if the file does not exist then return a 404 Error.
```
cshfang, line 144: Nit:
```suggestion
You can check for the existence of a file in the storage using the `validateObjectExistence` option. When enabled a `get` call will return a pre-signed URL if the file exists and a `404` error if it does not.
```

src/fragments/lib/storage/js/download.mdx
```diff
@@ -137,11 +139,19 @@ You can use `expires` option to limit the availability of your URLs. This config
...
+
+```javascript
+// To check existence of file
```
cshfang, line 147: Same nit:
```suggestion
// To check for existence of a file
```

cshfang (dismissed):

src/fragments/lib/storage/js/download.mdx
jimblanc, line 16: Suggested changes to this line with some changes to the existing wording, please double check me.

```suggestion
`Storage.get` returns a signed URL `string` to your file if `download` is false, which is the default. You can use this to create a download link for users to click on. Note that this method **does not check if the file actually exists by default**. You can check that the file exists when generating a signed URL by setting `validateObjectExistence` to `true`.
```

src/fragments/lib/storage/js/download.mdx
```diff
@@ -32,13 +35,11 @@ The `download` option send you the object data for download or programmatic mani
...
 
 // data.Body is a Blob
-result.Body.text().then(string => { 
-  // handle the String data return String 
+result.Body.text().then((string) => {
+  // handle the String data return String
 });
 ```
 
-Note that the `Blob` methods like `.text()` are not supported on [IE/Opera/Safari](https://developer.mozilla.org/en-US/docs/Web/API/Blob/text); in those cases you can [parse manually](https://developer.mozilla.org/en-US/docs/Web/API/Blob#JavaScript).
```
jimblanc, line 40: Just to double check, this is being removed because it's not longer relevant for our supported browsers?

src/fragments/lib/storage/js/download.mdx
jimblanc: +1, I think the change at the top of the page is adequate

jimblanc (commented):
Looks good, some nits

src/fragments/lib/storage/js/download.mdx
kvramyasri7: Yes, Correct currently .text() method is being supported in the browsers.

src/fragments/lib/storage/js/download.mdx
kvramyasri7: Thanks for the suggestion.

src/fragments/lib/storage/js/download.mdx
kvramyasri7: This is a good question. If we have set download to true it would return complete Blob object in the API call response. In order to avoid loading complete object with get call we recommend setting download flag to false. please let me know if have any concerns.

src/fragments/lib/storage/js/download.mdx
```diff
@@ -137,11 +139,19 @@ You can use `expires` option to limit the availability of your URLs. This config
 await Storage.get('filename.txt', { expires: 60 });
 ```
 
+### Check for existence of a file
+
+You can check for the existence of a file in the storage using the `validateObjectExistence` option. When enabled a `get` call will return a pre-signed URL if the file exists and a `404` error if it does not.
```
haverchuck, line 144:
```suggestion
You can check for the existence of a file in the storage category's `get` method using the `validateObjectExistence` option. When this flag is enabled a `get` call will return a pre-signed URL if the file exists and a `404` error if it does not.
```

src/fragments/lib/storage/js/download.mdx
kvramyasri7: Updated. Thanks.

src/fragments/lib/storage/js/download.mdx
abdallahshaban557: I think we need to add a bit of a description on why that is our recommended option. what do you think @jimblanc and @stocaaro ?

jimblanc (approved):
Looks good, just to confirm with the doc team redirects don't appear to be needed in this case?

abdallahshaban557:
@jimblanc - yea, they would not. FYI @aws-amplify/documentation-team - we do not need redirects here.

abdallahshaban557 (approved):
LGTM

katiegoines (approved):

PR #5986, opened 2023-10-09
GH action to check bundlesize of PR

katiegoines:
#### Description of changes:
GitHub action to check bundle size of below selected pages on PR branch vs on main branch. PR to fail if bundle size increases by +5% on any of the selected pages.

- `/cli`
- `/cli/start/install`
- `/lib/auth/getting-started/q/platform/[platform]`
- `/start`
- `/cli`

Example failure from testing on my local fork (with console logs):

Screenshot 2023-10-09 at 12 59 10 PM

### Instructions

- [ ] amplify-cli
- [ ] amplify-ui
- [ ] amplify-studio
- [ ] amplify-hosting
- [ ] amplify-libraries

- [ ] JS
- [ ] iOS
- [ ] Android
- [ ] Flutter
- [ ] React Native

#### Checks

- [ ] Ready to merge

### DIFF
--- a/.github/workflows/scripts/check_bundle_size.js
+++ b/.github/workflows/scripts/check_bundle_size.js
@@ -28,15 +28,20 @@ module.exports = {
   },
 
   compareBundles: (baseBundles, headBundles) => {
-    baseBundles.forEach((page) => {
-      headBundles.forEach((prPage) => {
+    const fails = [];
+    baseBundles.forEach((basePage) => {
+      headBundles.forEach((headPage) => {
         if (
-          page.page == prPage.page &&
-          page.parsedSize * 1.05 < prPage.parsedSize
+          basePage.page == headPage.page &&
+          basePage.parsedSize * 1.05 < headPage.parsedSize
         ) {
-          return true;
+          fails.push(basePage.page);
         }
       });
     });
+    console.log(
+      `The bundle size of ${fails} increased by more than 5% with this PR`
+    );
+    return fails.length;
   }
 };
--- a/.github/workflows/scripts/check_bundle_size.js
+++ b/.github/workflows/scripts/check_bundle_size.js
@@ -28,12 +28,15 @@ module.exports = {
   },
 
   compareBundles: (baseBundles, headBundles) => {
-    for (let i = 0; i < baseBundles.length; i++) {
-      if (baseBundles[i] !== headBundles[i]) {
-        if (baseBundles[i].parsedSize * 1.05 < headBundles[i].parsedSize) {
-          return false;
+    baseBundles.forEach((page) => {
+      headBundles.forEach((prPage) => {
+        if (
+          page.page == prPage.page &&
+          page.parsedSize * 1.05 < prPage.parsedSize
+        ) {
+          return true;
         }
-      }
-    }
+      });
+    });
   }
 };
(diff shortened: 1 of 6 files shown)


.github/workflows/scripts/check_bundle_size.js
```diff
@@ -0,0 +1,67 @@
...
+      }
+    });
+    return headBundleSizes;
+  },
+
+  compareBundles: (baseBundles, headBundles) => {
+    for (let i = 0; i < baseBundles.length; i++) {
+      if (baseBundles[i] !== headBundles[i]) {
+        if (baseBundles[i].parsedSize * 1.05 < headBundles[i].parsedSize) {
+          return false;
```
jacoblogan, line 62: should this be `return true` it looks like the final steps condition `if: ${{ steps.compare-bundles.outputs.result }}` is looking for a value of `true` to block the PR or is it just any non null value?

.github/workflows/scripts/check_bundle_size.js
```diff
@@ -0,0 +1,67 @@
...
+          totalStatSize: page.totalStatSize
+        });
+      }
+    });
+    return headBundleSizes;
+  },
+
+  compareBundles: (baseBundles, headBundles) => {
+    for (let i = 0; i < baseBundles.length; i++) {
+      if (baseBundles[i] !== headBundles[i]) {
```
jacoblogan, line 60: Is the purpose of this `if` statement just to skip over the comparison if the same object is passed into the function as `baseBundles` and `headBundles`?  Are we guaranteed that the pages Array generated will always be in the same order or is there a situation where the filter function from the `checkBundleSize` function produces arrays that are in a different order between head and base (such as pages being added or removed)?

.github/workflows/scripts/check_bundle_size.js
katiegoines: It's looking for any non-null value, but `true` is better. Changing

package.json
```diff
@@ -20,6 +20,7 @@
     "@emotion/react": "^11.1.5",
     "@emotion/styled": "^11.3.0",
     "@mdx-js/mdx": "^2.0.0",
+    "@next/bundle-analyzer": "^13.5.3",
```
timngyn, line 23: Could we move this into the `DevDependencies`? I think the workflow should still work if we move it there and that way this wouldn't be required for the "production" build

.github/workflows/scripts/check_bundle_size.js
```diff
@@ -0,0 +1,67 @@
+module.exports = {
+  checkBaseBundleSize: () => {
+    const fs = require('fs');
+    const jsonString = fs.readFileSync('.github/analyze/bundles.json');
```
timngyn, line 4: Sorry if I missed it somewhere, but is this file getting generated from a step or something like that?

timngyn: Ah nvm, I see it now, you make the build script run with the `ANALYZE` env var, sorry about that!

.github/workflows/scripts/check_bundle_size.js
katiegoines: Fixed this logic to be able to console.log any problem pages.

.github/workflows/scripts/check_bundle_size.js
katiegoines: Fixed this logic to check more explicitly.

jacoblogan (approved):

timngyn (approved):

PR #5994, opened 2023-10-10
[DRAFT] - Initial design updates

hbuchel:
#### Description of changes:

https://next-release-design-update.d2bfwhpcsj9awv.amplifyapp.com/

### Instructions

- [ ] amplify-cli
- [ ] amplify-ui
- [ ] amplify-studio
- [ ] amplify-hosting
- [ ] amplify-libraries

- [ ] JS
- [ ] iOS
- [ ] Android
- [ ] Flutter
- [ ] React Native

#### Checks

- [ ] Ready to merge


src/components/GlobalNav/GlobalNav.module.scss
hbuchel, line 1: I realize the changes here means we can't do a direct copy/paste to other sites but... are we really going to ever do that? :)

src/components/TestNav/TestNav.tsx
hbuchel, line 1: This is just here for testing.

src/pages-old/ChooseFilterPage.tsx
hbuchel, line 1: This will eventually go away but needed to remove some of the theme-ui stuff.

src/pages/[platform]/index.tsx
Comment (hbuchel) [inline src/pages/[platform]/index.tsx:1]: This content is just for me testing scrollability with the sticky headers.

src/pages/index.tsx
```diff
@@ -22,73 +11,59 @@ const meta = {
...
-        <TestNav />
-        <View>{page}</View>
-      </Flex>
-      <Footer />
-    </>
+    <Layout
+      pageTitle={meta.title}
+      pageDescription={meta.description}
+      url={meta.url}
+      pageType="home"
```
hbuchel, line 64: The home page will have some "special" styling versus other pages, so I added a pageType prop to Layout that we can use to stick some classes where we need.

src/themes/defaultTheme.ts
hbuchel, line 1: This is WIP but I wanted to get at least a base theme setup for us to work with.

yarn.lock
hbuchel, line 1: Removing some dependencies!

package.json
hbuchel, line 1: We'll remove emotion also, but some components need to be updated first.

hbuchel (commented):
Some changes:

jacoblogan:
It looks like the plan is to use scss files in the `/styles` directory for styling.  I'm all for moving away from emotion, but is there a reason to move away from cssModules as well.  Some of the benefits of cssModules include scoping of css and co-locating styles and code and the main downside that I see is having one more dependency and step in the build.  These seem like a reasonable tradeoff to improve the developer experience, but are there other downsides to using cssModules?

hbuchel:
> It looks like the plan is to use scss files in the `/styles` directory for styling. I'm all for moving away from emotion, but is there a reason to move away from cssModules as well. Some of the benefits of cssModules include scoping of css and co-locating styles and code and the main downside that I see is having one more dependency and step in the build. These seem like a reasonable tradeoff to improve the developer experience, but are there other downsides to using cssModules?

Mainly, I don't think the docs site is complex enough to warrant css modules. We're not importing multiple design systems (anymore) or 3rd party custom components to really worry about name clashing to that extent where we truly need scoped styles. I think having to find the styles in a separate folder inside of /styles, where it can still be named appropriate to the component, doesn't prohibit the experience enough to need to layer on css modules. It's maybe, IMO, a better developer experience, as we look to make sure our Docs site has a more consistent style going forward, to colocate the component styles together. It also seemed to be a simpler solution to go with plain css/scss after removing emotion, rather than replacing it with a new convention for class naming in our components. It's probably less of a dependency issue, since Next.JS is already loading in the utility for css modules, postcss, scss, etc. anyways, and moreso simplifying so we have a good base going forwards with.

src/styles/footer.scss
```diff
@@ -0,0 +1,26 @@
+.footer {
```
timngyn, line 1: Should we add styling so that the footer is always at the bottom irrespective of the content height?

timngyn (dismissed):
Looks pretty good to me!! Thanks @hbuchel 🙌 

Just had some comments, but they aren't blocking or anything like that

src/styles/footer.scss
hbuchel: good point, will update!

src/themes/gen2Theme.ts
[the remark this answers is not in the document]
hbuchel: I think they want the gen2 theme to be purple, basically. But let me remove this for now till we have that direction. I *think* we'll essentially just use a new layout for those pages, and wrap that layout in a Themeprovider that uses it's own theme versus the default docs theme.

katiegoines (approved):

katiegoines (approved):

timngyn (approved):

PR #6005, opened 2023-10-12
Feature list component

sowmyadev-kota:
#### Description of changes:

This change will add `feature-list` component to Home component.

Staging site: https://feature-list-component.d1pti72ynkspxb.amplifyapp.com/start/

### Instructions

- [ ] amplify-cli
- [ ] amplify-ui
- [ ] amplify-studio
- [ ] amplify-hosting
- [ ] amplify-libraries

- [ ] JS
- [ ] iOS
- [ ] Android
- [ ] Flutter
- [ ] React Native

#### Checks

- [ ] Ready to merge

### DIFF
--- a/src/components/CategoryFeature/CategoryFeature.tsx
+++ b/src/components/CategoryFeature/CategoryFeature.tsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import { Flex, View, Heading } from '@aws-amplify/ui-react';
+
+interface CategoryFeatureProps {
+  category: string;
+  children: React.ReactNode;
+}
+const CategoryFeature: React.FC<CategoryFeatureProps> = ({
+  category,
+  children
+}) => {
+  return (
+    <Flex
+      direction="column"
+      justifyContent="flex-start"
+      alignItems="stretch"
+      alignContent="flex-start"
+      wrap="nowrap"
+      gap="1rem"
+      paddingRight={'xxl'}
+      as="ul"
+      className="category-list"
+    >
+      <View as="li" columnSpan={2}>
+        <Heading
+          level={4}
+          textDecoration="none"
+          marginBottom="-0.75rem"
+        >{`${category}`}</Heading>
+      </View>
+      {children}
+    </Flex>
+  );
+};
+
+export default CategoryFeature;
--- a/src/components/CategoryFeatureLink/CategoryFeatureLink.tsx
+++ b/src/components/CategoryFeatureLink/CategoryFeatureLink.tsx
@@ -1,9 +1,9 @@
 import React from 'react';
-import { Link } from '@aws-amplify/ui-react';
+import { Link, View } from '@aws-amplify/ui-react';
 
 interface CategoryFeatureLinkProps {
-  isExternal?: boolean;
-  href?: string;
+  isExternal: boolean;
+  href: string;
   children: React.ReactNode;
 }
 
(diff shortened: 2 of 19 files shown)


src/components/CategoryFeature/CategoryFeature.tsx
```diff
@@ -0,0 +1,36 @@
...
+      className="category-list"
+    >
+      <View as="li" columnSpan={2}>
+        <Heading
+          level={4}
+          textDecoration="none"
+          marginBottom="-0.75rem"
+        >{`${category}`}</Heading>
+      </View>
+      {children}
```
timngyn, line 31: Would it make sense to have the children list elements be nested as a new `ul` within your `View` component here so that there is more of a hierarchy? I think this would convey that the links belong to the category that way.

For instance there's an example here on MDN: https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/HTML_text_fundamentals#nesting_lists

src/components/CategoryFeatureLink/CategoryFeatureLink.tsx
```diff
@@ -0,0 +1,29 @@
...
+  return (
+    <View as="li">
+      <Link
+        href={href}
+        className="external-link"
+        isExternal={isExternal}
+        textDecoration={'underline'}
+      >
+        {children}
+      </Link>
```
timngyn, line 24: You don't have to change this now, but we might have to change this to be the `Link` component from `next/link` later so that it can navigate to the pages in the docs app.

I saw below we're hardcoding the platform as "javascript", but what happens if we're on a different homepage for a different platform? How do we build out the correct links from the platform that gets passed into the `PlatformFreatures` component?

src/components/Page/metaContent.tsx
```diff
@@ -63,8 +69,167 @@ export default function MetaContent({
...
+                  isExternal={true}
+                  href="https://docs.amplify.aws/"
+                >
+                  {'Enable Signup sign-in, and sign-out'}
+                </CategoryFeatureLink>
+                <CategoryFeatureText>
+                  {
+                    'Lorem Ipsum is simply dummy text of the printing and typesetting industry.'
+                  }
+                </CategoryFeatureText>
```
timngyn, line 103: What's the reason for separating out the `CategoryFeatureText` and `CategoryFeatureLink` into different components?

src/components/PlatformFeatures/PlatformFeatures.tsx
```diff
@@ -0,0 +1,32 @@
...
+      }}
+      gap="small"
+    >
+      <View columnSpan={2}>
+        <Heading
+          level={2}
+          fontWeight={'bold'}
+        >{`Features for ${platform}`}</Heading>
+      </View>
+      {children}
```
timngyn, line 27: Should the children fit into one column for the `base` screen sizes? When I adjust my browser window, it looks like there are still two columns instead of everything fitting into one:

image

src/components/Page/metaContent.tsx
sowmyadev-kota: I was matching the expected developer experience, I can change it to a single component.

src/components/PlatformFeatures/PlatformFeatures.tsx
sowmyadev-kota: Good catch, will keep this  in mind! thank you.

src/components/CategoryFeatureLink/CategoryFeatureLink.tsx
sowmyadev-kota: Since it is a `prop` we can pass the` platform` to the component, javascript should be default, will change thanks.
Good, catch will discuss about the `Link` in offline, thanks

src/components/Page/metaContent.tsx
timngyn: Sorry, what do you mean by the that Sowmya? I'm not sure what the expected developer experience is, so if it makes sense to split them out then that's okay with me

I was just thinking that since the `CategoryFeatureLink` is related to the `CategoryFeatureText` then it would make sense to have them grouped together

Let me know what you think though

PR #6033, opened 2023-10-17
General layout updates, add spaceship, add debounce to scroll event o…

hbuchel:
#### Description of changes:

- General layout/style updates
- Added a scroll event to home page that changes the search background from transparent to solid white. Only the home page needs this due to the spaceship illustration. Added a small debounce utility for scroll event.
- Adds spaceship

https://layoutdesignupdates.d2bfwhpcsj9awv.amplifyapp.com/

### Instructions

- [ ] amplify-cli
- [ ] amplify-ui
- [ ] amplify-studio
- [ ] amplify-hosting
- [ ] amplify-libraries

- [ ] JS
- [ ] iOS
- [ ] Android
- [ ] Flutter
- [ ] React Native

#### Checks

- [ ] Ready to merge


src/components/ExternalLink/index.tsx
hbuchel, line 1: I think we need to clean up our usage of "Link" type components. We have this, next/Link, and Amplify UI Link (which has isExternal prop). Should we always be using next/Link unless it's external?

hbuchel, line 1: I think we need to clean up our usage of "Link" type components. We have this, next/Link, and Amplify UI Link (which has isExternal prop). Should we always be using next/Link unless it's external?

src/pages/index.tsx
hbuchel, line 1: The home page is the only page that needs this scroll event for the search bar background color. I added a very basic debounce utility so that there is some delay between checking.

src/styles/spaceship.scss
hbuchel, line 1: .spaceship-inner is a (css size) container for .spaceship. This lets us make the spaceship size adjustments respond to the width of .spaceship-inner, instead of the viewport. Yay container queries!

src/components/ExternalLink/index.tsx
timngyn: Would it make sense to refactor this component to use the Amplify UI `Link` with the `isExternal` prop?

For your question, I'm leaning towards yes to using `next/Link` unless it's external. That way we get the benefits of benefits of the pre-fetching and the client-side navigation defaults that come with next?

src/pages/index.tsx
```diff
@@ -17,6 +18,22 @@ export default function Page() {
...
+  const handleScroll = debounce((e) => {
+    const bodyScroll = e.target.documentElement.scrollTop;
+    if (bodyScroll > 50) {
+      document.body.classList.add('scrolled');
+    } else if (document.body.classList.contains('scrolled')) {
+      document.body.classList.remove('scrolled');
+    }
+  });
+
+  React.useLayoutEffect(() => {
```
timngyn, line 30: What's the reason for using `React.useLayoutEffect` instead of importing it?

timngyn (approved):
LGTM! Had one suggestion about the `ExternalLink` component but it's not blocking

src/pages/index.tsx
hbuchel: I think you typically want to add event listeners inside of useEffect or useLayoutEffect so that it can also be removed and you don't end up with weird side effects of the event handler persisting.

And (I think, anyways) that useLayoutEffect is preferred for events like scroll events, measuring UI elements via getBoundingClientRect(), etc.

src/components/ExternalLink/index.tsx
hbuchel: Good point! I'll address the ExternalLink refactoring in a different PR if that's ok.

timngyn: Sounds good!

src/pages/index.tsx
timngyn: Ah sorry I wasn't clear. I just noticed that you had the `React.` before the `useLayoutEffect`. Looking at the imports it looks like we're already importing the `useEffect` and so I was wondering if there was a reason for accessing `useLayoutEffect` from within the `React` import.

hbuchel: oh! lollll my bad.  Good callout, I tidied those up in the latest.

timngyn (approved):

sowmyadev-kota (approved):

PR #6101, opened 2023-10-28
Next release/breadcrumbs

jacoblogan:
#### Description of changes:
Add breadcrumbs component

Staging site: https://next-release-breadcrumbs.dmhjrabi3pkql.amplifyapp.com/

### Instructions

- [ ] amplify-cli
- [ ] amplify-ui
- [ ] amplify-studio
- [ ] amplify-hosting
- [ ] amplify-libraries

- [ ] JS
- [ ] iOS
- [ ] Android
- [ ] Flutter
- [ ] React Native

#### Checks

- [ ] Ready to merge


src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+        {items?.map(({ href, label }, i) => {
+          const isCurrent = i === items.length - 1;
+          return (
+            <Breadcrumbs.Item key={href}>
+                <Link href={href} passHref>
+                  <Breadcrumbs.Link isCurrent={isCurrent}>
+                    {label}
+                  </Breadcrumbs.Link>
+                </Link>
+              {isCurrent ? null : <Breadcrumbs.Separator />}
```
timngyn, line 54: I think for the separator, we want to use the left angle bracket `>`. Could we do that with the custom way? https://ui.docs.amplify.aws/react/components/breadcrumbs#custom-separator

src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  urls.splice(0,1,"/");
+
+  urls.forEach((url) => {
+    const directoryEntry = findNode(url);
+    breadcrumbs.push({
+      href:{
+        pathname: url,
+        query: { platform: platform }
+      },
+      label: directoryEntry ? directoryEntry.title : url
```
timngyn, line 33: Should we add some logic here to catch the root path and show "Home" instead of "Amplify Docs"? I think from the designs it originally has "Home"

I think also it should only show when we're visited a path with depth 1? So it wouldn't show up on the "Home" page, but would show up on "Home > Get started" for instance

timngyn (commented):
Looking good!! Just had a comment about whether or not we need to display the breadcrumbs when on the "root" or home page

hbuchel (commented):
Just one comment:
- The first link seems to be adding the query parameter for platform as a ? query string parameter in the link, like:
`https://next-release-breadcrumbs.dmhjrabi3pkql.amplifyapp.com/?platform=javascript`

src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  label: string;
+  isDisabled?: boolean;
+};
+
+type Props = {
+  route: string;
+  platform: string;
+};
+
+const overrides = {
```
jacoblogan, line 16: Not sure what kind of usage this will get but a basic override system, currently used to turn '/' to 'Home' but in the future might be used for turning things like '/android/prev' to 'V0' while also changing '/javascript/prev' to 'V6(preview)'  Anyway we can remove this if you all want and just put in a straight replace '/' for 'Home' and nothing else

src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  '/' : 'Home'
+}
+
+function generateBreadcrumbs(route:string, platform:string):BreadcrumbItem[]{
```
sowmyadev-kota, line 20: Thank you  for doing this @jacoblogan!

sowmyadev-kota (approved):

timngyn (approved):

sowmyadev-kota (approved):

PR #6114, opened 2023-10-30
[WIP] Last edited remark plugin

timngyn:
### Instructions

- [ ] amplify-cli
- [ ] amplify-ui
- [ ] amplify-studio
- [ ] amplify-hosting
- [ ] amplify-libraries

- [ ] JS
- [ ] iOS
- [ ] Android
- [ ] Flutter
- [ ] React Native

#### Checks

- [ ] Ready to merge


src/plugins/frontmatter.mjs
timngyn, line 1: For this file, I was trying to modify the `meta` object to include a new `lastUpdated` property, thinking that if I could modify the AST in the correct way then it would be reflected in `getStaticProps` since we pass down the `meta` object. This doesn't work though even though we modify the AST before `getStaticProps` runs.

I console.log() the AST and the `meta` object in `getStaticProps`, but the change never showed up there

src/plugins/frontmatter.tsx
timngyn, line 1: In the old way, we added actual front matter to each file with the `thematicBreaks` and placed the last updated date between them. In this remark plugin, we locate the lines between the thematicBreaks and create a new exported variable called `frontmatter`. After this, we remove the frontmatter from the markdown so that it doesn't get displayed/used anywhere else

In the Page plugin, we then pass the `frontmatter` object as a prop so that we can access the values inside the `frontmatter` object

PR #6149, opened 2023-11-01
Add cononical link tag

timngyn:
#### Description of changes:
- Add list of canonical urls based on their "route"
- Add some logic in `_app.tsx` to set the link with `rel='canonical'` based on if the current route is part of the list

To test:
1. Go to https://canonical-urls.d1ywzrxfkb9wgg.amplifyapp.com/android/tools/cli/commands/
2. Inspect the page source and verify that the canonical link is for the `javascript` platform

### Instructions

- [ ] amplify-cli
- [ ] amplify-ui
- [ ] amplify-studio
- [ ] amplify-hosting
- [ ] amplify-libraries

- [ ] JS
- [ ] iOS
- [ ] Android
- [ ] Flutter
- [ ] React Native

#### Checks

- [ ] Ready to merge


src/pages/_app.tsx
```diff
@@ -31,6 +33,14 @@ function MyApp({ Component, pageProps }) {
         {page}
       </Layout>
     ));
+
+  const router = useRouter();
+  let canonicalUrl = 'https://docs.amplify.aws';
```
timngyn, line 38: This is hardcoded to the prod domain. I was thinking about maybe pulling this from an env variable so that we could pass down different domains like the staging ones. But I'm not sure if we would really need this often or not. Open to any feedback on it

src/data/canonical-urls.ts
timngyn, line 1: Adding this list of URLs using the `[platform]` route. Since we want to use `javascript` as the canonical url I thought it would be easier to just use this path so that we can match it with `router.pathname`, which has the exact format as the URLs in this list (including the `[platform]` string).

When visiting any of these urls with the real platform, like android, ios, etc., the `router.pathname` will be the same since for all of the them because we are showing the same content/file.

src/pages/_app.tsx
```diff
@@ -31,6 +33,14 @@ function MyApp({ Component, pageProps }) {
...
+  const canonicalPath = CANONICAL_URLS.includes(router.pathname)
+    ? router.pathname.replace('[platform]', 'javascript')
+    : router.asPath;
```
timngyn, line 41: If the current `router.pathname` doesn't exist in the list, then we just use the `router.asPath` as the canonical URL. `router.asPath` will be the url that includes the current platform

hbuchel (approved):
Makes sense to me! Poked around to different URLS and on /gen2 and looks like it's working as expected.

jacoblogan (approved):
LGTM, I wonder if there will be a better way to maintain this list going into the future.  If we can rely on full branches of the tree being platform independent then we could do this simply based off the front of the url i.e. `/[platform]/tools` that way new pages added to `/tools` would not also need to be added to this list.

timngyn:
@jacoblogan 

> LGTM, I wonder if there will be a better way to maintain this list going into the future. If we can rely on full branches of the tree being platform independent then we could do this simply based off the front of the url i.e. `/[platform]/tools` that way new pages added to `/tools` would not also need to be added to this list.

That's a good point. I can look into this some more. Might just need to separate the logic out into it's own component and basically just check if the pathname starts with`/[platform]/tools`?

PR #6291, opened 2023-11-10
Add margin top and bottom for codeblock when placed in a callout

timngyn:
#### Description of changes:
- Add margin top and bottom around codeblock when placed in a callout

Staging site: https://codeblock-padding-callout.d1ywzrxfkb9wgg.amplifyapp.com/

To test:
1.  Go to https://codeblock-padding-callout.d1ywzrxfkb9wgg.amplifyapp.com/javascript/build-a-backend/more-features/predictions/example-app/

### Instructions

- [ ] amplify-cli
- [ ] amplify-ui
- [ ] amplify-studio
- [ ] amplify-hosting
- [ ] amplify-libraries

- [ ] JS
- [ ] iOS
- [ ] Android
- [ ] Flutter
- [ ] React Native

#### Checks

- [ ] Ready to merge

### DIFF
--- a/src/styles/callout.scss
+++ b/src/styles/callout.scss
@@ -19,4 +19,9 @@
   &__content {
     min-width: 0;
   }
+
+  .pre-wrapper {
+    /* Class for code block */
+    margin: var(--amplify-space-large) 0;
+  }
 }


src/styles/callout.scss
```diff
@@ -19,4 +19,9 @@
   &__content {
     min-width: 0;
   }
+
+  .pre-wrapper {
+    /* Class for code block */
+    margin: var(--amplify-space-large) 0;
+  }
```
timngyn, line 26: Is this the way to do this? I was looking for some examples where the callout had multiple paragraphs and saw that each paragraph was very close to each other too. One other thing we could do is modify `Callout` and instead of wrapping the `children` within a `View`, we could use `Flex` instead and have it flex column with the gap of `--amplify-space-large`

Would this approach be better?

Some examples of multiple paragraphs in a callout:
- https://next-docs.amplify.aws/javascript/tools/cli-legacy/graphql-transformer/overview/
   - Right above the "Rebuild GraphQL API" heading
- https://next-docs.amplify.aws/gen2/build-a-backend/data/mutate-data/#update-an-item

hbuchel: I'm ok with this implementation. I think I did have a Flex with column direction originally, but there were instances where there were elements that were inline that were then flexing vertically so it got weird (sorry I don't remember the page now).  In that case, authors would always have to take care to make sure things were wrapped in another div or p tag, so I think this solution works.

hbuchel (approved):

reesscot (approved):

PR #6360, opened 2023-11-14
Mobile SDK Review 

salihgueler:


src/directory/directory.mjs
```diff
@@ -1644,57 +1653,50 @@ export const directory = {
...
-            {
-              path: 'src/pages/[platform]/sdk/info/index.mdx',
-              children: [
-                {
-                  path: 'src/pages/[platform]/sdk/info/app-uninstall/index.mdx'
-                },
-                { path: 'src/pages/[platform]/sdk/info/overview/index.mdx' }
-              ]
-            },
+            // Uncommented because this feature is not shown
```
heatheramz, line 1679: @salihgueler Looks like these pages only exsit for Swift - https://docs.amplify.aws/sdk/info/overview/q/platform/ios/ Can you check that they were imported and have the correct platform filter on them? Once that is done I think this can be unhidden

salihgueler: Can we then limit this to swift here as well? Reverted the comment btw.

heatheramz (approved):
LGTM

renebrandel (approved):

PR #6797, opened 2024-01-17
Add tests for getCustomStaticPath

timngyn:
#### Description of changes:
- Add tests for `getCustomStaticPath`

### Instructions

- [ ] amplify-cli
- [ ] amplify-ui
- [ ] amplify-studio
- [ ] amplify-hosting
- [ ] amplify-libraries

- [ ] JS
- [ ] Swift
- [ ] Android
- [ ] Flutter
- [ ] React Native

#### Checks

- [ ] Ready to merge

### DIFF
--- a/src/utils/__tests__/getCustomStaticPath.test.ts
+++ b/src/utils/__tests__/getCustomStaticPath.test.ts
@@ -1,16 +1,6 @@
 import { getCustomStaticPath } from '../getCustomStaticPath';
-import { PLATFORMS } from '@/data/platforms';
 
 describe('getCustomStaticPath', () => {
-  it('should handle a single platform string', () => {
-    const result = getCustomStaticPath('angular');
-
-    expect(result).toEqual({
-      paths: [{ params: { platform: 'angular' } }],
-      fallback: false
-    });
-  });
-
   it('should handle multiple platforms as an array', () => {
     const result = getCustomStaticPath(['angular', 'flutter']);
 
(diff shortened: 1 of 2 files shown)


src/utils/__tests__/getCustomStaticPath.test.ts
```diff
@@ -0,0 +1,34 @@
...
+    expect(result).toEqual({
+      paths: [
+        { params: { platform: 'angular' } },
+        { params: { platform: 'flutter' } }
+      ],
+      fallback: false
+    });
+  });
+
+  it('should handle the "all" platforms string', () => {
```
hbuchel, line 26: Sorry this is maybe outside of the scope of adding a test for this, but I think this 'all' logic in getCustomStaticPath() was a copy/paste from what I stole from the UI library. 😬 Do you think we should remove it since we never implemented using it? (like, it MIGHT work, but I'm pretty sure we don't have any pages to verify using 'all' doesn't break anything else). In which case, we wouldn't need this test.

hbuchel (commented):
Left a question about our getCustomStaticPath implementation.

src/utils/__tests__/getCustomStaticPath.test.ts
timngyn: @hbuchel gotcha, yea I can clean up the logic for this and update my test. For the logic, this means that we would just be checking if the platforms array is actually an array right?

hbuchel (approved):
Thanks for the cleanup!

katiegoines (approved):

PR #7500, opened 2024-05-06
feat: add Video component

hbuchel:
#### Description of changes:

Adds a `<Video />` component to replace manually adding HTML video element. Implementation and preview links can be found in this separate PR https://github.com/aws-amplify/docs/pull/7503 that converts all inline video to use this component.

**Usage**
```
<Video src="path/to/video.mp4" />

// Override width
<Video width="40%" src="path/to/video.mp4" />
```

- Adds consistent styling to video elements.
- Defaults videos to autoplay, muted, 100% width, playsinline, and loop. Authors can still pass these as attributes/props to override the defaults.
- Addresses [WCAG SC 2.2.2](https://www.w3.org/WAI/WCAG21/Understanding/pause-stop-hide.html) "Pause, Stop, Hide" by adding video controls. Additionally, video will only auto play if user has not checked prefers-reduced-motion in their operating system.
- Defaults video to `aria-hidden="true"`. When reviewing video usage, it's most often the case that we are showing a video that closely matches a list of instructions in text content. It's likely that if we add a text description for the video, it will be redundant and closely match text in the surrounding content and then create more noise for screen readers. In this scenario, it should be safe to hide the videos from screen readers.

### Instructions

- [ ] amplify-cli
- [ ] amplify-ui
- [ ] amplify-studio
- [ ] amplify-hosting
- [ ] amplify-libraries

- [ ] JS
- [ ] Swift
- [ ] Android
- [ ] Flutter
- [ ] React Native

#### Checks

- [ ] Ready to merge


jacoblogan (dismissed):

src/components/Video/Video.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const [prefersReducedMotion, setPrefersReducedMotion] =
+    useState<boolean>(true);
+
+  const videoRef = useRef<HTMLVideoElement>(null);
+
+  useEffect(() => {
+    const mediaQueryList = window.matchMedia(reducedMotionMediaQuery);
+
+    setPrefersReducedMotion(
+      !window.matchMedia(reducedMotionMediaQuery).matches
```
timngyn, line 42: nit: Could we reuse mediaQueryList from above for this line?
```suggestion
      !mediaQueryList.matches
```

src/components/Video/__tests__/Video.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+    }))
+  });
+}
+
+// Mock HTMLMediaElement.prototype.pause
+jest
+  .spyOn(window.HTMLMediaElement.prototype, 'pause')
+  .mockImplementation(() => {});
+
+const playSpy = jest
```
jacoblogan, line 27: I don't think playSpy is ever being reset, making these tests (callCounts) order dependent.

src/components/Video/Video.tsx
hbuchel: Thanks for the catch, fixed in latest

timngyn (approved):

src/components/Video/__tests__/Video.test.tsx
hbuchel: Ah thanks, I think you're right, and also explains why I was having trouble with a test earlier. In the latest I added a:
```
afterEach(() => {
    jest.clearAllMocks();
  }); 
```
And updated the tests to be a little more specific (one tests with the media query set to true, one now explicitly sets it to false in the test).

jacoblogan (approved):

Issue #7594, opened 2024-05-15
Gen 2 Set up a Function yields compilation error


sobolk:
**Describe the bug**

When following https://docs.amplify.aws/react/build-a-backend/functions/set-up-function/ I ended up with compilation error.

**To Reproduce**
1. `mkdir testapp023`
2. `cd testapp023`
3. `npm create amplify@latest`
4. Follow https://docs.amplify.aws/react/build-a-backend/functions/set-up-function/
5. Observe compilation error at `npx ampx sandbox`

**Expected behavior**
There is no compilation error

**Screenshots**

![image](https://github.com/aws-amplify/docs/assets/5849952/70ea2b6d-d66c-451a-8c71-28ee8c0c722a)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

josefaidt:
this is now resolved

PR #7595, opened 2024-05-15
Adds instruction to invoke a function via its url

johnpc:
#### Description of changes:

This PR updates the [Set up a Function](http://docs.amplify.aws/react/build-a-backend/functions/set-up-function/#pageMain) page to include instructions on how to to invoke the function you've created via a function url.

We may prefer to edit this to put the function behind the data schema instead, but this option works. It also fixes the existing type issue in https://github.com/aws-amplify/docs/issues/7594 because it is able to type the function event as a `LambdaFunctionURLEvent`.

#### Related GitHub issue #, if available:

Fixes https://github.com/aws-amplify/docs/issues/7594

#### Checks

- [x] Does this PR conform to [the styleguide](https://github.com/aws-amplify/docs/blob/main/STYLEGUIDE.md)?

- [x] Does this PR include filetypes other than markdown or images? Please add or update unit tests accordingly.

- [x] Are any files being deleted with this PR? If so, have the needed redirects been created?

- [x] Are all links in MDX files using the MDX link syntax rather than HTML link syntax? <br /> 

- [x] Ready to merge

### DIFF
--- a/src/pages/[platform]/build-a-backend/functions/set-up-function/index.mdx
+++ b/src/pages/[platform]/build-a-backend/functions/set-up-function/index.mdx
@@ -52,7 +52,8 @@ export const sayHello = defineFunction({
 Next, create the corresponding handler file at `amplify/functions/say-hello/handler.ts`. This is where your function code will go.
 
 ```ts title="amplify/functions/say-hello/handler.ts"
-export const handler = async (event) => {
+import { LambdaFunctionURLEvent } from "aws-lambda";
+export const handler = async (event: LambdaFunctionURLEvent) => {
   // your function code goes here
   return 'Hello, World!';
 };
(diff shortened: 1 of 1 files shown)


jacoblogan (approved):

renebrandel (changes requested):
this shouldn't be the "primary flow", as it'll likely result in a few unexpected situations for the customer. There's no authZ protection, e2e type safety, and customers need to think about "CORS". 

I think we should this "Enable function URLs" as a separate with a callout that we don't recommend this directly.

renebrandel:
Instead the default flow, should go through data.

josefaidt:
Closing this in favor of the data approach with https://github.com/aws-amplify/docs/pull/7759

PR #7659, opened 2024-05-24
Cbonif/single-table-design

chrisbonifacio:
#### Description of changes:

Adds page that goes over single table design in DynamoDB with an Amplify Gen 2 schema example and code snippets

### Instructions

- [ ] amplify-cli
- [ ] amplify-ui
- [ ] amplify-studio
- [ ] amplify-hosting
- [ ] amplify-libraries

- [ ] JS
- [ ] Swift
- [ ] Android
- [ ] Flutter
- [ ] React Native

#### Checks

- [ ] Ready to merge

### DIFF
--- a/src/pages/[platform]/build-a-backend/data/connect-to-existing-data-sources/single-table-design/index.mdx
+++ b/src/pages/[platform]/build-a-backend/data/connect-to-existing-data-sources/single-table-design/index.mdx
@@ -334,8 +334,8 @@ export function request(ctx) {
     query: {
       pk: `class-${classId}`,
       sk: "results",
-      index: "LocalSecondaryIndex",
     },
+    index: "LocalSecondaryIndex",
   });
 }
 


printSamuel:
Past me (1 week ago) would have loved this! 
I'm not sure if this is an officially supported feature, but I think it would be interesting to learn how to use functions like begins_with, greater_than, and between on the sort key. 
It works with the "old schema", like this:

```
export function request(ctx) {
  return {
    version: '2017-02-28',
    operation: 'Query',
    query: {
      expression: 'PK = :PK and begins_with(SK, :SK)',
      expressionValues: util.dynamodb.toMapValues({
        ':PK': ctx.arguments.PK,
        ':SK': ctx.arguments.SK
      }),
    },
  };
}
```
Thanks for your effort. :)

src/pages/[platform]/build-a-backend/data/connect-to-existing-data-sources/single-table-design/index.mdx
```diff
@@ -0,0 +1,490 @@
...
+<BlockSwitcher>
+<Block name="Resolver">
+```ts title="amplify/data/getRaceResultsByRacerId.js"
+import { util } from "@aws-appsync/utils";
+import * as ddb from "@aws-appsync/utils/dynamodb";
+
+export function request(ctx) {
+  const { racerId } = ctx.args;
+
+  return ddb.query({
```
Comment (printSamuel) [inline src/pages/[platform]/build-a-backend/data/connect-to-existing-data-sources/single-table-design/index.mdx:306]: I couldn't find any information about a query command. I tried it and it doesn't seem to work.

> "message": "code.js:7:10: Property 'query' is expected to be a valid condition object",
> I/flutter ( 8117):     "locations": [
> I/flutter ( 8117):       {
> I/flutter ( 8117):         "line": 2,
> I/flutter ( 8117):         "column": 7
> I/flutter ( 8117):       }
> I/flutter ( 8117):     ],
> I/flutter ( 8117):     "path": [
> I/flutter ( 8117):       "querySuE"
> I/flutter ( 8117):     ],
> I/flutter ( 8117):     "errorType": "Code"

Comment (chrisbonifacio) [reply src/pages/[platform]/build-a-backend/data/connect-to-existing-data-sources/single-table-design/index.mdx:306]: Strange. .query should be a valid method based on the AppSync docs:

https://docs.aws.amazon.com/appsync/latest/devguide/tutorial-dynamodb-resolvers-js.html#configure-query

I didn't run into issues deploying or invoking it in a request from the client side

Comment (printSamuel) [reply src/pages/[platform]/build-a-backend/data/connect-to-existing-data-sources/single-table-design/index.mdx:306]: My mistake... Sry
It works!

Comment (sisygoboom) [reply src/pages/[platform]/build-a-backend/data/connect-to-existing-data-sources/single-table-design/index.mdx:301]: This might be useful from the amplify gen 2 docs, seems it's been done differently: https://docs.amplify.aws/react/build-a-backend/data/custom-business-logic/#step-2---configure-custom-business-logic-handler-code

src/pages/[platform]/build-a-backend/data/connect-to-existing-data-sources/single-table-design/index.mdx
```diff
@@ -0,0 +1,486 @@
...
+      classId: a.string().required(),
+    })
+    .returns(a.ref("RacingTable").array())
+    .handler(
+      a.handler.custom({
+        dataSource: "RacingTableDataSource",
+        entry: "./getBestPerformanceByClassId.js",
+      })
+    )
+    .authorization((allow) => [allow.publicApiKey()]),
```
Comment (johnpc) [inline src/pages/[platform]/build-a-backend/data/connect-to-existing-data-sources/single-table-design/index.mdx:269]: why not just `.public()`? In general I don't like the idea of using api keys in our examples / docs because they're a time bomb that causes customer projects to break in the future. Since the default `npx create amplify` includes the auth category, I think it's safe to use everywhere in our docs.

src/pages/[platform]/build-a-backend/data/connect-to-existing-data-sources/single-table-design/index.mdx
```diff
@@ -0,0 +1,486 @@
...
+
+In this example, the frontend application requires these queries:
+
+1. Get results for each race by racer ID.
+2. Get a list of races by class ID.
+3. Get the best performance by racer for a class ID.
+4. Get the list of top scores by race ID.
+5. Get the second-by-second performance by racer for all races.
+
+While implementing this design with multiple DynamoDB tables is possible, it's unnecessary and inefficient. A key goal in querying DynamoDB data is retrieving all required data in a single query request. This is a core concept in NoSQL databases, and single-table design simplifies data management and maximizes query throughput.
```
Comment (johnpc) [inline src/pages/[platform]/build-a-backend/data/connect-to-existing-data-sources/single-table-design/index.mdx:92]: The language makes me wonder why typical Amplify UX does it the  "unnecessary and inefficient" way. It might be worth exploring the tradeoffs of taking this approach. Or, if we really believe this is a superior approach, we should consider adding this feature more natively into our L3 constructs and client libraries.

Comment (sisygoboom) [reply src/pages/[platform]/build-a-backend/data/connect-to-existing-data-sources/single-table-design/index.mdx:92]: It's not widely used partly because:
1) Few people know it exists; Amazon, for whatever reason, hasn't made a huge effort to promote it despite using it themselves on amazon.com.
2) You must define your access patterns upfront; many new projects prefer flexibility over efficiency.
3) It can cause problems if you don't know what you're doing. Having everything on one table is undeniably better, but for novices, wrapping their heads around partitions and profile details, sharing the same virtual space as order history, for example, might be confusing.
4) Denormalisation: This pattern sometimes requires repeated data, which can be challenging considering most people are familiar with relational normalised data.

That being said, it would be nice to see some native code for it, most people who follow this pattern currently define a table externally and use [ElectroDB](https://electrodb.dev/en/core-concepts/introduction/) within a lambda.

src/pages/[platform]/build-a-backend/data/connect-to-existing-data-sources/single-table-design/index.mdx
Comment (sisygoboom) [reply src/pages/[platform]/build-a-backend/data/connect-to-existing-data-sources/single-table-design/index.mdx:301]: This deployed successfully for me and has the bonus of being typesafe. Under the hood I imagine @aws-appsync/utils/dynamodb is just using aws-sdk anyway:
```
// resourse.ts
...
const getMessageByConversationId = defineFunction({
  entry: './getMessagesByConversationId.ts'
})
...
  getMessagesByConversationId: a
    .query()
    .arguments({
      conversationId: a.string().required(),
      lastKey: a.string(),
    })
    .returns(a.ref("ChatTable").array())
    .handler(
      a.handler.function(getMessageByConversationId)
    )
    .authorization(allow => [allow.publicApiKey()]),
```

then my handler:
```
// getMessageByConversationId.ts

import { DynamoDB } from 'aws-sdk';
import type { Schema } from './resource';

const ddb = new DynamoDB.DocumentClient();

type GetMessagesByConversationIdResult = {
  pk: string;
  sk: string;
  type: string;
  message?: string | null;
  userId?: string | null;
  conversationId?: string | null;
  timestamp?: string | null;
}[];

export const handler: Schema["getMessagesByConversationId"]["functionHandler"] = async (
  event
): Promise<GetMessagesByConversationIdResult | null> => {
  const { conversationId, lastKey } = event.arguments;

  const queryInput: DynamoDB.DocumentClient.QueryInput = {
    TableName: "ChatTable",
    KeyConditionExpression: 'pk = :conversationId AND begins_with(sk, :prefix)',
    ExpressionAttributeValues: {
      ':conversationId': `CONVERSATION#${conversationId}`,
      ':prefix': 'MESSAGE#',
    },
    ScanIndexForward: true,
  };

  if (lastKey) {
    queryInput.ExclusiveStartKey = {
      pk: `CONVERSATION#${conversationId}`,
      sk: lastKey,
    };
  }

  try {
    const result = await ddb.query(queryInput).promise();

    // Transform the result.Items into the expected format
    const items: GetMessagesByConversationIdResult = (result.Items || []).map(item => ({
      pk: item.pk as string,
      sk: item.sk as string,
      type: item.type as string,
      message: item.message as string | null,
      userId: item.userId as string | null,
      conversationId: item.conversationId as string | null,
      timestamp: item.timestamp as string | null,
    }));

    return items;
  } catch (error) {
    console.error('Error processing request:', error);
    throw new Error('Failed to process request');
  }
};
```

PR #7882, opened 2024-08-09
a11y- Callouts should use descriptive text for icons

katiegoines:
#### Description of changes:
[Issue] There are meaningful "Info" icons but they are marked as decorative.
[Fix] Since the icon is part of Amplify UI component we cannot remove the aria-hidden without updating that component. We could move those icons inline to the content (and use hasIcon={false} on the component)

Staging: https://a11y-callout-icons.d1egzztxsxq9xz.amplifyapp.com/gen1/react/start/getting-started/

Example of page with protected callout: https://a11y-callout-icons.d1egzztxsxq9xz.amplifyapp.com/react/build-a-backend/data/data-modeling/relationships/

### Instructions

- [ ] amplify-cli
- [ ] amplify-ui
- [ ] amplify-studio
- [ ] amplify-hosting
- [ ] amplify-libraries

- [ ] JS
- [ ] Swift
- [ ] Android
- [ ] Flutter
- [ ] React Native

#### Checks

- [ ] Ready to merge

### DIFF
--- a/src/components/Callout/Callout.tsx
+++ b/src/components/Callout/Callout.tsx
@@ -1,4 +1,5 @@
-import { Message, View } from '@aws-amplify/ui-react';
+import { Flex, Message } from '@aws-amplify/ui-react';
+import { IconInfo } from '../Icons/IconInfo';
 
 interface CalloutProps {
   info?: boolean;
@@ -17,8 +18,14 @@ export const Callout = ({
       variation="filled"
       colorTheme={warning ? 'warning' : 'info'}
       backgroundColor={backgroundColor}
+      hasIcon={false}
     >
-      <View>{children}</View>
+      <Flex>
+        <div role="img" aria-label="info" className="amplify-message__icon">
+          <IconInfo aria-hidden={true} />
+        </div>
+        <div>{children}</div>
+      </Flex>
     </Message>
   );
 };
--- a/src/components/Callout/Callout.tsx
+++ b/src/components/Callout/Callout.tsx
@@ -23,7 +23,7 @@ export const Callout = ({
       <Flex>
         <div className="amplify-message__icon">
           {warning ? (
-            <IconWarning aria-hidden={false} aria-label="warning icon" />
+            <IconWarning aria-hidden={false} aria-label="Warning" />
           ) : (
             <IconInfo aria-hidden={false} aria-label="info icon" />
           )}
--- a/src/components/Callout/Callout.tsx
+++ b/src/components/Callout/Callout.tsx
@@ -28,7 +28,7 @@ export const Callout = ({
             <IconInfo aria-hidden={false} aria-label="Important information" />
           )}
         </div>
-        <div>{children}</div>
+        <div className="amplify-message__content">{children}</div>
       </Flex>
     </Message>
   );
--- a/src/components/Callout/Callout.tsx
+++ b/src/components/Callout/Callout.tsx
@@ -21,8 +21,11 @@ export const Callout = ({
       hasIcon={false}
     >
       <Flex>
-        <div role="img" aria-label="info" className="amplify-message__icon">
-          <IconInfo aria-hidden={true} />
+        <div className="amplify-message__icon">
+          <IconInfo
+            aria-hidden={false}
+            aria-label={warning ? 'warning icon' : 'info icon'}
+          />
         </div>
         <div>{children}</div>
       </Flex>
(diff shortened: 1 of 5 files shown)


jacoblogan (dismissed):

src/components/Callout/Callout.tsx
```diff
@@ -17,8 +18,14 @@ export const Callout = ({
       variation="filled"
       colorTheme={warning ? 'warning' : 'info'}
       backgroundColor={backgroundColor}
+      hasIcon={false}
     >
-      <View>{children}</View>
+      <Flex>
+        <div role="img" aria-label="info" className="amplify-message__icon">
```
hbuchel, line 24: Instead of wrapping this in an IMG role, I think you should be able to pass `aria-hidden="false"` to the Icon and pass it an appropriate aria-label.

src/components/Callout/Callout.tsx
```diff
@@ -17,8 +18,14 @@ export const Callout = ({
...
+          <IconInfo aria-hidden={true} />
```
hbuchel, line 25: I think we should to account for the warning variation as well

hbuchel (commented):
Couple comments about the IMG wrapper and warning variation.

src/components/Callout/Callout.tsx
katiegoines: I've added in labeling to account for the variation however currently only the Info icon is being used regardless of the message type

hbuchel: There is a warning variation that is used for Callout, like here: https://docs.amplify.aws/swift/build-a-backend/add-aws-services/analytics/set-up-analytics/#initialize-amplify-analytics

hbuchel: Although I just checked and we don't have an IconWarning so it would have to be added.

katiegoines: Happy to add it! Had just been attempting to maintain the current appearance on the prod site.

katiegoines: Added the IconWarning and adjusted the logic to account for the variation

src/components/Callout/Callout.tsx
```diff
@@ -17,8 +18,18 @@ export const Callout = ({
       variation="filled"
       colorTheme={warning ? 'warning' : 'info'}
       backgroundColor={backgroundColor}
+      hasIcon={false}
     >
-      <View>{children}</View>
+      <Flex>
+        <div className="amplify-message__icon">
+          {warning ? (
+            <IconWarning aria-hidden={false} aria-label="warning icon" />
```
hbuchel, line 26:
```suggestion
            <IconWarning aria-hidden={false} aria-label="Warning" />
```

These can just say "Warning/Important information". For alt text or aria-labels, you *usually* don't need to add "image of" "icon", etc. In this case we're conveying the context of what the Callout is; that it's important information, or a warning to the developer, not so much that an icon is present.

src/components/Callout/Callout.tsx
```diff
@@ -17,8 +18,18 @@ export const Callout = ({
...
+          ) : (
+            <IconInfo aria-hidden={false} aria-label="info icon" />
```
hbuchel, line 28:
```suggestion
            <IconInfo aria-hidden={false} aria-label="Important information" />
```

src/components/Callout/Callout.tsx
```diff
@@ -17,8 +18,18 @@ export const Callout = ({
...
+          )}
+        </div>
+        <div>{children}</div>
```
hbuchel, line 31:
```suggestion
        <div className="amplify-message__content">{children}</div>
```

I think this needs this class to prevent content from stretching. You can test on this page which currently has code snippets overflowing the warning Callout https://a11y-callout-icons.d1egzztxsxq9xz.amplifyapp.com/react/start/manual-installation/

hbuchel (commented):
Left a note about aria-label content and an issue with missing class name.

jacoblogan (approved):

src/components/Callout/__tests__/Callout.test.tsx
```diff
@@ -29,7 +29,6 @@ describe('Callout', () => {
     );
 
     const styles = getComputedStyle(ele.container.children[0]);
-    console.log(styles);
```
jordanvn, line 32: 🚀

jordanvn (approved):
These changes look good to me. It appears that @hbuchel's feedback has all been addressed and I've confirmed expected functionality locally. This is a nice change for accessibility!

srquinn21 (approved):

PR #8536, opened 2026-03-30
fix: add client-side redirect for moved subscribe-data page

harsh62:
## Summary

- Adds a client-side redirect page at the old `/build-a-backend/data/subscribe-data/` URL that redirects to the new `/frontend/data/subscribe-data/` location
- The page was moved during content restructuring (#8531) and a redirect exists in `redirects.json`, but users are reporting 404s on production — suggesting the Amplify Hosting redirect isn't being applied

## Why the redirect wasn't working

The `redirects.json` file contains the correct redirect rule, but it appears Amplify Hosting isn't picking it up. This could be because:
1. The redirect rules need to be manually synced to the Amplify Console
2. The `<platform>` wildcard syntax may not be matching as expected
3. There may be a deployment lag

## Broader concern

There are **31 redirect rules** in `redirects.json` for pages moved from `/build-a-backend/` to `/frontend/` during content restructuring. If the hosting redirects aren't working, ALL of these could be 404s. This PR fixes the reported URL, but we should investigate the redirect deployment pipeline to fix them all.

## Test plan

- [ ] Visit `/react/build-a-backend/data/subscribe-data/` — should redirect to `/react/frontend/data/subscribe-data/`
- [ ] The actual content page at `/react/frontend/data/subscribe-data/` loads correctly
- [ ] Build passes


harsh62:
Closing this PR — the hacky MDX redirect page is not the right approach.

## Root cause

The `redirects.json` file has the correct redirect rules, but **none of them are working on production**. I tested multiple URLs:

- `/react/build-a-backend/data/subscribe-data/` → 404 (should redirect to `/react/frontend/data/subscribe-data/`)
- `/react/build-a-backend/data/query-data/` → 404 (should redirect to `/react/frontend/data/query-data/`)
- `/react/build-a-backend/auth/connect-your-frontend/sign-in/` → 404 (should redirect to `/react/frontend/auth/sign-in/`)

**All 31 redirect rules from the content restructuring PR (#8531) are broken.** The `redirects.json` file is in the repo but Amplify Hosting isn't applying them.

## Required fix

The redirect rules need to be uploaded to the **Amplify Hosting Console** under "Rewrites and redirects". The `redirects.json` file in the repo is not automatically consumed by Amplify Hosting — it needs to be manually synced or there needs to be a deployment script that pushes these rules to the Amplify API.

PR #8607, opened 2026-06-29
feat: improve AI agent discoverability

harsh62:
## Summary

Improves how AI agents and crawlers discover and consume the Amplify docs, implemented entirely within what the static export + Amplify Hosting can serve today. These changes came out of running the site through an "agent readiness" scan and fixing every gap that has a legitimate, non-misleading solution.

All additions point at content or capabilities that **actually exist** (the generated `llms.txt`/markdown exports, the real `awslabs/agent-plugins` skill, AWS's public managed MCP server, and read-only browser tools backed by real data) — no stub endpoints or fabricated capabilities.

## What's included

**Discoverability**
- **Content Signals in `robots.txt`** — `Content-Signal: search=yes, ai-input=yes, ai-train=yes`.
- **`Link` response headers** (`customHttp.yml`, RFC 8288) advertising the API catalog, agent skills index, MCP server card, and the `llms.txt` index.

**Agent discovery files (generated at build time, alongside `robots.txt`/`sitemap.xml`)**
- **`/.well-known/api-catalog`** (RFC 9727 linkset, `application/linkset+json`) with the required relations mapped to the artifacts the build produces: `service-desc` → `llms-full.txt`, `service-doc` → `llms.txt`, `service-meta` → `sitemap.xml`.
- **`/.well-known/agent-skills/index.json`** (Agent Skills Discovery RFC v0.2.0) advertising the real `amplify-workflow` skill from `awslabs/agent-plugins`. Entries point at the docs/marketplace install page, so no `sha256` is published (the page is install guidance, not a downloadable artifact).
- **`/.well-known/mcp/server-card.json`** describing the public, no-auth **AWS Knowledge MCP Server** (`https://knowledge-mcp.global.api.aws`), which authoritatively indexes Amplify docs. An honest pointer to AWS's managed server — the card explicitly states this site does not host its own MCP server.

**Markdown vending** (the per-page `.md` files are already generated under `/ai/pages/**`)
- **Per-page autodiscovery** — each Gen2 content page's `<head>` emits `<link rel="alternate" type="text/markdown" href="/ai/pages/….md">`, reusing `MarkdownMenu`'s `getMarkdownUrl` mapping and gate (skips gen1/home/overview pages that have no `.md` twin).
- **Correct media type** — `/ai/**/*.md` served as `text/markdown; charset=utf-8`.

**WebMCP (in-browser agent tools)**
- Registers read-only tools via `document.modelContext` (with `navigator.modelContext` fallback) so in-browser AI agents can call the site's key read actions:
  - `get_current_page_markdown` — returns the current page's generated Markdown twin
  - `get_documentation_index` — returns the `llms.txt` index
- Both are backed by content the build already produces (real data, not stubs), feature-detected (silent no-op without WebMCP), and torn down on unmount via `AbortSignal`.

## Out of scope (intentionally)

Several scanned standards require live services / DNS / viewer-request edge compute that don't exist behind a static public docs site, and publishing files for them would mislead agents that trust them:

- **OAuth Protected Resource (`/.well-known/oauth-protected-resource`) and `auth.md`** — both declare that the site's resources are access-controlled and tell agents how to obtain tokens to reach them. `docs.amplify.aws` is fully public with no protected API and no agent auth. The scanner passes on the metadata file alone, but asserting a protected resource that doesn't exist could make agents refuse to read public docs or attempt pointless token flows. Deliberately skipped.
- **Markdown content negotiation (same-URL `Accept: text/markdown`)** — requires reading a request header at the edge. Amplify Hosting rewrites match on path/query only (confirmed by `redirects.json`'s own validator), and the managed CloudFront distribution exposes no viewer-request function hook to this repo. The per-page `<link rel="alternate">` + `text/markdown` content-type above is the static-friendly equivalent (agents get the markdown at a sibling URL). True same-URL negotiation needs a CloudFront Function on a fronting distribution, owned outside this repo.
- **OAuth/OIDC discovery, WebMCP for write actions, DNS-AID, Web Bot Auth, commerce protocols (x402/ACP/UCP/MPP)** — no protected API, no site-side mutating actions, DNS/key infrastructure owned elsewhere, and not an e-commerce site.

## Testing

- New unit tests for the API catalog (incl. RFC 9727 `service-desc`/`service-meta` structure), MCP server card, agent skills index, and the WebMCP component (no-op without API, tool registration, real fetch on execute).
- Full unit suite passes (291 tests); `tsc --noEmit` clean on changed components.
- Verified generated output for `robots.txt`, `api-catalog`, `server-card.json`, and `agent-skills/index.json` end-to-end.

### DIFF
--- a/customHttp.yml
+++ b/customHttp.yml
@@ -18,6 +18,15 @@ customHeaders:
       # Link headers advertise agent-discovery resources (RFC 8288 / RFC 9727):
       # the API catalog, the agent skills index, the MCP server card, and the
       # LLM-friendly documentation index.
+      #
+      # Deliberately kept on the global block. Amplify header patterns are
+      # positive-match only (no exclusion), and with trailingSlash: true the
+      # actual page requests are extensionless directory URLs (e.g.
+      # /react/build-a-backend/), so an `**/*.html` pattern would miss real
+      # page loads and drop the header where agents need it most. The header
+      # therefore also rides asset responses (a few extra bytes); the
+      # missing-file risk this could have amplified is eliminated because the
+      # generators now fail the build if a linked file is not written.
       - key: 'Link'
         value: '</.well-known/api-catalog>; rel="api-catalog", </.well-known/agent-skills/index.json>; rel="agent-skills"; type="application/json", </.well-known/mcp/server-card.json>; rel="mcp-server"; type="application/json", </ai/llms.txt>; rel="service-doc"; type="text/plain"'
   # Serve the API catalog as a linkset (RFC 9727 / RFC 9264) at both the
--- a/customHttp.yml
+++ b/customHttp.yml
@@ -16,9 +16,9 @@ customHeaders:
         value: 'upgrade-insecure-requests;'
       # CSP also set in _document.tsx meta tag
       # Link headers advertise agent-discovery resources (RFC 8288 / RFC 9727):
-      # the API catalog and the LLM-friendly documentation index.
+      # the API catalog, the agent skills index, and the LLM-friendly docs index.
       - key: 'Link'
-        value: '</.well-known/api-catalog>; rel="api-catalog", </ai/llms.txt>; rel="service-doc"; type="text/plain"'
+        value: '</.well-known/api-catalog>; rel="api-catalog", </.well-known/agent-skills/index.json>; rel="agent-skills"; type="application/json", </ai/llms.txt>; rel="service-doc"; type="text/plain"'
   # The API catalog has no file extension, so set its media type explicitly
   # (RFC 9727 / RFC 9264).
   - pattern: '.well-known/api-catalog'
--- a/customHttp.yml
+++ b/customHttp.yml
@@ -15,3 +15,13 @@ customHeaders:
       - key: 'Content-Security-Policy'
         value: 'upgrade-insecure-requests;'
       # CSP also set in _document.tsx meta tag
+      # Link headers advertise agent-discovery resources (RFC 8288 / RFC 9727):
+      # the API catalog and the LLM-friendly documentation index.
+      - key: 'Link'
+        value: '</.well-known/api-catalog>; rel="api-catalog", </ai/llms.txt>; rel="service-doc"; type="text/plain"'
+  # The API catalog has no file extension, so set its media type explicitly
+  # (RFC 9727 / RFC 9264).
+  - pattern: '.well-known/api-catalog'
+    headers:
+      - key: 'Content-Type'
+        value: 'application/linkset+json'
--- a/customHttp.yml
+++ b/customHttp.yml
@@ -20,9 +20,14 @@ customHeaders:
       # LLM-friendly documentation index.
       - key: 'Link'
         value: '</.well-known/api-catalog>; rel="api-catalog", </.well-known/agent-skills/index.json>; rel="agent-skills"; type="application/json", </.well-known/mcp/server-card.json>; rel="mcp-server"; type="application/json", </ai/llms.txt>; rel="service-doc"; type="text/plain"'
-  # The API catalog has no file extension, so set its media type explicitly
-  # (RFC 9727 / RFC 9264).
-  - pattern: '.well-known/api-catalog'
+  # Serve the API catalog as a linkset (RFC 9727 / RFC 9264) at both the
+  # canonical extensionless path (served via a 200-rewrite in redirects.json)
+  # and the underlying .json file.
+  - pattern: '/.well-known/api-catalog'
+    headers:
+      - key: 'Content-Type'
+        value: 'application/linkset+json'
+  - pattern: '/.well-known/api-catalog.json'
     headers:
       - key: 'Content-Type'
         value: 'application/linkset+json'
--- a/customHttp.yml
+++ b/customHttp.yml
@@ -16,12 +16,19 @@ customHeaders:
         value: 'upgrade-insecure-requests;'
       # CSP also set in _document.tsx meta tag
       # Link headers advertise agent-discovery resources (RFC 8288 / RFC 9727):
-      # the API catalog, the agent skills index, and the LLM-friendly docs index.
+      # the API catalog, the agent skills index, the MCP server card, and the
+      # LLM-friendly documentation index.
       - key: 'Link'
-        value: '</.well-known/api-catalog>; rel="api-catalog", </.well-known/agent-skills/index.json>; rel="agent-skills"; type="application/json", </ai/llms.txt>; rel="service-doc"; type="text/plain"'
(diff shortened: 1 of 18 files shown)


tasks/generate-wellknown.mjs
```diff
@@ -0,0 +1,148 @@
...
+export async function writeApiCatalog() {
+  const wellKnownDir = path.join(ROOT_PATH, '.well-known');
+  const catalogPath = path.join(wellKnownDir, 'api-catalog.json');
+
+  try {
+    await fs.mkdir(wellKnownDir, { recursive: true });
+    await fs.writeFile(catalogPath, generateApiCatalog());
+    console.log(`api-catalog written to ${catalogPath}`);
+  } catch (error) {
+    console.error(`Error writing api-catalog to ${catalogPath}:`, error);
```
osama-rizk, line 76: **Swallowed write error ships a green build that advertises 404s.** This `catch` logs and returns, so a failed write still passes. Meanwhile `customHttp.yml` emits a global `Link` header to `/.well-known/api-catalog`, whose links point at `llms-full.txt`/`llms.txt`. If a generator throws, agents follow a `Link` header to a missing file. `writeSitemap`/`writeRobots` do the same, but they aren't advertised in a response header — the blast radius is new here. Consider failing the build on write error (same applies to `generate-agent-skills.mjs`).

src/components/MarkdownMenu/MarkdownMenu.tsx
```diff
@@ -8,7 +8,7 @@ interface MarkdownMenuProps {
   isOverview?: boolean;
 }
 
-function getMarkdownUrl(route: string): string {
+export function getMarkdownUrl(route: string): string {
```
osama-rizk, line 11: **`getMarkdownUrl` doesn't strip the query string.** `usePathWithoutHash` splits on `#` only, so `/react/build-a-backend/auth/?foo=bar` → `/ai/pages/build-a-backend/auth/?foo=bar.md` (404). This PR now routes this function into three consumers (`<link rel="alternate">`, the WebMcp fetch, and the copy/open menu), so one bad URL propagates everywhere.

tasks/generate-agent-skills.mjs
```diff
@@ -0,0 +1,74 @@
...
+ * @returns {Array<object>} Skill entries for the discovery index
+ */
+function getSkills(domain) {
+  return [
+    {
+      name: 'amplify-workflow',
+      type: 'claude-skill',
+      description:
+        'Build and deploy full-stack web and mobile apps with AWS Amplify Gen2 (TypeScript code-first). Covers auth (Cognito), data (AppSync/DynamoDB), storage (S3), functions, APIs, and AI (Amplify AI Kit with Bedrock) across React, Next.js, Vue, Angular, React Native, Flutter, Swift, and Android.',
+      url: `${domain}/react/develop-with-ai/agent-plugins/`
```
osama-rizk: **Skill URL hardcodes `/react/` for a platform-agnostic page.** The discovery index is global; a docs restructure off `/react/` silently publishes a 404 to agents with no error anywhere. Use a platform-neutral/canonical path.

src/components/Layout/Layout.tsx
```diff
@@ -170,6 +171,14 @@ export const Layout = ({
     children?.props?.childPageNodes?.length != 'undefined' &&
```
osama-rizk: **`isOverview` guard is inert.** `children?.props?.childPageNodes?.length != 'undefined'` compares a number to the string `"undefined"` — always true (meant `typeof … !== 'undefined'`). It works today only because the `> 0` clause carries the whole predicate. Pre-existing, but this PR now depends on `isOverview` to gate `markdownUrl`, so it re-exposes it.

src/components/WebMcp/WebMcp.tsx
```diff
@@ -0,0 +1,120 @@
...
+    if (fromNav?.registerTool) return fromNav;
+  }
+  return null;
+}
+
+/**
+ * Fetch a markdown document and return its text, guarding against the SPA
+ * fallback returning an HTML page (e.g. a 404) instead of markdown.
+ */
+async function fetchMarkdown(url: string): Promise<string> {
```
osama-rizk: **`fetchMarkdown` duplicates `MarkdownMenu.handleCopy`.** Both fetch a `/ai/pages/*.md` URL and reject the SPA HTML fallback with the identical regex pair (`/^\s*<!doctype/i`, `/^\s*<html/i`). Fix the fallback detection in one and the other rots. Extract a shared `fetchPageMarkdown` next to `getMarkdownUrl` — you already made that move for `getMarkdownUrl`.

tasks/generate-wellknown.mjs
```diff
@@ -0,0 +1,148 @@
+import { promises as fs } from 'fs';
+import path from 'path';
+import dotenv from 'dotenv';
+
+dotenv.config({ path: './.env.custom' });
+
+const DOMAIN = process.env.SITEMAP_DOMAIN
```
osama-rizk: **`DOMAIN` + `ROOT_PATH` are copy-pasted across three task files** (`generate-sitemap`, `generate-wellknown`, `generate-agent-skills`). Change the output dir or default domain and you edit three files in lockstep. Extract a shared `tasks/build-constants.mjs`.

src/components/WebMcp/WebMcp.tsx
```diff
@@ -0,0 +1,120 @@
...
+    const modelContext = getModelContext();
+    if (!modelContext) return;
+
+    const controller = new AbortController();
+    const { signal } = controller;
+    const origin = window.location.origin;
+
+    const register = async () => {
+      try {
+        await modelContext.registerTool(
```
osama-rizk: **Both `registerTool` calls share one `try`.** If the first rejects (e.g. a transient duplicate-name error during the abort/re-register on fast client-side nav — both names are route-independent), the second tool never registers and the `catch` swallows it, leaving the page with one or zero tools. Independent `try`s per tool isolate them.

customHttp.yml
```diff
@@ -15,3 +15,25 @@ customHeaders:
       - key: 'Content-Security-Policy'
         value: 'upgrade-insecure-requests;'
       # CSP also set in _document.tsx meta tag
+      # Link headers advertise agent-discovery resources (RFC 8288 / RFC 9727):
+      # the API catalog, the agent skills index, the MCP server card, and the
+      # LLM-friendly documentation index.
+      - key: 'Link'
```
osama-rizk, line 30: **The `Link` header sits on the global `**/*` block,** so it rides every response (HTML, images, JSON), not just discovery routes — bytes on every request and a wider blast radius for the missing-file cases above. Worth a conscious choice vs. scoping it to the relevant paths.

osama-rizk (commented):
Nice, well-scoped change — solid tests and an honest scope section. No crash-level bugs; inline comments below, most-impactful first. (Checked and NOT flagging: `redirects.json` — the AJV validator accepts `status: "200"` as a string and the rule sits ahead of the `/<*>` catch-all, so it resolves 200 today; only nit is the test asserts existence, not ordering.)

osama-rizk (approved):

bobbor (approved):
LGTM.

only smaller nits that are not blocking. we can go forward with this

PR #8611, opened 2026-07-21
chore: amplify-js api references update

aws-amplify-ops:
Merge the api references changes from the most recent release.


soberm:
There is a newer version of this.

PR #8613, opened 2026-07-29
docs: add Amazon Connect Customer Profiles push notifications

soberm:
## Description of changes

Adds documentation for **Amazon Connect Customer Profiles push notifications** in Amplify Gen 2, covering both the backend (`defineNotifications`) and the client APIs (`aws-amplify/push-notifications/customer-profiles`).

This is a new feature area, so all pages are new; no existing pages are modified apart from registering the new nodes in `src/directory/directory.mjs` and scoping the Pinpoint end-of-life banner so it does not appear on the new Customer Profiles client pages.

### Backend — `build-a-backend/add-aws-services/notifications/`

| Page | Contents |
| --- | --- |
| `index.mdx` | Overview. What deploying `defineNotifications` provisions: the `AmplifyProfile` Customer Profiles object type keyed on the server-derived `principalId`, a DynamoDB device store, three IAM/SigV4-authorized HTTP API routes (`identify-user`, `register-device`, `remove-device`), a push-delivery Lambda for an Amazon Connect journey custom action, and an AWS End User Messaging application. Documents the `notifications.amazon_connect { endpoint, aws_region }` shape in `amplify_outputs.json`, and points readers arriving from Pinpoint push at the migration guide. |
| `set-up-notifications/index.mdx` | Prerequisites, then a `<BlockSwitcher>` covering the two modes: **create a new Amazon Connect instance** (omit `domainName`; optional `instanceAlias`, `expirationDays`) and **use an existing Customer Profiles domain** (`domainName` required). Includes the full `defineNotifications` / `defineBackend` sample with `secret()` for `apns.tokenKey` and `fcm.serviceJson`, per-channel property tables, and the resulting `amplify_outputs.json`. |
| `author-message-templates/index.mdx` | The required post-deploy step: authoring a `PUSH` message template in Amazon Q in Connect whose **name exactly matches the journey custom-action block name**, how the push-delivery Lambda resolves and renders it, and `{{Attributes.*}}` personalization. Push-channel templates are authored with the `qconnect` AWS CLI, so the page carries the exact `list-integration-associations` and `create-message-template` invocations. |
| `existing-resources/index.mdx` | Attach-mode deep dive: same account/Region as the Amplify project, and why the domain name must begin with `amazon-connect-` (the AWS-managed `AmazonConnectServiceLinkedRolePolicy` grants Customer Profiles access only on `arn:aws:profile:*:*:domains/amazon-connect-*`). |

### Client — `frontend/push-notifications/customer-profiles/`

| Page | Contents |
| --- | --- |
| `index.mdx` | Overview, install, `Amplify.configure`, and `initializePushNotifications()`. |
| `identify-user/index.mdx` | `identifyUser({ userProfile })` field table (`email` / `name` / `phone` / `location.{city,country,postalCode,region}` / `customAttributes`), the 255-character bounds, the reserved `principalId` key, and that `principalId` is server-derived and never sent by the client. |
| `register-device/index.mdx` | `registerDevice({ token })`, that registration happens automatically once a push token is issued, permission handling, and that the device belongs to the current identity. |
| `remove-device/index.mdx` | `removeDevice()` plus a prominent `<Callout warning>` for the pre-`signOut` contract. |
| `request-permissions/index.mdx` | Requesting notification permissions and reading permission status before registering a device. |
| `manage-device-token/index.mdx` | Observing token issuance and rotation with `onTokenReceived`. |
| `interact-with-notifications/index.mdx` | Handling received and opened notifications across foreground, background, and terminated app states. |
| `app-badge-count/index.mdx` | Reading and setting the application badge count. |
| `guest-and-authenticated-users/index.mdx` | Guest and authenticated callers are both supported via Identity Pool SigV4, and the device re-homes to the authenticated principal on sign-in. |

## Scope decisions

- **Backend pages** set the full `build-a-backend` platform set (`javascript`, `react-native`, `swift`, `android`, `flutter`, `angular`, `nextjs`, `react`, `vue`). Provisioning the resource is identical for every client: `defineNotifications`, the `apns`/`fcm` channel config, the `qconnect` template authoring, and the `domainName` attach mode are all backend concerns, so the pages resolve on every platform route.
- **Client pages** set `meta.platforms = ['react-native']`. The device flow that the guide is built around (`initializePushNotifications`, `registerDevice`, `removeDevice`) is React Native only and throws `PlatformNotSupportedError` on web. `identifyUser` does have a web implementation, and it is documented inside this React Native scoped set so the guide stays cohesive and its links into the device pages resolve.
- **No "Migrate from Pinpoint" page** is added here and the existing Pinpoint push notification pages are unchanged. The backend overview links to the separate migration guide so readers coming from Pinpoint push have a path forward.
- Attach vs. create-from-scratch is a single `set-up-notifications` page with a `<BlockSwitcher>` rather than two pages.
- `secret()` and guest access link out to the canonical pages rather than being re-explained.
- The Pinpoint end-of-life banner is suppressed on the Customer Profiles client pages, since this feature is not Pinpoint backed. The carve-out is anchored on `/push-notifications/customer-profiles/` so sibling routes keep the banner.

## Verification

- A full `next build` completes successfully (exit 0) on Node 20.
- All **13** new `.mdx` files (4 backend + 9 client) compile, and every page exports `meta`, `getStaticPaths`, and `getStaticProps` following the existing pattern; both overview pages set `meta.route` and use `getChildPageNodes`.
- The **13** new nodes are registered in `src/directory/directory.mjs`, which imports cleanly as ESM. All **921** node paths in the directory resolve to files that exist on disk (0 missing).
- All **45** routes generated by these pages prerender to HTML: 36 backend (4 pages across 9 platforms) and 9 client (9 pages on `react-native`).
- All **28** internal `/[platform]/...` links in the new pages resolve against the built output on every platform the containing page ships to (0 broken).
- Every JSX component used (`Callout`, `BlockSwitcher`, `Block`, `Overview`) is already registered in `mdx-components.tsx`; no new components were introduced.
- The `qconnect` invocations on the message-template page were checked against the AWS CLI service model: `Q_MESSAGE_TEMPLATES` is a valid `list-integration-associations` integration type, and the `create-message-template` call with `--channel-subtype PUSH` and the `{"push":{"apns":…,"fcm":…}}` content shape passes client-side parameter validation.
- `cspell` passes across all 2972 MDX files. `prettier --check` passes on `src/directory/directory.mjs`, the Layout component tests, and `cspell.json`. `.prettierignore` excludes `*.mdx`, so the new pages are correctly out of prettier's scope.
- The Layout component test suite passes (11 tests), including coverage for the Customer Profiles banner carve-out, its child routes, and a sibling route that keeps the banner.
- Prose checked against `STYLEGUIDE.md`: no "just" / "easy" / "simple", no "Amplify app", and no first-person "we" / "our".


cspell.json
```diff
@@ -463,6 +463,7 @@
     "confirmSignUp",
     "ConfirmSignUp",
     "confirmSignUpConfig",
+    "connectcampaignsv",
```
thisisabhash: cspell.json adds a word the PR never uses.

connectcampaignsv is added, but the diff's only occurrence of connectcampaign is that cspell entry itself — no page uses it (grep -c "connectcampaign" = 1, the definition).

src/components/Layout/__tests__/Layout.test.tsx
```diff
@@ -108,4 +118,26 @@ describe('Layout', () => {
       'layout-sidebar__mobile-toggle--open'
     );
   });
+
+  describe('PinpointEOLBanner', () => {
```
thisisabhash, line 122: The new test asserts the banner is hidden at /react/frontend/push-notifications/customer-profiles/, which passes. But the guard in [Layout.tsx:399-402](vscode-webview://1c1u45av80cib68odqklgma3pfhg7e4098to69pqq0qj0r2flspp/src/components/Layout/Layout.tsx#L399-L402) matches on .includes('/push-notifications/customer-profiles') without a trailing slash, so the eight child routes (.../customer-profiles/identify-user/ etc.) are the ones users will actually land on and none is covered.

Worth adding one child-route case, e.g. /react/frontend/push-notifications/customer-profiles/register-device/. The current implementation handles it correctly — this is just locking in the behaviour so a future refactor to an exact-path check doesn't silently reintroduce a Pinpoint EOL banner on Customer Profiles pages.

src/pages/[platform]/build-a-backend/add-aws-services/notifications/index.mdx
```diff
@@ -0,0 +1,77 @@
+import { getChildPageNodes } from '@/utils/getChildPageNodes';
+import { getCustomStaticPath } from '@/utils/getCustomStaticPath';
+
+export const meta = {
+  title: 'Push Notifications',
+  description:
+    'Send push notifications to your users with a backend powered by Amazon Connect Customer Profiles.',
+  route: '/[platform]/build-a-backend/add-aws-services/notifications',
+  platforms: ['react-native']
```
Comment (thisisabhash) [inline src/pages/[platform]/build-a-backend/add-aws-services/notifications/index.mdx]: All four build-a-backend/add-aws-services/notifications/ pages declare platforms: ['react-native'], but nothing on them is React Native-specific: defineNotifications, the apns/fcm channel config, the qconnect CLI template authoring, and the domainName attach mode are all backend-only and identical for every client. A Swift or Android developer landing on /swift/build-a-backend/add-aws-services/notifications/ gets a 404, and there's no other page telling them how to provision the resource.

src/pages/[platform]/build-a-backend/add-aws-services/notifications/index.mdx
```diff
@@ -0,0 +1,77 @@
+import { getChildPageNodes } from '@/utils/getChildPageNodes';
```
Comment (thisisabhash) [inline src/pages/[platform]/build-a-backend/add-aws-services/notifications/index.mdx:1]: The PR body notes migration content is handled separately; that's now merged as [add-aws-services/pinpoint-migration/index.mdx](vscode-webview://1c1u45av80cib68odqklgma3pfhg7e4098to69pqq0qj0r2flspp/src/pages/%5Bplatform%5D/build-a-backend/add-aws-services/pinpoint-migration/index.mdx) (#8612, on main). Since a large share of readers here are arriving from Pinpoint push, a link from this overview to the migration guide would close the loop.

cspell.json
soberm: Good catch, removed `connectcampaignsv` since it isn't referenced anywhere else in the diff. Kept `qconnect` and `unconfigured` as both are used by the new pages.

src/components/Layout/__tests__/Layout.test.tsx
soberm: Added a child-route test case (register-device). Confirmed the guard is genuine: tightening it to an exact-path match makes the new test fail, so this locks in the intended behavior.

src/pages/[platform]/build-a-backend/add-aws-services/notifications/index.mdx
Comment (soberm) [reply src/pages/[platform]/build-a-backend/add-aws-services/notifications/index.mdx]: Agreed. The backend setup is identical across platforms, so I expanded all four pages to the full build-a-backend platform set, which fixes the Swift/Android 404. I also removed the links to the React Native only client pages so they don't become dead links on the other platforms; those client docs are still reachable via the sidebar and link back to these pages.

src/pages/[platform]/build-a-backend/add-aws-services/notifications/index.mdx
Comment (soberm) [reply src/pages/[platform]/build-a-backend/add-aws-services/notifications/index.mdx:1]: Added a pointer to the Pinpoint migration guide from the overview. It's merged on main now, so the link resolves.

thisisabhash (dismissed):

osama-rizk (commented):
Thanks for the thorough write-up — the code changes here (`Layout.tsx`, tests, `directory.mjs`, `cspell`) are clean, minimal, and correct. Registration ↔ file parity checks out: all 13 new `.mdx` files (4 backend + 9 client) are both registered in `directory.mjs` and present, no orphans. A few things to reconcile before merge, one of which may be a real discoverability bug rather than a doc typo.

## 1. Platform scope: the PR description contradicts the diff (please confirm intent)

The "Scope decisions" section says `meta.platforms = ['javascript', 'react-native']` on every new page, that `identifyUser` is supported on both web and RN, and that the client overview has a platform-support table + `<InlineFilter>`. None of that matches the diff:

- **Backend pages** (`index`, `set-up-notifications`, `author-message-templates`, `existing-resources`) list **all 9 platforms**, not `['javascript','react-native']`.
- **Client pages** (all 9, including `index` and `identify-user`) are **`['react-native']` only** — `javascript` is excluded.
- I don't see the platform-support table or `<InlineFilter>` in the client `index.mdx`.

Why this matters: if `identifyUser` really is web-supported, shipping `identify-user/index.mdx` as `['react-native']`-only **hides it from every web/JS reader** — a genuine coverage gap. If the feature is RN-only, the description's web claims are just stale. Could you confirm which is correct, and either fix the description or add `'javascript'` to the client pages that support web?

## 2. Verification claims appear stale ("9 files" / "916 nodes")

The Verification section references "9 new `.mdx` files" and "916 node paths … including the 9 new nodes," but the diff adds **13** new files / **13** new nodes. So the "0 broken links / 0 missing files" evidence predates the current diff and likely didn't exercise `request-permissions`, `manage-device-token`, `interact-with-notifications`, or `app-badge-count`. Worth re-running the compile/link check across all 13 (the registration/file parity itself is fine — I checked manually).

## 3. `Layout.tsx` carve-out — correct, one boundary note

The banner-suppression logic is right: precedence is correct, it only removes the Pinpoint EOL banner from the new Customer Profiles client pages, and the backend `notifications/` pages were never in scope (no `/push-notifications/` substring). 

Minor: `.includes('/push-notifications/customer-profiles')` is an unbounded substring match. Nothing collides today, but a future sibling like `.../customer-profiles-legacy` would also be silently carved out. Not a blocker — just flagging.

## 4. Test coverage — good refactor, one gap

Moving `routerMock` to a mutable `routerState` + `beforeEach` reset is the right call, and existing tests keep their prior values. The new tests cover the client carve-out (index + child) and two positive cases (gen1 legacy, analytics).

Missing: a test asserting the banner **still renders for a Gen 2 Pinpoint push page**. The carve-out's actual risk is over-matching *within* gen2 `/push-notifications/`; the current positive tests only prove gen1 and analytics. A gen2-Pinpoint assertion would pin that boundary. Nice-to-have.

## 5. `author-message-templates` CLI — please have a SME run it once

This is the highest-risk content page because it's copy-pasteable CLI: `aws connect list-integration-associations --integration-type Q_MESSAGE_TEMPLATES`, the `knowledge-base/<KB_ID>` ARN-suffix parse, and the `qconnect create-message-template --channel-subtype PUSH --content '{"push":{"apns":…,"fcm":…}}'` schema. I can't verify the exact flag names or content JSON shape from the diff. Linking the `CreateMessageTemplate` API reference is a good hedge, but a wrong key here fails silently for the reader — worth one real run-through.

## Nits (non-blocking)
- `interact-with-notifications/index.mdx`: the two markdown tables have ragged dash/column widths. Renders fine and `.prettierignore` excludes `*.mdx`, so nothing enforces it — source readability only.
- `cspell` additions (`qconnect`, `unconfigured`) are both used and correctly alphabetized. 👍
- Secret names are consistent between the `secret set` prerequisite and the `secret('…')` config sample. 👍

**Bottom line:** the only thing I'd hold on is #1 — reconcile the platform scope, since RN-only client pages would be a real discoverability bug if web is supported. #2 and #5 are "re-verify before trusting"; the rest are non-blocking.

soberm:
Thanks for the detailed review. Responses to each item.

**1. Platform scope.** You are right that the description did not match the diff, and I have corrected the description. On the substance, I am keeping the client pages scoped to `react-native`. `identifyUser` does have a real web implementation, so your reading of the source is correct, but the device flow the client guide is built around (`initializePushNotifications`, `registerDevice`, `removeDevice`) is React Native only and throws `PlatformNotSupportedError` on web. `identify-user` also links into `register-device`, `remove-device`, and `guest-and-authenticated-users`, so adding `javascript` to that one page would surface it to web readers with three links that resolve only on `react-native`. Keeping the client set cohesive at `react-native` avoids that. The backend pages are the ones that needed widening, and they now carry the full nine platform `build-a-backend` set, which is what fixes the Swift and Android 404.

**2. Stale verification numbers.** Correct, those numbers predated the current diff. The description now reflects a re-run against a clean production build: 13 new `.mdx` files (4 backend and 9 client), 13 new `directory.mjs` nodes, 921 total node paths all resolving on disk with 0 missing, 45 generated routes all prerendering (36 backend, being 4 pages across 9 platforms, plus 9 client pages on `react-native`), and 28 internal `/[platform]/...` links with 0 broken. That pass does exercise `request-permissions`, `manage-device-token`, `interact-with-notifications`, and `app-badge-count`.

**3. Unbounded substring match.** Good catch, fixed. The carve-out is now anchored on `/push-notifications/customer-profiles/` with a trailing slash, so a sibling such as `customer-profiles-legacy` keeps the banner. `next.config` sets `trailingSlash: true`, so the overview route and all child routes still match.

**4. Positive test for the banner boundary.** Worth noting there is no Gen 2 Pinpoint push page in the repo today; the only Gen 2 `/push-notifications/` pages are the Customer Profiles ones, and all 29 Pinpoint push pages are Gen 1. So I added the test that captures the intent, a positive case on a Gen 2 `/push-notifications/` sibling route that is not Customer Profiles. It pins the over-matching boundary the existing Gen 1 and analytics cases did not cover, and I confirmed it is meaningful: it fails against the old unbounded match and passes with the anchored one. The Layout suite is at 11 tests.

**5. CLI accuracy on `author-message-templates`.** I validated the three specific parameters. `Q_MESSAGE_TEMPLATES` is confirmed in the construct source as the integration type used for this lookup, and the knowledge base ID being the `knowledge-base/<KB_ID>` suffix of `IntegrationArn` matches the ARN the push Lambda parses. For the template call, I ran the page's exact `qconnect create-message-template` invocation through the AWS CLI: `--channel-subtype PUSH` and the `{"push":{"apns":...,"fcm":...}}` content shape pass client-side parameter validation and the request reaches the service, failing only on credentials. The CLI model also confirms `content.push` accepts exactly `adm`, `apns`, `fcm`, and `baidu`, so the `apns` and `fcm` keys are right. I agree a full end to end run is still worth doing, since validation does not prove the rendered output; that needs a live Connect instance and credentials, so I am treating it as a SME follow-up rather than something I can close here.

The table alignment nit in `interact-with-notifications` is also fixed, padding only with no cell text changes.

osama-rizk (approved):

thisisabhash (approved):

PR #8616, opened 2026-07-31
docs: expand Customer Profiles push notifications to Android, Flutter, and Swift

ekjotmultani:
## Description

The standalone Amazon Connect Customer Profiles clients are now released on all three mobile platforms:

- **Android**: `aws-connect` (`AmplifyConnectClient`) in amplify-android
- **Flutter**: `amplify_connect_client` in amplify-flutter
- **Swift**: `AmplifyConnectClient` in amplify-swift

This expands the five Customer Profiles client pages under **frontend > Push Notifications** from `platforms: ['react-native']` to include `android`, `flutter`, and `swift`, following up on the platform scoping deliberately deferred in #8613.

### Pages expanded

- **Overview** — per-platform install and client creation (Gradle/pubspec/SPM), including the Swift credentials bridge and the Flutter Amplify Auth default
- **Identify a user** — per-platform `UserProfile` snippets and error handling
- **Register a device** — FCM/APNs token acquisition per platform and the per-platform resolved device fields
- **Remove a device** — remove-before-sign-out guidance and per-platform never-registered behavior
- **Guest and authenticated users** — documents that the mobile clients require an explicit `registerDevice` after sign-in (React Native handles this automatically)

### Deliberately unchanged

- The four React Native push-handling pages (request permissions, manage device token, interact with notifications, app badge count) stay React Native only — the standalone clients register tokens with the backend but do not handle notification display.
- The Analytics client pages (Kinesis, Firehose) already cover all three platforms.

All snippets were written against the released public API surfaces in each repo.

## Testing

- `yarn test:unit`: 305 tests, 62 suites, all passing
- `yarn spellcheck`: clean (3,023 files)
- Full production build: all four platforms prerender the customer-profiles routes; RN-only child pages correctly absent from the new platforms; all internal link targets verified to build on android/flutter/swift

### DIFF
--- a/src/pages/[platform]/frontend/push-notifications/customer-profiles/guest-and-authenticated-users/index.mdx
+++ b/src/pages/[platform]/frontend/push-notifications/customer-profiles/guest-and-authenticated-users/index.mdx
@@ -145,8 +145,15 @@ await Amplify.Auth.signOut();
 <InlineFilter filters={['swift']}>
 
 ```swift
-try await client.removeDevice()
-_ = await Amplify.Auth.signOut()
+func handleSignOut() async {
+    do {
+        try await client.removeDevice()
+    } catch {
+        print("Failed to remove device before sign out: \(error)")
+    }
+
+    _ = await Amplify.Auth.signOut()
+}
 ```
 
 </InlineFilter>
--- a/src/pages/[platform]/frontend/push-notifications/customer-profiles/guest-and-authenticated-users/index.mdx
+++ b/src/pages/[platform]/frontend/push-notifications/customer-profiles/guest-and-authenticated-users/index.mdx
@@ -4,7 +4,7 @@ export const meta = {
   title: 'Guest and authenticated users',
   description:
     'How Amazon Connect Customer Profiles notifications work for guest and signed-in users, and how a device moves to the authenticated identity on sign-in.',
-  platforms: ['react-native']
+  platforms: ['android', 'flutter', 'react-native', 'swift']
 };
 
 export const getStaticPaths = async () => {
@@ -43,6 +43,8 @@ A guest identity is scoped to the installation, not to a person. Treat a guest p
 
 A guest user's push token and a signed-in user's push token are the same value, because the native platform issues the token to the installation rather than to a user. Without any handling, a device registered as a guest would stay attached to the guest identity after the user signed in, and journeys targeting the authenticated profile would not reach it.
 
+<InlineFilter filters={['react-native']}>
+
 The library handles this for you. After `initializePushNotifications` has been called, it listens for the sign-in event and re-registers the device, which moves the existing registration from the guest identity to the authenticated one. Because registration is an idempotent operation keyed on a device identifier that is stable for the installation, this updates the existing record rather than creating a second one.
 
 No application code is required for this to happen:
@@ -60,9 +62,57 @@ Re-registration on sign-in is a best-effort operation. A failure is logged and d
 
 If no push token has been issued yet when a user signs in, there is nothing to move. The first registration then happens against the authenticated identity once the native platform issues the token.
 
+</InlineFilter>
+
+<InlineFilter filters={['android', 'flutter', 'swift']}>
+
+Call `registerDevice` again after a user signs in to move the registration. The call is signed with the authenticated credentials, and because registration is an idempotent operation keyed on a device identifier that is stable for the installation, it moves the existing record to the authenticated identity rather than creating a second one.
+
+</InlineFilter>
+
+<InlineFilter filters={['android']}>
+
+```kotlin
+// After sign-in completes:
+val token = FirebaseMessaging.getInstance().token.await()
+client.registerDevice(token)
+```
+
+</InlineFilter>
+
+<InlineFilter filters={['flutter']}>
+
+```dart
+// After sign-in completes:
+final token = await FirebaseMessaging.instance.getToken();
+if (token != null) {
+  await client.registerDevice(token: token);
+}
+```
+
+</InlineFilter>
+
+<InlineFilter filters={['swift']}>
+
+```swift
+// After sign-in completes, with the APNs token you received
+// in didRegisterForRemoteNotificationsWithDeviceToken:
+try await client.registerDevice(token: token)
+```
+
+</InlineFilter>
+
+<InlineFilter filters={['android', 'flutter', 'swift']}>
+
+If no push token has been issued yet when a user signs in, there is nothing to move. The first registration then happens against the authenticated identity once the platform issues the token.
+
+</InlineFilter>
+
 ## Sign a user out
 
 Signing a user out does not de-register their device. Call `removeDevice` while the user is still signed in, because de-registration is signed with their credentials:
+
+<InlineFilter filters={['react-native']}>
 
 ```ts
 import { signOut } from 'aws-amplify/auth';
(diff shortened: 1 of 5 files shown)


src/pages/[platform]/frontend/push-notifications/customer-profiles/index.mdx
```diff
@@ -72,9 +137,156 @@ AppRegistry.registerComponent(appName, () => App);
...
+import 'package:amplify_flutter/amplify_flutter.dart' hide UserProfile;
+import 'package:flutter/services.dart' show rootBundle;
+
+final raw = await rootBundle.loadString('amplify_outputs.json');
+final outputs = jsonDecode(raw) as Map<String, dynamic>;
+
+// Amplify.configure parses only its own sections, so remove the
+// notifications key before configuring. The Connect client reads it
+// from the decoded map instead.
+final authOutputs = Map<String, dynamic>.from(outputs)..remove('notifications');
```
Comment (soberm) [inline src/pages/[platform]/frontend/push-notifications/customer-profiles/index.mdx:206]: [suggestion] Is this stripping-the-notifications-key step actually required? The published `amplify_connect_client` README just calls `AmplifyConnectClientFlutter.createFromAmplifyOutputs(amplifyOutputs: amplifyOutputs)` on the full decoded outputs, no mention of removing the notifications key before `Amplify.configure()`. If `Amplify.configure` genuinely chokes on that key on the current release, keep this, but if it doesn't we should drop the workaround since it adds unnecessary complexity and could send readers down the wrong path.

src/pages/[platform]/frontend/push-notifications/customer-profiles/index.mdx
```diff
@@ -47,6 +71,47 @@ Because push notifications interact with the native platform, you also need to l
...
 
+</InlineFilter>
+
+<InlineFilter filters={['android']}>
+
+Add the dependency to your module's `build.gradle.kts`:
+
+```kotlin
+dependencies {
+    implementation("com.amplifyframework:aws-connect:LATEST_VERSION")
```
Comment (soberm) [inline src/pages/[platform]/frontend/push-notifications/customer-profiles/index.mdx:82]: [nitpick] Couldn't confirm `com.amplifyframework:aws-connect` on Maven Central when I checked (0 results), though that could just be an indexing lag since the Flutter package was released only a couple days ago too. Worth a quick sanity check of the exact group/artifact coordinate before merge so nobody copies a build.gradle line that 404s.

src/pages/[platform]/frontend/push-notifications/customer-profiles/identify-user/index.mdx
```diff
@@ -54,9 +137,9 @@ Every field is optional, so send only the values your application has. Each call
 | `name` | `string` | The user's name. |
 | `phone` | `string` | The user's phone number. |
 | `location` | `object` | The user's location. Accepts `city`, `country`, `postalCode`, and `region`, each a `string`. |
-| `customAttributes` | `Record<string, string>` | Additional key-value pairs to store on the profile. |
+| `customAttributes` | `map` | Additional string key-value pairs to store on the profile. |
```
Comment (soberm) [inline src/pages/[platform]/frontend/push-notifications/customer-profiles/identify-user/index.mdx:140]: [nitpick] The old copy explicitly said "customAttributes values must be strings", this drops that constraint when the type became `map` instead of `Record<string, string>`. If all four SDKs still require string values here, worth restoring that note so it doesn't read like arbitrary values are now allowed.

src/pages/[platform]/frontend/push-notifications/customer-profiles/register-device/index.mdx
```diff
@@ -72,14 +74,125 @@ const listener = onTokenReceived(async (token) => {
...
+FirebaseMessaging.instance.onTokenRefresh.listen((token) async {
+  await client.registerDevice(token: token);
+});
+```
+
+Requesting notification permission and receiving messages remain your application's responsibility through your push provider; the client only registers the token with your backend.
+
+<Callout info>
+
+Device registration requires a push channel, so `registerDevice` throws `ConnectUnsupportedOperationException` on web and desktop platforms. `identifyUser` works on every platform.
```
Comment (soberm) [inline src/pages/[platform]/frontend/push-notifications/customer-profiles/register-device/index.mdx:121]: [nitpick] Flagging for awareness rather than asking for a change: this callout implies the `flutter` platform tag here covers non-mobile targets too (web/desktop), which matches the package's platform support matrix on pub.dev, but nothing else on this page discusses non-mobile Flutter usage. Fine to leave as is if that's intentional.

src/pages/[platform]/frontend/push-notifications/customer-profiles/guest-and-authenticated-users/index.mdx
```diff
@@ -72,23 +122,39 @@ await removeDevice();
...
+
+```swift
+try await client.removeDevice()
+_ = await Amplify.Auth.signOut()
 ```
 
-See [Identify a user](/[platform]/frontend/push-notifications/customer-profiles/identify-user/).
+</InlineFilter>
+
+After sign-out completes, subsequent calls are signed as a new guest identity. See [Remove a device](/[platform]/frontend/push-notifications/customer-profiles/remove-device/).
```
Comment (soberm) [inline src/pages/[platform]/frontend/push-notifications/customer-profiles/guest-and-authenticated-users/index.mdx:161]: Good restructuring here, moving the ordering callout to wrap all four platform variants instead of duplicating it per platform, and generalizing "after signOut completes" to "after sign-out completes" so the prose doesn't leak a specific API name into shared copy.

src/pages/[platform]/frontend/push-notifications/customer-profiles/register-device/index.mdx
```diff
@@ -72,14 +74,125 @@ const listener = onTokenReceived(async (token) => {
...
+</InlineFilter>
+
+<InlineFilter filters={['android']}>
+
+| Field | Source |
+| --- | --- |
+| `deviceId` | A stable identifier the library generates and persists per installation, shared with other Amplify libraries. |
+| `platform` | The `platform` string passed when creating the client. |
+| `appVersion` | The `appVersion` string passed when creating the client. |
+| `channelType` | The `channelType` passed when creating the client. Defaults to `ChannelType.GCM`. |
```
Comment (soberm) [inline src/pages/[platform]/frontend/push-notifications/customer-profiles/register-device/index.mdx:170]: The `ChannelType.GCM` default here lines up with what the overview page's client-creation snippet documents, and the Flutter/Swift derivation logic in the rows below is consistent with their respective overview sections too. Nice to see that stayed in sync across pages.

src/pages/[platform]/frontend/push-notifications/customer-profiles/guest-and-authenticated-users/index.mdx
```diff
@@ -72,23 +122,39 @@ await removeDevice();
...
+await client.removeDevice();
+await Amplify.Auth.signOut();
+```
+
+</InlineFilter>
+
+<InlineFilter filters={['swift']}>
+
+```swift
+try await client.removeDevice()
```
Comment (thisisabhash) [inline src/pages/[platform]/frontend/push-notifications/customer-profiles/guest-and-authenticated-users/index.mdx]: should be something like
```
func handleSignOut() async {
    do {
        try await client.removeDevice()
    } catch {
        print("Failed to remove device before sign out: \(error)")
    }

    _ = await Amplify.Auth.signOut()
}
```

src/pages/[platform]/frontend/push-notifications/customer-profiles/index.mdx
Comment (ekjotmultani) [reply src/pages/[platform]/frontend/push-notifications/customer-profiles/index.mdx:206]: It is, unfortunately. `NotificationsOutputs` in amplify_core still has the pinpoint fields as required, so `Amplify.configure` throws on an outputs file where notifications only has `amazon_connect`. The README doesn't hit it because its snippet never calls `Amplify.configure`, but the example app in the repo does the same strip for the same reason. We have a parser fix in flight on amplify-flutter to accept the new shape, once that releases I'll simplify this snippet to a plain configure.

src/pages/[platform]/frontend/push-notifications/customer-profiles/index.mdx
Comment (ekjotmultani) [reply src/pages/[platform]/frontend/push-notifications/customer-profiles/index.mdx:82]: It's there, the search index just hasn't caught up. Direct path resolves: https://repo1.maven.org/maven2/com/amplifyframework/aws-connect/2.39.0/ (went out with the 2.39.0 release on Jul 30).

src/pages/[platform]/frontend/push-notifications/customer-profiles/identify-user/index.mdx
Comment (ekjotmultani) [reply src/pages/[platform]/frontend/push-notifications/customer-profiles/identify-user/index.mdx:140]: Restored the note.

src/pages/[platform]/frontend/push-notifications/customer-profiles/register-device/index.mdx
Comment (ekjotmultani) [reply src/pages/[platform]/frontend/push-notifications/customer-profiles/register-device/index.mdx:121]: Yeah that's intentional. `identifyUser` works everywhere Flutter runs, only device registration needs a push channel, so the callout scopes exactly that.

src/pages/[platform]/frontend/push-notifications/customer-profiles/guest-and-authenticated-users/index.mdx
Comment (ekjotmultani) [reply src/pages/[platform]/frontend/push-notifications/customer-profiles/guest-and-authenticated-users/index.mdx]: Done, switched it to that shape. It also matches what the remove-device page shows, so the two pages agree now.

thisisabhash (approved):

src/pages/[platform]/frontend/push-notifications/customer-profiles/index.mdx
Comment (soberm) [reply src/pages/[platform]/frontend/push-notifications/customer-profiles/index.mdx:206]: Why not make the pinpoint fields optional?

soberm (approved):

PR #8620, opened 2026-08-31
feat(banner): add maintenance-mode banner to Amplify JavaScript v5 pages

soberm:
## Description of changes

Amplify JavaScript v5 is in maintenance mode, but the v5 documentation pages currently give readers no in-page signal of that, and no pointer to the upgrade path. Readers can land on a v5 page from search and follow deprecated guidance without realizing a newer major version exists.

This adds a maintenance-mode banner to all Amplify JavaScript v5 pages, linking to the v5 → v6 migration guide.

Banner copy:

> **JavaScript v5** — Amplify JavaScript v5 has entered maintenance mode. We recommend upgrading to v6. A [migration guide](https://docs.amplify.aws/gen1/javascript/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/) is available to help you upgrade from v5 to v6.

Screenshot 2026-08-31 at 13 29 53

### Implementation

- **`src/components/JsV5MaintenanceBanner/`** — new component. Follows `Gen1Banner`'s conventions: takes `currentPlatform` and builds a **root-relative, platform-aware** link, so a reader on `/gen1/vue/prev/...` goes to the Vue guide and stays within the current environment (no jump to production docs from a preview build).
- **`src/utils/isJsV5Page.ts`** — route detection. v5 pages live under `/gen1/<platform>/prev/...`, and which version `prev` resolves to is read from **`PLATFORM_VERSIONS`** (v5 for JS platforms, v1 for Android/Swift/Flutter) rather than inferred from the platform, so the banner can't outlive v5 when a platform's `prev` rolls forward.
- **`src/components/Layout/Layout.tsx`** — renders the banner alongside the existing `Gen1Banner` / `Gen2MaintenanceBanner` calls.
- **`src/styles/global-nav.scss`** — extracted shared `%page-banner` placeholders now used by all three banners (the styles were about to be duplicated a third time). The v5 banner is a deliberately lighter left-accented variant with an outlined badge, so it reads as a version-specific sub-notice beneath the Gen1 banner instead of a second identical amber banner. Net compiled CSS: 79,336 → 78,250 bytes.

### Verification

Unit tests (9, new) — `yarn test:unit`:

```
PASS src/utils/__tests__/isJsV5Page.test.ts
PASS src/components/JsV5MaintenanceBanner/__tests__/JsV5MaintenanceBanner.test.tsx
Test Suites: 2 passed, 2 total
Tests:       9 passed, 9 total
```

Against a running dev server (`next dev`), inspecting rendered HTML:

- Banner **present** on `/gen1/react/prev/...` and `/gen1/javascript/prev/...`, with a platform-correct root-relative href:
  ```html
  <div class="js-v5-banner"><span class="js-v5-banner__badge">JavaScript v5</span><span class="js-v5-banner__text">Amplify JavaScript v5 has entered maintenance mode. We recommend upgrading to v6. A <a class="js-v5-banner__link" href="/gen1/react/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/">migration guide</a> is available to help you upgrade from v5 to v6.</span></div>
  ```
- Banner **absent** (0 occurrences) on `/gen1/react/build-a-backend/...` (Gen 1, v6), `/gen1/swift/prev/...` (non-JS `prev` = v1), and `/react/build-a-backend/...` (Gen 2).
- Link target returns **HTTP 200 for all six JS platforms**.
- Refactor safety: compiled `styles.scss` before/after and compared effective declarations for every pre-existing banner selector (including dark-mode variants) — all identical; only declaration order changed.

`eslint` clean on changed files; `prettier` reports no changes wanted on any added line.

Not verified: no rendered screenshot — Playwright isn't available in my environment, so the visual claims come from compiled CSS + rendered HTML. Worth an eyeball on the preview build.

One open product question: unlike the Gen 1 banner, this copy names no end-of-life date, because I found no published EOL date for JS v5. If one is committed, that sentence should be added.

## Checklist

- [x] PR description included
- [x] `yarn test:unit` passes
- [x] No documentation content changes (presentation/component change only)

### DIFF
--- a/src/components/JsV5MaintenanceBanner/JsV5MaintenanceBanner.tsx
+++ b/src/components/JsV5MaintenanceBanner/JsV5MaintenanceBanner.tsx
@@ -1,16 +1,33 @@
 import Link from 'next/link';
+import { Platform } from '@/data/platforms';
 
-export const MIGRATION_GUIDE_URL =
-  'https://docs.amplify.aws/gen1/javascript/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/';
+/**
+ * Builds the route to the v5 -> v6 migration guide for a given platform.
+ *
+ * The guide is a `[platform]` route served by this app and its content is
+ * filtered per platform, so linking within the reader's current platform keeps
+ * their platform/nav context intact.
+ */
+export const getMigrationGuideUrl = (platform: Platform) =>
+  `/gen1/${platform}/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/`;
 
-export const JsV5MaintenanceBanner = () => {
+interface JsV5MaintenanceBannerProps {
+  currentPlatform: Platform;
+}
+
+export const JsV5MaintenanceBanner = ({
+  currentPlatform
+}: JsV5MaintenanceBannerProps) => {
   return (
     <div className="js-v5-banner">
-      <span className="js-v5-banner__badge">Maintenance Mode</span>
+      <span className="js-v5-banner__badge">JavaScript v5</span>
       <span className="js-v5-banner__text">
         Amplify JavaScript v5 has entered maintenance mode. We recommend
         upgrading to v6. A{' '}
-        <Link href={MIGRATION_GUIDE_URL} className="js-v5-banner__link">
+        <Link
+          href={getMigrationGuideUrl(currentPlatform)}
+          className="js-v5-banner__link"
+        >
           migration guide
         </Link>{' '}
         is available to help you upgrade from v5 to v6.
--- a/src/components/JsV5MaintenanceBanner/JsV5MaintenanceBanner.tsx
+++ b/src/components/JsV5MaintenanceBanner/JsV5MaintenanceBanner.tsx
@@ -1,0 +1,20 @@
+import Link from 'next/link';
+
+export const MIGRATION_GUIDE_URL =
+  'https://docs.amplify.aws/gen1/javascript/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/';
+
+export const JsV5MaintenanceBanner = () => {
+  return (
+    <div className="js-v5-banner">
+      <span className="js-v5-banner__badge">Maintenance Mode</span>
+      <span className="js-v5-banner__text">
+        Amplify JavaScript v5 has entered maintenance mode. We recommend
+        upgrading to v6. A{' '}
+        <Link href={MIGRATION_GUIDE_URL} className="js-v5-banner__link">
+          migration guide
+        </Link>{' '}
+        is available to help you upgrade from v5 to v6.
+      </span>
+    </div>
+  );
+};
--- a/src/components/Layout/Layout.tsx
+++ b/src/components/Layout/Layout.tsx
@@ -37,6 +37,8 @@ import {
 } from '@/components/NextPrevious';
 import { Modal } from '@/components/Modal';
 import { Gen1Banner } from '@/components/Gen1Banner';
+import { JsV5MaintenanceBanner } from '@/components/JsV5MaintenanceBanner';
+import { isJsV5Page } from '@/utils/isJsV5Page';
 import { Gen2MaintenanceBanner } from '@/components/Gen2MaintenanceBanner';
 import { CrossLink } from '@/components/CrossLink';
 import { findDirectoryNode } from '@/utils/findDirectoryNode';
@@ -81,6 +83,7 @@ export const Layout = ({
   const metaUrl = url ? url : basePath + asPathWithNoHash;
   const pathname = router.pathname;
   const isGen1 = asPathWithNoHash.split('/')[1] === 'gen1';
+  const isJsV5 = isJsV5Page(asPathWithNoHash);
   const isContributor = asPathWithNoHash.split('/')[1] === 'contribute';
   const currentGlobalNavMenuItem = isContributor ? 'Contribute' : 'Docs';
   const isHome = pageType === 'home';
@@ -379,6 +382,7 @@ export const Layout = ({
                       />
                     ) : null}
                     {isGen1 && <Gen1Banner currentPlatform={currentPlatform} />}
+                    {isJsV5 && <JsV5MaintenanceBanner />}
                     {!isGen1 && <Gen2MaintenanceBanner />}
                     {crossLinkProps && <CrossLink {...crossLinkProps} />}
                     {useCustomTitle ? null : (
--- a/src/components/Layout/Layout.tsx
+++ b/src/components/Layout/Layout.tsx
@@ -382,7 +382,11 @@ export const Layout = ({
                       />
                     ) : null}
                     {isGen1 && <Gen1Banner currentPlatform={currentPlatform} />}
-                    {isJsV5 && <JsV5MaintenanceBanner />}
+                    {isJsV5 && (
+                      <JsV5MaintenanceBanner
+                        currentPlatform={currentPlatform}
+                      />
+                    )}
                     {!isGen1 && <Gen2MaintenanceBanner />}
                     {crossLinkProps && <CrossLink {...crossLinkProps} />}
                     {useCustomTitle ? null : (
--- a/src/styles/global-nav.scss
+++ b/src/styles/global-nav.scss
@@ -367,6 +367,50 @@
   }
 }
 
+// Amplify JavaScript v5 maintenance-mode banner
+.js-v5-banner {
+  display: flex;
+  align-items: center;
+  gap: var(--amplify-space-small);
+  padding: var(--amplify-space-small) var(--amplify-space-medium);
+  background-color: #fef3c7;
+  border: 1px solid #f59e0b;
+  border-radius: var(--amplify-radii-medium);
+  margin-bottom: var(--amplify-space-medium);
+
+  @include darkMode {
+    background-color: rgba(245, 158, 11, 0.1);
+    border-color: rgba(245, 158, 11, 0.3);
+  }
+}
+
+.js-v5-banner__badge {
+  display: inline-block;
+  background-color: #b45309;
+  color: var(--amplify-colors-white);
+  font-size: 11px;
+  font-weight: 700;
+  text-transform: uppercase;
+  letter-spacing: 0.5px;
+  padding: 2px 8px;
+  border-radius: var(--amplify-radii-small);
+  flex-shrink: 0;
+}
+
+.js-v5-banner__text {
+  font-size: var(--amplify-font-sizes-small);
+  color: var(--amplify-colors-font-primary);
+}
+
+.js-v5-banner__link {
+  color: var(--amplify-colors-primary-80);
+  font-weight: 600;
+  text-decoration: none;
+  &:hover {
+    text-decoration: underline;
+  }
+}
+
 // Gen2 maintenance-mode banner (dismissable)
 .gen2-maintenance-banner {
   display: flex;
--- a/src/styles/global-nav.scss
+++ b/src/styles/global-nav.scss
@@ -323,16 +323,21 @@
   margin-top: -2px;
 }
 
-// Gen1 page-level legacy banner
-.gen1-page-banner {
+// Shared building blocks for page-level notice banners. Extended by the Gen1,
+// Gen2 maintenance and JavaScript v5 banners below.
+%page-banner {
   display: flex;
   align-items: center;
   gap: var(--amplify-space-small);
   padding: var(--amplify-space-small) var(--amplify-space-medium);
-  background-color: #fef3c7;
-  border: 1px solid #f59e0b;
   border-radius: var(--amplify-radii-medium);
   margin-bottom: var(--amplify-space-medium);
+}
+
+// Filled amber treatment used by the top-level maintenance banners.
+%page-banner--amber {
+  background-color: #fef3c7;
+  border: 1px solid #f59e0b;
 
   @include darkMode {
     background-color: rgba(245, 158, 11, 0.1);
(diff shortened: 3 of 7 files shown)


cadivus (dismissed):

src/utils/isJsV5Page.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+export function isJsV5Page(path: string): boolean {
+  if (!path) return false;
+
+  const [, gen, platform, version] = path.split('/');
+
+  return (
+    gen === 'gen1' &&
+    version === 'prev' &&
+    JS_PLATFORMS.includes(platform as JSPlatform)
```
osama-rizk: This infers "v5" from a proxy — `gen1 && prev && JS_PLATFORMS.includes(platform)` — rather than from the source of truth. `PLATFORM_VERSIONS` is what actually maps each platform's `prev` to a version. Today `(JS platform + prev) ⟺ v5` holds, but it's a derived correlation, not enforced.

When the next version cycle lands (v6 → maintenance, v7 current) and a JS platform's `prev` becomes `v6`, `/gen1/nextjs/prev/` would be a **v6** page but this still returns true → the "v5 maintenance" banner shows on a v6 page. Reading the truth directly is the same cost and can't drift:

```ts
import { PLATFORM_VERSIONS } from '@/data/platforms';
// ...
return gen === 'gen1' && PLATFORM_VERSIONS[platform as Platform]?.prev === 'v5';
```

src/components/JsV5MaintenanceBanner/JsV5MaintenanceBanner.tsx
```diff
@@ -0,0 +1,20 @@
+import Link from 'next/link';
+
+export const MIGRATION_GUIDE_URL =
```
osama-rizk: The migration guide is served by this same app (`src/pages/gen1/[platform]/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/index.mdx`), but this hardcodes the absolute production URL. Two consequences: on preview/staging/localhost the link jumps to **production** docs (leaving the environment being previewed), and it forces a full page reload instead of client-side nav.

A root-relative path fixes both and matches `Gen1Banner`, which the PR description says this mirrors:

```ts
export const MIGRATION_GUIDE_URL =
  '/gen1/javascript/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/';
```

(`Gen2MaintenanceBanner` does use an absolute URL, so there's precedent — not a blocker.)

src/components/JsV5MaintenanceBanner/JsV5MaintenanceBanner.tsx
```diff
@@ -0,0 +1,20 @@
...
+  'https://docs.amplify.aws/gen1/javascript/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/';
+
+export const JsV5MaintenanceBanner = () => {
```
osama-rizk: The link target hardcodes the `javascript` platform, so a reader on `/gen1/react/prev/...` (or vue/angular/nextjs) is sent to the **javascript** guide, switching their platform/nav context. `Gen1Banner` takes a `currentPlatform` prop and links within-platform (`/${currentPlatform}/...`); since the migration guide is a `[platform]` route, passing the current platform here would keep readers in place. Worth either doing that or confirming a single canonical guide is intended (note Next.js also has a dedicated `nextjs-v5-to-v6-migration-guide`).

src/styles/global-nav.scss
```diff
@@ -367,6 +367,50 @@
   }
 }
 
+// Amplify JavaScript v5 maintenance-mode banner
+.js-v5-banner {
```
osama-rizk, line 397: `.js-v5-banner*` (371-412) is byte-for-byte identical to `.gen1-page-banner*` (327-368), and `.gen2-maintenance-banner*` is nearly the same. Since the ask was to mirror the existing banner, this isn't "you duplicated" — it's that the pattern is now copied 3×. A shared `%banner-base` placeholder (or mixin) with per-banner color overrides would collapse all three. Not this PR's job to refactor the existing two, but worth considering before a third copy sets the pattern.

src/components/Layout/Layout.tsx
```diff
@@ -379,6 +382,7 @@ export const Layout = ({
                       />
                     ) : null}
                     {isGen1 && <Gen1Banner currentPlatform={currentPlatform} />}
+                    {isJsV5 && <JsV5MaintenanceBanner />}
```
osama-rizk: On a v5 JS page both `isGen1` and `isJsV5` are true, so `Gen1Banner` and this banner both render back-to-back — and they share the same amber style + uppercase badge, so it reads as two nearly-identical banners stacked. The PR intends both to show; just worth confirming the stacked result looks deliberate (the screenshot suggests it's acceptable) or differentiating them visually.

osama-rizk (commented):
Clean, well-scoped change and the detection logic is correct — I verified `isJsV5Page` returns true for all six JS platforms' `prev` routes and false for non-JS `prev` (v1), v6, gen2, and empty paths, and that `Layout` feeds it the real URL (consistent with `isGen1`). Test coverage is genuinely good for a banner.

One note on the description: it says this "mirrors the existing Gen 1 maintenance banner mechanism." It mirrors Gen1Banner's markup, but not its link style (Gen1Banner uses relative links) or its platform-awareness (Gen1Banner takes `currentPlatform`) — see the inline notes on the link.

Findings inline are all low-to-medium: one that matters for the next version cycle (route detection derives v5 from a proxy rather than `PLATFORM_VERSIONS`), a link-portability one (absolute prod URL for a same-site page), plus a few consistency/UX nits.

src/utils/isJsV5Page.ts
soberm: Agreed and fixed in 6870fe2 — this now reads `PLATFORM_VERSIONS` directly:

```ts
const [, gen, platform, version] = path.split('/');
if (gen !== 'gen1' || version !== 'prev') return false;
return PLATFORM_VERSIONS[platform as Platform]?.prev === 'v5';
```

One deviation from your snippet, deliberately: I kept the `version !== 'prev'` check. `PLATFORM_VERSIONS[platform].prev === 'v5'` answers "does this platform's prev mean v5", but not "is this reader actually on a prev route" — dropping the segment check makes `/gen1/react/build-a-backend/...` (a **v6** page) return true, so the v5 banner would show on every Gen1 v6 JS page. So the two checks are doing different jobs: the segment says *which version line the route is on*, `PLATFORM_VERSIONS` says *what that line currently resolves to*. Together they can't drift the way the old `JS_PLATFORMS` proxy could.

I also added a test that derives expectations from `PLATFORM_VERSIONS` rather than hardcoding platform names, so if a platform's `prev` rolls to `v6` the test flips automatically instead of silently passing:

```ts
Object.keys(PLATFORM_VERSIONS).forEach((platform) => {
  const isV5 = PLATFORM_VERSIONS[platform as Platform].prev === 'v5';
  expect(isJsV5Page(`/gen1/${platform}/prev/build-a-backend/`)).toBe(isV5);
});
```

src/components/JsV5MaintenanceBanner/JsV5MaintenanceBanner.tsx
soberm: Agreed and fixed in 6870fe2 — the URL is now built root-relative:

```ts
export const getMigrationGuideUrl = (platform: Platform) =>
  `/gen1/${platform}/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/`;
```

Verified on a dev server that the rendered href is now same-origin and keeps its trailing slash (`trailingSlash: true`):

```html
<a class="js-v5-banner__link" href="/gen1/react/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/">migration guide</a>
```

Worth recording one thing this surfaced: a unit test asserting the trailing slash failed, because `next/link` normalizes it from `next.config.mjs`, which jest doesn't load — so the href is slash-less in jsdom but correct in the real app. Rather than assert environment-specific output, the test now compares ignoring the trailing slash and the real form is covered by the server check above.

I also confirmed the target resolves for all six JS platforms rather than assuming it does — `HTTP 200` for each of `/gen1/{react,javascript,vue,angular,nextjs,react-native}/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/`.

src/components/JsV5MaintenanceBanner/JsV5MaintenanceBanner.tsx
soberm: Agreed and fixed in 6870fe2 — the banner now takes `currentPlatform` (fed from `Layout`, same source as `Gen1Banner`) and links within the reader's platform:

```tsx
<JsV5MaintenanceBanner currentPlatform={currentPlatform} />
```

Verified per-platform on a dev server — a reader on `/gen1/vue/prev/...` now gets `/gen1/vue/build-a-backend/...` rather than being switched to `javascript`.

On your second question, I checked rather than assumed: **a single canonical guide is intended.** The guide is one `[platform]` route whose `meta.platforms` lists all six JS platforms, and its body is a shared fragment (`src/fragments/lib/troubleshooting/common/upgrading.mdx`) filtered per platform. The dedicated `nextjs-v5-to-v6-migration-guide` you mention is already linked *from inside* that fragment, behind `<InlineFilter filters={["nextjs"]}>`:

> `<Callout>Find a comprehensive summary of changes to Server-Side Rendering in the [NextJS migration guide](/gen1/nextjs/build-a-backend/server-side-rendering/nextjs-v5-to-v6-migration-guide/)</Callout>`

So a Next.js reader following this banner lands on the Next.js-filtered canonical guide and gets the SSR-specific guide from there. Special-casing `nextjs` in the banner would skip the general migration content and duplicate routing logic that the guide already owns, so I left it pointing at the canonical guide for all six platforms.

src/styles/global-nav.scss
soberm: Good call — done in 6870fe2, before the third copy set the pattern. I extracted shared placeholders and all three banners now extend them:

```scss
%page-banner { /* layout: flex, gap, padding, radius, margin */ }
%page-banner--amber { /* filled amber treatment + darkMode */ }
%page-banner__badge { /* uppercase pill, no colors */ }
%page-banner__text { }
%page-banner__link { }
```

`.gen1-page-banner` and `.gen2-maintenance-banner` extend the base plus `--amber`, and only declare what actually differs (badge colors `#b45309` / `#92400e`, and gen2's `flex: 1` on text). Net effect on the stylesheet is a **reduction**: compiled CSS went 79,336 → 78,250 bytes.

I did touch the two existing banners, which you'd scoped out — the reasoning is that extracting a placeholder and leaving two verbatim copies beside it would have left the duplication in place while adding indirection. Since this is only worth doing if it's zero-risk, I verified it is rather than eyeballing it: I compiled `styles.scss` before and after, resolved the effective declarations for every pre-existing selector, and compared them as property→value maps.

```
UNCHANGED  .gen1-page-banner                                     UNCHANGED  .gen2-maintenance-banner
UNCHANGED  .gen1-page-banner__badge                              UNCHANGED  .gen2-maintenance-banner__badge
UNCHANGED  .gen1-page-banner__text / __link / __link:hover       UNCHANGED  .gen2-maintenance-banner__text / __link / __link:hover
UNCHANGED  [data-amplify-color-mode=dark] .gen1-page-banner      UNCHANGED  [data-amplify-color-mode=dark] .gen2-maintenance-banner
UNCHANGED  [data-amplify-color-mode=system] .gen1-page-banner    UNCHANGED  [data-amplify-color-mode=system] .gen2-maintenance-banner
UNCHANGED  [data-amplify-color-mode=dark] .gen2-maintenance-banner__close:hover
RESULT: all selectors identical
```

The only difference in raw output is declaration *order* (`background-color` now lands after the inherited layout props); no property is declared twice in a rule, so computed styles are identical. Happy to split the refactor into a separate PR if you'd rather keep this one to just the new banner.

src/components/Layout/Layout.tsx
soberm: You're right to push on this — I'd said "both are intended" but hadn't actually justified two amber banners with identical uppercase badges stacked on top of each other. Looking again, the stacked result was **not** deliberate enough: both banners had byte-identical styling and the badge text was literally the same string ("Maintenance Mode"), so it read as an accidental duplicate render rather than two distinct notices.

Fixed in 6870fe2 by differentiating them instead of leaving it to chance. The v5 banner is now a lighter, left-accented sub-notice with an outlined badge, and its badge text is "JavaScript v5":

|                  | `.gen1-page-banner` | `.js-v5-banner` |
|---|---|---|
| background       | `#fef3c7` (filled) | `rgba(245,158,11,.08)` (tint) |
| border           | `1px solid #f59e0b` (all sides) | `border-left: 3px solid #b45309` |
| badge            | filled `#b45309`, white text | transparent, `1px solid #b45309`, `#b45309` text |
| badge text       | "Maintenance Mode" | "JavaScript v5" |

So the Gen1 banner stays the primary generation-level notice and the v5 banner reads as a narrower version-specific note beneath it, rather than a second banner of equal weight. Both still render — the Gen1 message (EOL date, Gen 2 migration) and the v5 message (v5→v6 upgrade) carry different information and different link targets, so suppressing either on v5 pages would lose something.

Note I couldn't attach a rendered screenshot: Playwright isn't installed in this environment, so the visual claims above are from the compiled CSS and the rendered HTML rather than an image. If you'd like a screenshot in the PR before merging, I'm happy to add one, or the values above should be enough to eyeball it in a preview build.

soberm:
Thanks for the thorough review — all five inline findings are addressed in 6870fe2, with a reply on each thread.

osama-rizk (approved):

cadivus (approved):

